PR #15, opened 2016-07-02
Minor improvements

leo:
- Use caps for readme and history
- Added comments to .gitignore
- Description within package.json added
- Instead of `npm run launch`, we can now use `npm start` for starting the app
- Tiny tweaks within README.md


rauchg:
😍

PR #20, opened 2016-07-02
Add support to Fullscreen shorcut and menu item

montogeek:
This PR implements https://github.com/zeit/hyperterm/issues/1, by adding `Ctrl+Cmd+F` shortcut to switch between Window and Fullscreen modes, it also adds a Menu item under the Window menu to switch between modes


rauchg:
Thanks! There's actually a not very well documented functionality of `role` that automatically sets up the whole thing for common actions. I just added it in 79ad131..61a29c9

montogeek:
Oh, thanks :)

PR #25, opened 2016-07-03
Centered traffic lights & disabled text selection for indicator

leo:
- Fixes https://github.com/zeit/hyperterm/issues/6
- Prevents the user from selecting text within the update/resize indicator. Also shows a default cursor instead of the one for text within these indicators.

How it looks (I've used Sketch.app to precisely measure the sizes/paddings):

screen shot 2016-07-03 at 13 25 47


developit:
Sweet.

leo:
Fucked up the commit history ¯\_(ツ)_/¯

PR #26, opened 2016-07-03
Add fallback monospace fonts for Menlo

OlliV:
These fonts should look similar to Menlo, especially DejaVu Sans Mono should be a decent fallback since Menlo seems to be based on it.

### DIFF
--- a/app/css/hyperterm.css
+++ b/app/css/hyperterm.css
@@ -47,7 +47,7 @@ header {
   background: rgba(255, 255, 255, .2);
   padding: 6px 14px;
   color: #fff;
-  font: 11px Menlo;
+  font: 11px Menlo, 'DejaVu Sans Mono', 'Lucida Console', monospace;
   position: fixed;
   bottom: 20px;
   right: 20px;
@@ -63,7 +63,7 @@ header {
   background: rgba(255, 51, 76, .8);
   padding: 6px 14px;
   color: #fff;
-  font: 11px Menlo;
+  font: 11px Menlo, 'DejaVu Sans Mono', 'Lucida Console', monospace;
   position: fixed;
   bottom: 20px;
   right: 20px;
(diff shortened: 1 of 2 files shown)


rauchg:
We have to fix this for `hterm` now. It uses a specific preference in JS world. Alternatively, we could adapt the code to read the font from CSS (probably better)

jhaynie:
i'm using zsh + iterm and the Powerline font.  when i manually change the font preference for the terminal i can get my zsh terminal to render fine.  would be nice to have a preference pane to allow custom font settings.

OlliV:
@rauchg regardless of that, as long as there is any fonts in the css files I think we should have proper alternatives listed for different platforms, unless those are actually unused now.

OlliV:
You are right, this patch doesn't fix the issue anymore :(

leo:
IMHO, we should switch to [San Francisco Mono](http://daringfireball.net/linked/2016/04/20/san-fran-mono) on macOS + fallback for Windows, as soon as Apple publishes the font.

OlliV:
Any idea what will be the license for it? I'm wondering if we could just bundle the font with the release package.

leo:
@OlliV San Francisco Mono? You mean so that we can use it on windows as well? Because on macOS, we could just specify something like `font-family: -apple-system-mono`.

OlliV:
@leo yes. If it's not possible then another option would be to possibly bundle some other font for other platforms. Otherwise we'd have to figure out what works for different linux distributions, Android, Windows etc.

leo:
@OlliV Android? 😸

OlliV:
@leo :D Well, I just mean that why wouldn't we make it platform independent if it's easily possible. Fonts are still something that's not always portable across platforms.

rauchg:
Fixed. Thanks @OlliV

PR #32, opened 2016-07-04
add standard behavior when you double click window

jhaynie:
add standard behavior when you double click window title to maximize / unmaximize the window

### DIFF
--- a/app/hyperterm.js
+++ b/app/hyperterm.js
@@ -410,7 +410,9 @@ export default class HyperTerm extends Component {
         this.clicks = 0;
         this.maximized = !this.maximized;
     } else {
-        this.clickTimer = setTimeout(() => this.clicks = 0, 400);
+        // http://www.quirksmode.org/dom/events/click.html
+        // https://en.wikipedia.org/wiki/Double-click
+        this.clickTimer = setTimeout(() => this.clicks = 0, 500);
     }
   }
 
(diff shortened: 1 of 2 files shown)


app/hyperterm.js
```diff
@@ -398,6 +398,20 @@ export default class HyperTerm extends Component {
...
+    if (this.clicks++ >= 2) {
+        if (this.maximized) {
+            this.rpc.emit('unmaximize');
+        } else {
+            this.rpc.emit('maximize');
+        }
+        this.clicks = 0;
+        this.maximized = !this.maximized;
+    } else {
+        this.clickTimer = setTimeout(() => this.clicks = 0, 400);
```
rauchg: Please clear this on `componentWillUnmount`

rauchg:
Thanks a lot for your contribution.

The double click rate is OS-specific and sometimes user-settings-specific. What stopped you from registering the `dblclick` event?

In addition, a further double-click should re-set the window to its prior size. (Terminal.app seems to do that, if you need an example).

jhaynie:
click and dblclick events on the same element are impossible (by design): http://www.quirksmode.org/dom/events/click.html

the functionality works where each click with invert the other to match terminal / iterm type behavior.  so if you click it will maximize. another double click will unmaximize.

certainly once we have settings capability we can make the settings configurable.

rauchg:
My bad, I didn't realize electron had `unmaximize` :) This is brilliant!

daryl:
This doesn't seem to be working for me.

PR #34, opened 2016-07-04
change font size with command shortcuts

jhaynie:
Change the font size for the window with standard command + or -

### DIFF
--- a/app/hyperterm.js
+++ b/app/hyperterm.js
@@ -398,6 +398,20 @@ export default class HyperTerm extends Component {
   onHeaderMouseDown () {
     this.headerMouseDownWindowX = window.screenX;
     this.headerMouseDownWindowY = window.screenY;
+
+    this.clicks = this.clicks || 1;
+
+    if (this.clicks++ >= 2) {
+        if (this.maximized) {
+            this.rpc.emit('unmaximize');
+        } else {
+            this.rpc.emit('maximize');
+        }
+        this.clicks = 0;
+        this.maximized = !this.maximized;
+    } else {
+        this.clickTimer = setTimeout(() => this.clicks = 0, 400);
+    }
   }
 
   componentWillUnmount () {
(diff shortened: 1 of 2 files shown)


montogeek:
`command+=` is `Cmd` + `+`?

rauchg:
This is brilliant. We should probably memorize the option @jhaynie. Perhaps `localStorage`?

johanbrook:
👍. 

@rauchg Regarding memorizing the option, that should ideally live in some sort of global config somewhere? (out of scope for this PR, but still). Perhaps related to the `hyperterm.json` mentioned in https://github.com/zeit/hyperterm/issues/14.

rauchg:
On further reflection, it'd be cool to register it as menu accelerators:

rauchg:
@johanbrook yes, was thinking exactly that. But in order to make quick progress I'd like to use `localStorage` for now and merge this sooner rather than later.

app/hyperterm.js
```diff
@@ -263,6 +265,32 @@ export default class HyperTerm extends Component {
...
+    const term = this.refs[`term-${uid}`];
+    if (term) {
+      try {
+        const size = term.term.prefs_.get('font-size');
+        term.term.prefs_.set('font-size', size - 1);
+      } catch (e) {
+        alert(e);
+      }
+    }
+  }
```
sindresorhus: This functions are almost identical. I would DRY them up into one that accepted an argument of whether to increment or decrement.

montogeek:
It should also support switch back to the default size

jhaynie:
i'll add the menu accelerators and DRY up the functions. I think we should memoize these in a separate PR to keep this one focused.

jhaynie:
OK, i added reset behavior and added window shortcuts and DRY up code.

rauchg:
@jhaynie this is beyond beautiful and it works flawlessly. Only nit: we need to let the user know about the font change.

Here's my suggestion:
screen shot 2016-07-04 at 6 46 48 pm

rauchg:
Another minor nit: from the perspective of `<HyperTerm>` it'd be more elegant if we simply alter `<Term fontSize={this.state.fontSize} />` in the render loop.

Then we can capture it in `shouldComponentUpdate` and manually perform the transformation on the local term reference and return `false`.

rauchg:
I'm filing issues for these two things.

Issue #45, opened 2016-07-05
Implement font size indicator


rauchg:
![](https://cloud.githubusercontent.com/assets/13041/16571918/ba5b7ad8-4217-11e6-8230-5ea1a1faafc9.png)

(from #34, ping @jhaynie)

montogeek:
Will those boxes be configurable via `./hyperterm.json`? Would be great

rauchg:
To not show them at all ?

montogeek:
That is a possibility, I was thinking about colors or size

rauchg:
Yeah since it's non-terminal stuff, semantically I think going with San Francisco makes more sense there.

jhaynie:
i think #57 picks this up.

leo:
Fixed in https://github.com/zeit/hyperterm/pull/57

PR #57, opened 2016-07-05
font size improvements

### LINKED ISSUE #44: Implement font size <Term> property
From the perspective of <HyperTerm> it'd be more elegant if we simply alter <Term fontSize={this.state.fontSize} /> in the render loop.

Then we can capture it in shouldComponentUpdate and manually perform the transformation on the local term reference and return false.

(from #34, ping @jhaynie)

nfcampos:
closes #44 and #45 

- fixed keyboard shortcut registering (listeners were being called twice)
- now saving fontSize in the state of Hyperterm
- Term now accepts fontSize as a prop and internally sets it
- added font size indicator on change (next to rows x cols)

(bonus improvement: now font size is shared between all tabs)

### DIFF
--- a/app/hyperterm.js
+++ b/app/hyperterm.js
@@ -22,6 +22,7 @@ export default class HyperTerm extends Component {
       activeMarkers: [],
       mac: /Mac/.test(navigator.userAgent),
       resizeIndicatorShowing: false,
+      fontSizeIndicatorShowing: false,
       updateVersion: null,
       updateNote: null,
       fontSize: 12
@@ -91,7 +92,7 @@ export default class HyperTerm extends Component {
         }</div>
       </div>
       <div className={classes('resize-indicator', { showing: this.state.resizeIndicatorShowing })}>
-        <div>{ this.state.fontSize }px</div>
+        {this.state.fontSizeIndicatorShowing && <div>{ this.state.fontSize }px</div>}
         <div>{ this.state.cols }x{ this.state.rows }</div>
       </div>
       <div className={classes('update-indicator', { showing: null !== this.state.updateVersion })}>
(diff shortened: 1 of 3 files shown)


jhaynie:
nice, great refactor!

app/hyperterm.js
```diff
@@ -89,7 +91,8 @@ export default class HyperTerm extends Component {
         }</div>
       </div>
       <div className={classes('resize-indicator', { showing: this.state.resizeIndicatorShowing })}>
-        { this.state.cols }x{ this.state.rows }
+        <div>{ this.state.fontSize }px</div>
```
rauchg: does this mean we always show the font size when resizing? perhaps a bit too much info?

rauchg: ideally, we only show the font size change when the font size changes, and the cols and rows when the cols and rows change

app/hyperterm.js
```diff
@@ -270,21 +273,12 @@ export default class HyperTerm extends Component {
...
 
   resetFontSize () {
-    const uid = this.state.sessions[this.state.active];
-    const term = this.refs[`term-${uid}`];
-    if (term) {
-      // TODO: once we have preferences, we need to read from it
-      term.term.prefs_.set('font-size', 12);
-    }
+    // TODO: once we have preferences, we need to read from it
+    this.setState({fontSize: 12});
```
rauchg: Spacing after `{` + `}`. I should clarify in the README that we follow `standard` by @feross

nfcampos: fixed

app/hyperterm.js
```diff
@@ -373,8 +367,8 @@ export default class HyperTerm extends Component {
       keys.bind('command+shift+]', this.moveRight);
       keys.bind('command+alt+left', this.moveLeft);
       keys.bind('command+alt+right', this.moveRight);
-      keys.bind('command+=', this.increaseFontSize);
-      keys.bind('command+-', this.decreaseFontSize);
+      keys.bind('command+plus', this.increaseFontSize);
+      keys.bind('command+minus', this.decreaseFontSize);
```
rauchg: is this still needed since we now use the electron accelerators in `index.js` ?

rauchg:
Super clean stuff @nfcampos

app/hyperterm.js
nfcampos: you're quite right (this is probably why the listener was being called twice) fixed

app/hyperterm.js
nfcampos: yep makes sense, i've made it like you said

app/hyperterm.js
```diff
@@ -269,30 +273,28 @@ export default class HyperTerm extends Component {
...
-    if (term) {
-      // TODO: once we have preferences, we need to read from it
-      term.term.prefs_.set('font-size', 12);
-    }
+    this.changeFontSize(12);
   }
 
   increaseFontSize () {
-    this.changeFontSize(1);
+    this.changeFontSize(1, true);
```
nfcampos: I'm not the biggest fan of boolean arguments but I didn't want to repeat that `setTimeout` thing in two functions...
I can change it to `this.changeFontSize(1, {increment: true})`, should I?

rauchg: `{ relative: true }` ?
Also, are we cleaning up `fontSizeIndicatorTimeout` on unmount?

nfcampos: yeah `relative` is better.
good catch, I'd forgotten about that
I've made both changes

rauchg:
Amazing work @nfcampos

jdsimcoe:
Does this include changing the types of fonts @rauchg?

rauchg:
We're doing that through config, see the other PR.

jdsimcoe:
What PR?

PR #58, opened 2016-07-05
Add close cross for tabs

johanbrook:
Concerns https://github.com/zeit/hyperterm/issues/40.

This patch adds a SVG close cross for each tab.

- Alters `Hyperterm#closeTab()` to take an argument: `id` (the tab index).
- Adds `Hyperterm#closeActiveTab()` method for closing the current tab.
- Adds `Icon` React component.

**Preview**

skarmavbild 2016-07-06 kl 01 50 22

skarmavbild 2016-07-06 kl 01 50 26


app/icon.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
```
rauchg: let's call this `close-icon.js` plz ?

rauchg: or even better, `icons/close.js`

johanbrook: Sure, but I thought it would work well as a reusable component for small inline SVG shapes.

rauchg: Yep, if you want to extend from it or whatnot. But isn't this the specific close icon ? (due to the `<path>`)

johanbrook: It can be extended with more `path`s in the `switch-case` block if wanted :) A SVG symbol sheet could be used instead, if we don't want a full React component for it. But I don't see any downside with it tbh.

rauchg: Right but the more "Componenty" approach for the extension would be composition:

```
<Icon>
 <path>…
</Icon>
```

the `switch` case would conflate all the possible icon combinations.

rauchg: For now, calling it `close-icon` sounds like a good idea to me, since that's what it is. We can revisit it in the future.

johanbrook: > the switch case would conflate all the possible icon combinations.

I agree on that, I found that particular solution a bit weird when I first read about it a while ago, but thought that "this is how they do it in React world" :shrug: ;)

I'll fix it.

johanbrook:
@rauchg I implemented it as an SVG sheet with symbols, which gives us flexibility of styling icons with colors and such, but still keeping the SVG code itself external. This scales really well too. Whatcha think?

```jsx
<svg className='icon'>
   <use xlinkHref='assets/icons.svg#close'></use>
</svg>
```

☝️ Could be made into a reusable component too I guess if we wanna.

rauchg:
That's beautifully done. Thanks for educating me @johanbrook

johanbrook:
⚡ 👍

PR #59, opened 2016-07-06
added JsonConfig component to read/set .hyperterm.json

nfcampos:
- JsonConfig renders its child and injects two props
  - config: the contents of the config file specified in path
  - setConfig: a function like setState that can be used to change the config
- fontSize is now saved and read from the config

@rauchg is this what you had in mind for saving config to `.hyperterm.json`?
 
Missing:
- JsonConfig should watch config file path and call `this.read` every time it changes


app/index.js
```diff
@@ -1,8 +1,14 @@
...
 import HyperTerm from './hyperterm';
 import React from 'react';
+import JsonConfig from './json-config';
 
 require('./css/hyperterm.css');
 require('./css/tabs.css');
 
-render(<HyperTerm />, document.getElementById('mount'));
+render(
+  <JsonConfig path='~/.hyperterm.json' defaults={{ fontSize: 12 }}>
```
nfcampos: the defaults should probably be defined in a separate file as json and imported

app/json-config.js
```diff
@@ -0,0 +1,60 @@
...
+    const child = React.Children.only(this.props.children);
+    return React.cloneElement(child, {
+      config: this.state.config,
+      setConfig: this.write
+    });
+  }
+
+  read (path = this.props.path) {
+    const absolutePath = expandPath(path);
+    const config = JSON.parse(fs.readFileSync(absolutePath));
```
nfcampos: this probably needs a try catch around this line to do something special if the file is invalid json. what exactly I'm not sure

app/webpack.config.js
```diff
@@ -5,6 +5,7 @@ const nodeEnv = process.env.NODE_ENV || 'development';
 const isProd = nodeEnv === 'production';
 
 module.exports = {
+  target: 'electron-renderer',
```
nfcampos: this is needed to be able to import electron from the renderer process

app/json-config.js
```diff
@@ -0,0 +1,60 @@
...
+  read (path = this.props.path, defaults = this.props.defaults) {
+    const absolutePath = expandPath(path);
+    const config = JSON.parse(fs.readFileSync(absolutePath));
+    const configWithDefaults = Object.assign({}, defaults, config);
+    this.setState({ config: configWithDefaults });
+  }
+
+  write (changes) {
+    const absolutePath = expandPath(this.props.path);
+    const config = JSON.parse(fs.readFileSync(absolutePath));
```
nfcampos: the read and write here can probably be async

rauchg:
Here's what I have in mind for `~/.hyperterm.json`.

1. Use something like https://github.com/kaelzhang/node-comment-json to preserve comments.

```js
{
  "config": {
    // default font size
    "font size": "13px",

    // font smoothing makes fonts look better in retina displays
    "font smoothing": true,

    // enable plugins
    plugins: true
  },
  
  // a list of plugins, which can simply be the module name,
  // or the module name and its options:
  //  "module"
  //  ["module", { with: { options: { } }]
  "plugins": [
    "@hyperterm/remember-pwd"
  ]
}
```

2. We should give plugins the ability to write to their namespace:

```js
module.exports = function (cfg) {
  // do something
  cfg.merge({
    some: 'data'
  });
}
```

I like the idea of the API for configuration being "merge" (kinda like `setState`), instead of set.
If you need to read + merge you do:

```js
cfg.merge((data) => {
  return { some: { extension: data.something } };
});
```

Set `undefined` to remove a key (like JSON.stringify does when it encounters it)

rauchg:
ps: I don't like camel case in config, but I can see how it's convenient when merging to components.

rauchg:
On further reflection, I think we should maintain `hyperterm.json` as something only *users* write to. 

Then we should have configuration that works more as a cache that the core and plugins can read and write to. Does that make sense?

rauchg:
I'm starting to think that makes a lot of sense. Ultimately, we want to preserve the user setting to that the user can *reset* their font size back to what they configured. Then when we load the editor, we set it first to the ephemeral cached value they arrived to with cmd+plus/minus (if any), falling back to the config one.

rauchg:
As such, we might want to have `this.props.config` (readonly) and `this.props.cache` (read+write). Both reloadable.

rauchg:
Our cache could be powered by: https://github.com/sindresorhus/electron-config

nfcampos:
- I see the point in keeping a cache (though I find the name confusing, this would be a cache that would be purposefully out of sync with what it caches?) so that no all changes affect what the user wrote in `hyperterm.json`.
- However, it does go a bit against the goal of making `hyperterm.json` shareable, because two users might have the same `hyperterm.json` but different caches so they'd end up seeing different things.
- If we do go with both I think only one `config` prop should be injected into `Hyperterm` (which would be the result of merging `config` with `cache`). is there any reason why the `Hyperterm` component should care whether some part of the config comes from `hyperterm.json` or the cache? I think at that level it should be transparent.
- the other prop would be a `setConfig` (or some other name) which only writes to (or clears) stuff in the cache, not touching `hyperterm.json`

nfcampos:
I'm not a big fan of saying a file is json and then adding comments, making it unparseable as json, but I do see the use of having comments in there. So maybe we should use some format that includes comments, like `yaml`?

rauchg:
@nfcampos there are a lot of cached settings that are very much not worth sharing precisely. For example, the cached window size sounds like overkill.

We can also go with webpack style and do a .js file with `exports`.

nfcampos:
@rauchg yeah you're probably right
what do you think about having just one prop `config` (that is the result of merging `config` and `cache`) instead of two?

package.json
```diff
@@ -6,6 +6,7 @@
   "dependencies": {
     "child_pty": "3.0.1",
     "default-shell": "1.0.1",
+    "electron-config": "^0.2.0",
```
nfcampos: this isn't a dependency in `app/package.json` because it has a dynamic require which makes it not work with webpack

nfcampos:
@rauchg how about this latest commit?

package.json
rauchg: can you please pin it?

rauchg:
I took the component decoration idea, added file watching, memoization, default configuration initialization, notifications, VM-based evaluation. The file is now called `~/.hyperterm.js`.

🚀

nfcampos:
@rauchg cool! I've taken a look at it and it looks great! I just left two comments here https://github.com/zeit/hyperterm/commit/2a6ec3e95508c2cf974b5d1d01e53498123715b8
You still want to add the cache for non-user-facing config stuff?

PR #60, opened 2016-07-06
Better default colors

dzannotti:
Provides better default colors for bash (more pastel), relates to https://github.com/zeit/hyperterm/issues/43

### DIFF
--- a/app/colors.js
+++ b/app/colors.js
@@ -7,4 +7,4 @@ export default [
   '#a468f7', // magenta
   '#f19c37', // cyan
   null
-]
+];
--- a/app/colors.js
+++ b/app/colors.js
@@ -1,0 +1,10 @@
+export default [
+  null,      // black
+  '#c4385e', // red
+  '#b2e147', // green
+  '#e5db7d', // yellow
+  '#7dd7ee', // blue
+  '#a468f7', // magenta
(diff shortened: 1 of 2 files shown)


rauchg:
So the overrides pref expects a list ? O.o

rauchg:
Care to share a screenshot @dzannotti ? Thank you!

dzannotti:
@rauchg sure thing here you go (keep in mind my .bash_prompt powerline additions that are not included in this PR) screen shot 2016-07-06 at 12 01 27

The reason it requires array form (or indexed object) is because it needs to overwrite the ascii palette (255 indexed colors), the source for this prefs is helpful to understand the context
https://github.com/dbkaplun/hterm-umdjs/blob/master/dist/index.js#L7867
https://github.com/dbkaplun/hterm-umdjs/blob/master/dist/index.js#L10815

I think a new settings page where each user could customize this colors, fontsize, font might be a better approach than forcing someone's defaults into everyone

nfcampos:
@dzannotti that red (to my eyes at least) is difficult to see against the black background

rauchg:
@hharnisc did a really great job. I like the new defaults! And config is coming

rauchg:
Thanks @dzannotti

PR #61, opened 2016-07-06
improved default color palette

hharnisc:
High contrast

screen shot 2016-07-05 at 9 38 53 pm

zeit.co (for reference)

screen shot 2016-07-05 at 10 08 29 pm


rauchg:
I wanna give it a shot :) thank you. If it's too startling we can try the pastel one.

PR #71, opened 2016-07-07
Get rid of border-radius in fullscreen mode (macOS) fixes #21

### LINKED ISSUE #21: Get rid of `border-radius` in fullscreen mode (macOS)
Hopefully the fullscreen events fire in the main browser `window` to make this easier.

hharnisc:
Tried this with and without the border, and it looks better with a border.

screen shot 2016-07-06 at 6 37 16 pm

### DIFF
--- a/app/css/hyperterm.css
+++ b/app/css/hyperterm.css
@@ -7,10 +7,6 @@
   border: 1px solid #333;
 }
 
-.mac.main {
-  border-radius: 5px;
-}
-
 header {
   position: fixed;
   top: 1px;


hharnisc:
It does remove the radius all the time in the CSS, however OS X puts rounded borders on it without the radius:

screen shot 2016-07-07 at 7 37 58 am

It's got a couple disadvantages though

- not consistent across OSs
- to the observant, the corners lose their border right around their edges

I can see the benefit of having a consistent way in the CSS to detect full screen mode. Any objection to adding that to the `body` tag?

```html
<body class='fullscreen'>
```

xaviervia:
I actually like the border in fullscreen.

rauchg:
The community has spoken. Everyone seems to like the border in fullscreen :D

PR #76, opened 2016-07-08
configurable cursor opacity

hharnisc:
Before
screen shot 2016-07-07 at 7 26 07 pm
After
screen shot 2016-07-07 at 8 09 02 pm

### DIFF
--- a/app/lib/components/term.js
+++ b/app/lib/components/term.js
@@ -1,5 +1,6 @@
 /* global Blob,URL,requestAnimationFrame */
 import React from 'react';
+import Color from 'color';
 import hterm from '../hterm';
 import Component from '../component';
 
@@ -26,7 +27,7 @@ export default class Term extends Component {
     this.term.prefs_.set('font-family', props.fontFamily);
     this.term.prefs_.set('font-size', props.fontSize);
     this.term.prefs_.set('font-smoothing', props.fontSmoothing);
-    this.term.prefs_.set('cursor-color', props.cursorColor);
+    this.term.prefs_.set('cursor-color', Color(props.cursorColor).rgbString());
     this.term.prefs_.set('enable-clipboard-notice', false);
     this.term.prefs_.set('foreground-color', props.foregroundColor);
     this.term.prefs_.set('background-color', props.backgroundColor);
@@ -181,7 +182,7 @@ export default class Term extends Component {
     }
 
     if (this.props.cursorColor !== nextProps.cursorColor) {
-      this.term.prefs_.set('cursor-color', nextProps.cursorColor);
+      this.term.prefs_.set('cursor-color', Color(nextProps.cursorColor).rgbString());
     }
 
     if (this.props.colors !== nextProps.colors) {
--- a/app/package.json
+++ b/app/package.json
@@ -6,6 +6,7 @@
   "repository": "zeit/hyperterm",
   "dependencies": {
     "aphrodite-simple": "0.4.1",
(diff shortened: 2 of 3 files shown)


rauchg:
I want this cfg flag! Will need a refactor.

hharnisc:
Working on refactoring this now, expect it soon as long as I don't get interrupted

hharnisc:
@rauchg could you give this another look? The refactor done here is beautiful by the way, very well done!

huytd:
Will it be conflicted with the `cursorColor` if we put a `rgba` value in it?

rauchg:
@huytd that actually sounds like a really good idea. @hharnisc do you think it's tractable to try to extra the alpha value from `cursorColor`?

hharnisc:
@huytd @rauchg this seems like a good idea, since it removes the need to add another line in the [getStyleSheet](https://github.com/zeit/hyperterm/blob/master/app/lib/components/term.js#L93) function.

I'm thinking using the `color` module to parse out the cursor color (hex, rgb(a), hsl(a), ...) then set the alpha value to the `cursorOpacity` setting.

Here's a code sample:

https://tonicdev.com/578b9f2f9cdcc013002a89e6/578b9f2f9cdcc013002a89e7

In this scheme the cursorOpacity always overrides the cursorColor alpha value.

hharnisc:
@rauchg updated the fork if you'd like to give it another go. 

This works and seems predictable, but I could see an argument for only having a cursor color that allows you to set alpha and not having a cursorOpacity as well. But I think that would need to be well documented.

I could also see an argument for allowing cursor opacity set to null or undefined to disable it. 

I think I'd lean toward this implementation (override cursor color) since I would expect more people would catch

```js
cursorColor: '#333',
cursorOpacity: 0.75
```

over 

```js
cursorColor: 'rgba(50, 50, 50, 0.75)',
```

huytd:
Yeah. This is better. So we can have both `rgba` value in `cursorColor` and a separate `cursorOpacity` property, which will be higher priority and it can overrides the `cursorColor`.

This is a good way. I think we need to have this documented carefully :D 

👍 my vote up for this

hharnisc:
Thank you for the feedback @huytd 😄

filipelinhares:
Hi guys, I've seen some theme packages using this snippet of code:

```css
.cursor-node {
    mix-blend-mode: difference;
}
```

**Before**
image

**After**
image

hharnisc:
@filipelinhares That's pretty cool! Wasn't familiar with that CSS rule. That seems more like another configuration flag to me though, since it's doing color blending rather than opacity.

hhsnopek:
I don't think splitting the color and opacity is a great idea, considering that we have `rgba` for the specific reason of adding alpha. You could also use css4 `#[alpha][alpha][color][color][color][color][color][color]`, `#DDffffff` light opacity on white

rauchg:
I would strongly prefer to have only *one* config (I don't want that file to be huge). What we can do is provide our default purple in the `rgba` format and clarify in the text that we support any css color format :)

Also, @filipelinhares you just blew my mind. I didn't know that existed.

Thanks @hharnisc and everyone else for pitching in!

hharnisc:
Glad we're getting down the bottom of this! This implementation feels pretty good and the configuration seems really straight forward. Nice work everyone!

@rauchg ready to go when you are 😄

rauchg:
Sooooooo awesome

PR #80, opened 2016-07-11
Add configurable styling for resize indicators

johnotander:
This provides the ability to configure the
styling of resize indicator through the
.hyperterm.js file.

Currently the `resizeIndicator` object
is passed to the div, perhaps we might
only want to allow a green list of properties?

Somewhat related to #45


rauchg:
This can probably be done as an extension now.

PR #85, opened 2016-07-15
Add missing shebang to scripts/install.sh

amio:
Fix the error while executing in some none-bash shell(eg. fish):

```
  Failed to execute process './scripts/install.sh'. Reason:
  exec: Exec format error
  The file './scripts/install.sh' is marked as an executable but could not be run by the operating system.
```


rauchg:
Great. Thank you

PR #90, opened 2016-07-15
Fix two comments in default config file

alireza-ahmadi:
The comments of termCSS and padding attributes in default config file had some problems, I think the comment of termCSS attribute was replaced by padding attribute's comment. I have fixed them, a super minor change.


rauchg:
beautiful. thanks!

PR #104, opened 2016-07-15
Improved installation method

amilajack:
Previous method assumes a global installation of webpack, which isnt necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ to the app are detected.
 ```bash
 $ cd app/
 $ npm install
-$ webpack --watch
+$ npm run dev
 ```
 
 Then you can run in the main directory:


leo:
@amilajack There's no script named "dev" within `package.json`.

amilajack:
There is one in the app/package.json file

leo:
@amilajack Oh, yep. That's true! — Maybe we should put all in one? Might be a bit confusing for people that are just getting started with collaborating on the project...

amilajack:
That would be a good idea

leo:
@amilajack Wanna do that within this PR? 🚀 ⭐

amilajack:
But @zeit might have reasons for arranging it this way.

leo:
@rauchg ^ ?

amilajack:
Oh my bad. Yea, @rauchg. He followed the two-package.json structure recommended by electron-builder. And how would we run an npm script from the parent directory (the root directory) of the app folder?

montogeek:
We should follow electron-builder recommendations

amilajack:
I agree.

rauchg:
Good stuff. Thank you. I didn't spot it cause I have `./node_modules/.bin` in `$PATH`.

PR #107, opened 2016-07-15
Initial travis ci support

amilajack:


rauchg:
Woah so cool! Thanks @amilajack

PR #109, opened 2016-07-15
Bring focus to main window when clicking preferences

### LINKED ISSUE #105: Bring focus to main window after clicking "Preferences..."
Currently, it simply stays in the background while doing that.

leo:
This closes #105

### DIFF
--- a/menu.js
+++ b/menu.js
@@ -6,7 +6,7 @@ const appName = app.getName();
 // based on and inspired by
 // https://github.com/sindresorhus/anatine/blob/master/menu.js
 
-module.exports = function createMenu ({ createWindow, updatePlugins }) {
+module.exports = function createMenu ({ createWindow, updatePlugins, mainWindow }) {
   return [
     {
       label: 'Application',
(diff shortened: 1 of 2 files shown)


leo:
@rauchg I would appreciate being a collaborator, btw 😊

menu.js
```diff
@@ -20,6 +20,10 @@ module.exports = function createMenu ({ createWindow, updatePlugins }) {
           click (item, focusedWindow) {
             if (focusedWindow) {
               focusedWindow.rpc.emit('preferences');
+              return;
+            } else if (!mainWindow.isFocused()) {
+              mainWindow.on('focus', () => mainWindow.rpc.emit('preferences'));
```
rauchg: - What if there's no `mainWindow` (all closed)? 
- If you don't use `once`, wouldn't preferences fire every subsequent time the window gets `focus`?

rauchg:
Needs rebase

leo:
Figured out that it makes a lot more sense to spawn a new window and open the preferences there instead of putting one of the open windows back into focus (since there are multiple windows, we shouldn't just pick one).

So I've created https://github.com/zeit/hyperterm/pull/235, which does exactly that.

PR #130, opened 2016-07-15
Remove $ from terminal commands in README

Tyriar:
So they can be copied and pasted a block at a time


README.md
```diff
@@ -16,23 +16,23 @@ To install `package.json` dependencies in a way where the native
...
 running `webpack` with `--watch` so that any changes you make
 to the app are detected.
 
 ```bash
-$ cd app/
-$ npm install
-$ webpack --watch
+cd app/
+npm install
+webpack --watch
```
amio: `npm run dev` instead of `webpack --watch` would be better :)
`webpack --watch` won't work if user doesn't have a global installed webpack.

rauchg:
Please rebase! Will definitely merge.

Tyriar:
Done

rauchg:
Thanks @Tyriar ;)

Issue #88, opened 2016-07-15
Preferences shortcut not working properly in zsh


tamagokun:
This happens when pressing Cmd + ,

```
zsh: command not found: #
```

Seems like the comments before `$EDITOR ~/.hyperterm.js` are not liked by zsh.

PR #153, opened 2016-07-16
Make "Open Preference" compatible with none-bash shell. Fix #143

### LINKED ISSUE #143: "Open Preference" command should not use Bash by default
I'm not using bash and `$EDITOR ~/.hyperterm.js && exit` is an invalid command in Fish

screen shot 2016-07-15 at 6 09 19 pm

amio:
Note: the tab no longer auto closed.
But I think it's an acceptable approach, still better than throwing error :P

### DIFF
--- a/app/lib/actions/ui.js
+++ b/app/lib/actions/ui.js
@@ -129,9 +129,9 @@ export function showPreferences () {
           rpc.once('session data', () => {
             dispatch(sendSessionData(
               uid,
-              ['# Attempting to open ~/.hyperterm.js with your $EDITOR',
+              ['# Attempting to open ~/.hyperterm.js with your editor',
                '# If this doesn\'t work, open it manually with your favorite editor!',
-               '$EDITOR ~/.hyperterm.js && exit',
+               'bash -ic \'${EDITOR:-vi} ~/.hyperterm.js\'',
                ''
               ].join('\n')
             ));
--- a/app/lib/actions/ui.js
+++ b/app/lib/actions/ui.js
@@ -129,8 +129,8 @@ export function showPreferences () {
           rpc.once('session data', () => {
             dispatch(sendSessionData(
               uid,
-              ['# Attempting to open ~/.hyperterm.js with your editor',
-               '# If this doesn\'t work, open it manually with your favorite editor!',
+              ['echo \'Attempting to open ~/.hyperterm.js with your editor\'',
+               'echo \'If this doesn\\\'t work, open it manually with your favorite editor!\'',
                'bash -ic \'${EDITOR:-vi} ~/.hyperterm.js\'',
                ''
               ].join('\n')


leo:
I'm not using an editor within my terminal, but it looks like vim is now **always** used for opening the preferences? As far as I know there are people out there who aren't using vim, but rather nano or something else? What about them? It's not configurable anymore.

amio:
Indeed some people may choose a specific editor over vi/vim, while I think more people even haven't set a default $EDITOR, but that's just my guess. 🤔

Always use vi maybe not the best choice for everyone, but it just worked, for most people. So I take this compromise 💁

rauchg:
Maybe we can provide an alternative to `$EDITOR` if not present instead?

rauchg:
`: ${EDITOR=vi}`

rauchg:
Also, it'd be nice to test this for people that set `$EDITOR` to use atom etc

amio:
Great, thanks for your hint!

I've tried `bash -ic \'${EDITOR:-vi} ~/.hyperterm.js\'`, it works fine, when `$EDITOR` was set to `atom` it will open atom and exit the bash session, is there anything we should deal for these `atom` or `choc` editor?

kribblo:
I'm a vim guy myself, but probably `nano` is a more sensible fallback for users that hasn't set `$EDITOR`.

Even more sensible would be to use something like [node-open](https://github.com/pwnall/node-open) so to maximize the chance of opening something sane and comfortable on all platforms. An override could also be provided in `.hyperterm.js` instead of checking `$EDITOR`.

mfpierre:
You could also take the opportunity to close #88 removing the comments and trying some `echo` instead ? Or using `setopt interactivecomments` ?

amio:
@kribblo `node-open` may be a more convenient solution overall, but I'm not sure if it fits HyperTerm's philosophy, such a heavy(than a simple shell script) solution could slow down the app, even just a little bit. 😁

Besides, I want to make a quick fix for this feature, I think it as a handy guide for people just installed HyperTerm, after that, people likely choose their own way to open the `.hyperterm.js`. (For me, I usually `choc ~/.hyperterm.js`)

@mfpierre I've updated the PR, it should work by now 🙏
image

c0b41:
Also fish shell doesn't support && string  :+1:  and @amio why use template string? electron already support template strings..

amio:
@c0b41 While, current eslint rule only allow single quotes 😢 unless specified in the config `allowTemplateLiterals` http://eslint.org/docs/rules/quotes

kribblo:
> node-open may be a more convenient solution overall, but I'm not sure if it fits HyperTerm's philosophy, such a heavy(than a simple shell script) solution could slow down the app, even just a little bit. 

@amio smilies aside, that comment is so strange it's almost "[not even wrong](https://en.wikipedia.org/wiki/Not_even_wrong)". Heavy? Wat? Slow down the app? How? Even if that was how things worked, it's not 2000 anymore. 

It does not even seem to be a goal either, however it does seem to be a goal to be cross platform given the promise of downloads for more platforms.

It could be a goal to launch an editor inside the terminal, but I see Atom mentioned and it *is* a GUI terminal application... 

It could also be a goal to minimize dependencies (some people like to do all the maintenance themselves) in which case those few lines (I'm guessing effectively *less* lines than now to make it work) could be copied from another project.

I can of course submit that patch myself, but won't if the stated goal would actually be a real one.

amio:
@kribblo After dig into the [node-open](https://github.com/pwnall/node-open)'s code, it has 55 sloc with no deps, this shouldn't be called "heavy", sorry about using that word.

But I still prefer the one-line-script over an-external-package base on two points:
- HyperTerm put a lot effort to speed up the booting. On my mac it logs: `total init time 139.3350000000001`, I don't want to add a single extra millisecond to it if I don't have to.
- "Open Preference" is not a feature people would use it frequently, not as important as the startup time. Like I said, it's a compromise.

kribblo:
@amio again, that is simply not how it works. But I'll leave you to it, making my own patch if I decide to keep using the terminal (still evaluating)

amio:
Since @szhu has fixed this with an even cleaner script `exec env $EDITOR ~/.hyperterm.js` #267, I'm going to close this PR.

@mfpierre For the `#` problem with zsh, would you make another PR? That could be easier getting merged I think. 😔

szhu:
@amio I think you have some good concerns and they deserve to be addressed, so I'll do so here:

> But I still prefer the one-line-script over an-external-package base on two points:

> 1. HyperTerm put a lot efforts to speed up the booting. On my mac it logs: total init time 139.3350000000001, I don't want to add a single extra millisecond to it if I don't really have to.

I don't think node packages that are unused at boot add to the boot time. Have you timed and seen that it added a single millisecond? This might be a case of premature optimization. (And I think the package might actually make it faster at opening the preferences file!)

> 2. "Open Preference" is not a feature people would use it frequently, not as important as the startup time. Like I said, it's a compromise.

This is a good point. However, almost everything in life is a trade-off to some extent. It's easy to look at a number and say "it must stay at 0 no matter what." It's much harder and much more meaningful to compare to proposed solutions and look at which one provides a better holistic experience. I haven't tested this myself, but I'd predict that even up to 10ms (barely noticeable!) for a better preferences experience might be worth it.

amio:
@szhu Glad to have friendly discussion on this topic,

> I don't think node packages that are unused at boot add to the boot time.

- Every imported package will be bundled into a [asar](https://github.com/electron/asar), it will take time to parse, it will use memory, so it does affect the boot time.

> Have you timed and seen that it added a single millisecond?

- Nope, I didn't take it as a serious problem, frankly I didn't think it would make significant difference either, I just **prefer** the shell script solution. `node-open` is a 55 sloc package, I won't against if someone make a PR using `node-open`, I'm fine with it, I just wouldn't do it myself.

> I'd predict that even up to 10ms (barely noticeable!) for a better preferences experience might be worth it.

- "10ms" is a very personal threshold, someone would give "100ms" for that and another one may give only "1ms".
- "better experience" - it depends. For me the shell script open `.hyperterm.js` with Chocolat, it's already good enough(or even best?) solution, `node-open` couldn't do more than that ╮(╯_╰)╭
- I think I would give "2ms" to trade for `node-open` solution, for better compatibility in the future on windows platform :)

For this only PR, it's more like a choice about personal taste, but the reason behind this decision, I view it as the project's taste:

1. Shall we trade 10ms boot time for xxx feature?
2. Shall we trade another 10ms boot time for yyy feature?
3. ...(this would be a long long list)

That's the reason I stick to "I don't want to add a single extra millisecond to it if I don't really have to.". For now HyperTerm has [10 deps](https://github.com/zeit/hyperterm/blob/0.6.0/package.json), I wish it could keep that low.

Overall, I'm not saying what's "right" or I believe what's "right", these just opinions. I'll stick to what I would/won't do, and welcome to anyone's contribution to this project, no matter the solution I prefer or not, a solution **is** solution.

szhu:
> Every imported package will be bundled into a asar, it will take time to parse, it will use memory, so it does affect the boot time.

Does it get parsed at boot? If not, it shouldn't add to boot time.

> > Have you timed and seen that it added a single millisecond?
>
> Nope, I didn't take it as a serious problem, frankly I didn't think it would make significant difference either, I just prefer the shell script solution. node-open is a 55 sloc package, I won't against if someone make a PR using node-open, I'm fine with it, I just wouldn't do it myself.

The point I want to make about premature optimization is that modern code is so complex that at first glance you can't even know if a change will add or remove time. Specifically, your argument seems based on the fact that using a package looks like it would add to boot time. This is not a fact. It's not "probably true." We don't know this to be true at all. This notion is hard to take in because it's so instinctive – of course if we add more imported things it'll take more time right? Nope, we can't know this at all until we verify. When you remove this assumption you see how what you have left isn't objective at all.

---

Oh also here's a pretty long sidenote (which you totally ignore this if you want).  Please don't take it personally, I people (including me) doing this a lot; I just like to address it every once in a while. And just in case it wasn't clear, I really appreciate your replies and I don't mean to offend!! Just trying to be constructive in the best way I can.

In case you're wondering why some other commenters might have not addressed your concerns directly, my guess is that it's because your initial response to `node-open` set the tone for the rest of the conversation:

> Besides, I want to make a quick fix for this feature, I think it as a handy guide for people just installed HyperTerm, after that, people likely choose their own way to open the .hyperterm.js. (For me, I use CMD-,, choc or vim depends on mood 😝)

If you look closely at your wording, I think you turned the discussion from what users want to how you personally feel about this change. When you're proposing a feature, no one's going to argue with you about what you personally want (you know yourself best) nor are they going to counter it with what they personally want (collectively the two of you are like 0.1% of all users and what most users want probably has much higher priority).

If you want a response, you might want to rephrase it like (and this is my best guess at what you meant to say) "I don't think many people will use this feature. It's okay if we even just display print out a note telling them where the file is, so it should be okay if this feature doesn't launch the file in the preferred way for some users." I disagree with this btw, but if this is what you meant, say it, and people will be happy to point out which parts they disagree with.

Also, a sidenote about emojis. Happy emojis always lighten the mood, so use them as much as possible right? Nah, I think what they did in this thread is make people feel bad for saying things that might counter your personal feelings. (Heck, I feel bad just writing this.) I'd recommend toning emojis down during discussion/debate-like conversations… they make subtle mockery of the discussion.

Last thing, you stated that "I'm fine with it, I just wouldn't do it myself." However, your original arguments all relate to how good you think this change is and not how hard it would be to submit such a PR. So it actually looks like you seem less than fine with the change, but it's stated so subtley that it's really hard to see and most people won't notice or call you out on it.

---

All these details are each really really minor! But they kind of do add up and probably make contributors want to look at other issues instead. Fixing them takes a lot of time (just writing this comment took about an hour). But making these changes will probably get you more feedback on your future contributions.

szhu:
>> I'd predict that even up to 10ms (barely noticeable!) for a better preferences experience might be worth it.
>
> - "10ms" is a very personal threshold, someone would give "100ms" for that and another one may give only "1ms".
> - "better experience" - it depends. For me the shell script open .hyperterm.js with Chocolat, it's already good enough(or even best?) solution, node-open couldn't do more than that ╮(╯_╰)╭
> - I think I would give "2ms" to trade for node-open solution, for better compatibility in the future on windows platform :)

Here's my statement corrected. I'd predict that

- The change would increase boot time by up to a barely noticeable (by anyone's standards) amount for virtually all users.
- The change would otherwise not harm any user's experience.
- The change would improve the preferences experience for a significant amount of users.

amio:
@szhu Really thank you for pointing out these mistakes I have made, and your valuable advice!

When I said "For me, I use ..." I was meant to make an example, I didn't realize those words make the statement more personally, less objective. My bad.

About the smiley, English is not my native language, I'm not so confident about whether I choose words proper to express my thoughts, especially when in arguments. So I use smiley as cover, in case I said something accidentally offensive, I want people know I'm friendly at least. While, it turns out I use it too much, making my expression even wired. I should practice more.

Appreciate these kindly reply, thank you!

kribblo:
@amio : Since @szhu had the patience to dish out some communication advice, I'll add some generic technical advice that pertains to this discussion.

* The idea that lines of code below tens or hundreds of thousands or importing modules add any noticeable overhead has not been true for at least 10 or 15 years (well maybe about 5 for javascript in browsers). It *was* true, it's *not* anymore, and you're gonna have this discussion *a lot* if you keep this view.
  * Already touched upon above, but *measure, measure, measure* if you're gonna do optimization. Gut feelings are wrong.
  * Science tells us that 100ms is instant to the brain and 1 second is not breaking flow. Above that, start looking at speeding up things. 10 seconds is when people lose track and go on Facebook for 20 minutes.
* That something is used infrequently is not no reason to not do the best possible solution; it may be a reason to not spend many hours on it, but not to do a bad solution.
  * In this specific case, users probably want to change settings among the first things they do; it's great if it's virtually guaranteed to work on any given computer, hence a solution like node-open.
  * Doubly so if it's users that don't know what `$EDITOR` does or how to close vim.
* Start-up time is rarely important to apps that stay open for a long time. What's even a second of startup time gonna do to hours or days of usage afterwards?
  * I've probably lost at least a minute configuring `$EDITOR` before I could open the settings, how many startup times is that? (Not that anyone should count that way!).

amio:
@kribblo I agreed most of your statement, but not conclusion:

- *Optimization should base on measurement.* totally agree.
- While it's not always for any situation. here is an example of exception:
  - By the time I call this PR, there was a rush of new user trying HyperTerm, a lot of them can't even open the preference, it's really anoying.
  - In that situation, a "worked" solution is more critical than a "perfect" solution.
  - That's why I said I want to make a "quick fix". A quick fix just like @szhu did #267
  - @szhu 's fix is better than mine, and that PR is more clear (without so many disscution, so many ideas to balance) to review & merge. #267 has done exactly what I want.
- Above is why I didn't make measurement: It's a problem big enough to do measurement, and I was meant to "quick" fix it.

Besides, seems you haven't do measurement either, so I don't think your are more "right" than me.

And more clarifition on my view of the `node-open` method:
- I didn't against it at first.
- I **feel** it's a non-trivial change to the project, it **might**(just not sure, maybe 50% to 50%) be not agreed by @rauchg . And, by that time I want to make a "quick" fix as soon as possible, a simple modification (like #267) will get more chance to get merged IMO.
- I'm sorry I didn't express this concern correctly at the begining, instead I used a stupid word "I'm not sure if it fits HyperTerm's philosophy" and go on to explain the 50% part (against the approach) of my view.

So, this is my further explanation, not a "reply".

rauchg:
Thank you all for passionately discussing this. I love the idea of fixing the `$EDITOR` trick in the short term like we have with this PR, and I'm very much open to discussing `node-open` as a way of having more predictability, a better user experience on Windows, etc!

szhu:
@amio good points. I'm going to summarize them to confirm and to make a summary people reading this thread in the future:

- In general, runtime optimization should be based on measurement. We simply don't know how long something will take to run.
- When submitting a quick fix, it is more likely to be merged faster if it does not introduce new dependencies.
- We can have a short-term quick fix and a longer-term fix later. These aren't mutually exclusive.

Also some replies:

- If you re-read the conversation, you originally said that you wanted to not use `node-open` because it would slow down runtime. Now it looks like your main reason for not using `node-open` in this PR is because you want the PR to be merged faster (even if using `node-open` might speed up runtime). So note how runtime isn't the most important thing :)
- It's not taboo to say "I'm not sure if it fits HyperTerm's philosophy"! Even if you're wrong, it's not stupid or offensive to make such a statement.

amio:
@szhu More explanation to your reply (sorry it's so verbose, feel free to ignore, it's only a trivial point):

- As I explained previously:

    > I didn't express this concern correctly at the beginning, instead, I said "I'm not sure if it fits HyperTerm's philosophy" and go on to explain the 50% part (against the approach) of my view."

    The phrase “go on to explain the 50% part (against the approach) of my view.” refers my view is `node-open` may or may not fits, and I explained **(if it gets rejected, 50% chances)** what the reasons could be. That's my statements you saw that against `node-open`.

    I didn't express that concern in time, it's just a feeling back then. The only evidence that I had that concern at the very beginning, is that I've said "I'm not sure if it fits HyperTerm's philosophy", to me I know it's coming from that concern deeply in my mind.

    In my mind the reason chains are:
    - (node-open) may slow down / add deps / or whatever -->
    - **may** not fits HyperTerm's principle (of how to balance these choices) -->
    - (blocking this quick fix get merged asap.)

    Apparently, I didn't express my whole thoughts properly, and even I myself realize this recently, back then, it's just a blurred **feeling**.

and btw, I never mention that I think using `node-open` might even speed up runtime, is there some phrases I gave this impression? I think it maybe a miss-expression too >.<

szhu:
Thanks for the clarification! I think I was confused when you said you were recommending final decisions based on feelings, but it looks like you meant something more like "there is a possibility this might be bad for this reason so we should make sure it's ok." That's a really good thing do! I'm sorry I took until now to understand what you meant.

I'm saying that node-open could potentially speed up runtime. This is because *we don't know* whether it will be faster or slower than the original proposed solution.

amio:
Glad that we got to understand each other on every aspect, finally!
And I've learned a lot English from you @szhu :D

kevinSuttle:
So, what's the resolution here, or is there one?

kevinSuttle:
Ah, I see this now after updating. Thanks!

`bash -c "exec env $EDITOR ~/.hyperterm.js"`

PR #160, opened 2016-07-16
Add a copy of the MIT license

Calinou:
It is good practice to add a copy of the license to the repository (I noticed the license in `package.json`).

Note: I implied copyright goes to Zeit Inc., which is maybe wrong; let me know if this needs to be changed.


rauchg:
Thank you!

Issue #137, opened 2016-07-16
Move config/plugins under a single ~/.hyperterm directory


caleb531:
Hi,

I just discovered Hyperterm today and I'm really liking the idea of a web-based terminal (I use Atom as my code editor of choice). However, reading through the documentation, I noticed that configuring the editor (plugins and all) creates several files in your home directory (*i.e.* `~/.hyperterm.js` and `~/. hyperterm_plugins `). I also found issues on this repo like #10 and #14 which propose a possible custom stylesheet file (as well as the addition of an Extensions API).

I'd prefer not to clutter my home directory with more hidden files, if possible. Therefore, I feel that all of these files should be consolidated under a single `~/.hyperterm` directory, similar to Atom's. It could have the following structure:

```
.hyperterm/
- plugins/
  - ...
- config.js
- styles.css (future)
```
...and whatever else might be added.

I figure to migrate current users to the new hierarchy, some backwards-compatibility code could be added which does the following:

1. Use `.hyperterm` for all configuration/plugins if the directory exists
2. If it does not exist, create a new `.hyperterm` directory and move `~/.hyperterm.js` and `~/.hyperterm_plugins` to their new respective paths

Thoughts? :)
Caleb

mcchrish:
Or XDG directories.

webdesserts:
I would highly recommend using the [XDB Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/index.html#introduction) and create a hyperterm folder under `$XDG_CONFIG_HOME` || `~/.config/`. neovim now uses this instead of the classic `.vimrc` & `.vim/` folder and I would love it if I could manage the config for all of my apps under a similar folder.

Okahyphen:
I've suggested a future-facing breaking change in patch #170 for supporting the XDG Base Directory Spec. I would argue its best to make this change sooner rather than later, becoming completely compliant, and not bother with oddball fall-backs like `$HOME/.hyperterm`.

sindresorhus:
HyperTerm should store it in [`app.getPath('userData')`](http://electron.atom.io/docs/api/app/#appgetpathname). For example on macOS, apps use `~/Library/Application Support` for app data, not the home directory.

OJFord:
> HyperTerm should store it in app.getPath('userData'). For example on macOS, apps use ~/Library/Application Support for app data, not the home directory.

[...]

Issue #14, opened 2016-07-02
Extensions API


rauchg:
Continuing from the "Allow custom stylesheet" thread (#10).

The way I want to do this is directly with the JS extensions API. You would basically register a plugin that registers CSS. 

The sample extension we'll have is one that makes the theme light during the day and dark at night. This will show off adding CSS and also adding a bit of JS logic, so it's a good all-around example.

I'll probably use webpack in the example build step so that you can `require('./something.css')`. People that want to use sass can simply use a sass loader or add another step to get the `.css` from the `.sass`.

rauchg:
As far as how someone register their extensions, I was thinking about having `$HOME/hyperterm.json`.
It would look as follows:

```json
[
  "@hyperterm/preserve-pwd",
  "hypersolar"
]
```

This has a few purposes:
- **Sharability**. "👨: Why does your hyperterm look so cool!? "👩: Here, have my `hyperterm.json`".
- **Configurability and extensibility in one**. If you want to remove functionality, simply remove a core plugin. Plugins admit optional parameters:
  
  ```json
  ["hypersolar", { "useWeatherAPI": true, "zipCode": 94123 }]
  ```
- **Declarative installation**. Kind of like the amazing `.vimrc` based extensions solutions (I use NeoBundle). When you make changes to the file, that triggers the installation or removal of the NPM modules.

rauchg:
As far as the API, I'm going to experiment with the following: plugins simply expose a function that is invoked with the main `<HyperTerm>` component to decorate.

This would give plugins complete control over the customizability of the terminal. They could register `componentDidMount`, add stuff to `render`, etc.

chabou:
Outdated

PR #170, opened 2016-07-16
Support Cross Platform Config Locations

Okahyphen:
This patch adds improved default configuration locations (alongside partial compliance for the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)).

This is a pseudo-breaking change. Existing users will be unaffected, while new users will be none the wiser.

Instead of placing various files and directories directly into the `$HOME` directory (as hinted at in #14), this gives the user more options to unify configuration files in a central place.

The new default config parent directory would be aligned with the result of calling [`app.getPath('appData')`](http://electron.atom.io/docs/api/app/#appgetpathname):
- `%APPDATA%` on Windows
- `$XDG_CONFIG_HOME` or `~/.config` on Linux
- `~/Library/Application Support` on macOS

Within the final directory, a unified directory `HyperTerm` is created. The new configuration file would be `HyperTerm/init.js`.

Plugins go in `HyperTerm/plugins` and `HyperTerm/plugins/local`.

This is following suit with other major pieces of software, such as [Neovim](https://neovim.io/), and addresses issues such as  #137.

If a user has a `~/.hyperterm.js` configuration file, then the legacy locations are used:
- Config: `~/.hyperterm.js`
- Plugins: `~/.hyperterm_plugins` 
- Local Plugins: `~/.hyperterm_plugins/local`

This patch introduces a new file `config-paths.js`, which determines correct set of file paths to use at runtime. New paths can be easily added (legacy and future objects must be kept in parallel), and the rest of the application should remain ignorant to the type of path it is using.

Thoughts?

### DIFF
--- a/config-paths.js
+++ b/config-paths.js
@@ -1,23 +1,23 @@
 const { app } = require('electron');
+const { join } = require('path');
 const { homedir } = require('os');
-const { resolve } = require('path');
 const { existsSync } = require('fs');
 
 const home = homedir();
-const root = resolve(app.getPath('appData'), 'HyperTerm');
-const legacyConfig = resolve(home, './.hyperterm.js');
+const root = app.getPath('userData');
+const legacyConfig = join(home, '.hyperterm.js');
 const legacy = existsSync(legacyConfig);
 
 module.exports = legacy? {
   root: home,
   config: legacyConfig,
-  plugins: resolve(home, './.hyperterm_plugins'),
-  localPlugins: resolve(home, './.hyperterm_plugins', './local'),
-  resolveConfigPath: (...paths) => resolve(home, ...paths)
+  plugins: join(home, '.hyperterm_plugins'),
+  localPlugins: join(home, '.hyperterm_plugins', 'local'),
+  joinConfigPath: (...paths) => join(home, ...paths)
 }:{
-  root: root,
-  config: resolve(root, './init.js'),
-  plugins: resolve(root, './plugins'),
-  localPlugins: resolve(root, './plugins', './local'),
-  resolveConfigPath: (...paths) => resolve(root, ...paths)
+  root,
+  config: join(root, 'init.js'),
+  plugins: join(root, 'plugins'),
+  localPlugins: join(root, 'plugins', 'local'),
+  joinConfigPath: (...paths) => join(root, ...paths)
 };


sindresorhus:
I have a module for this if of any help: https://github.com/sindresorhus/xdg-basedir

rauchg:
I definitely would love to comply for users who care about XDG (as made evident by the many likes to your post). Thank you for the work that you put into this.

That said, I wouldn't like to break the very obvious expectation of `~/.hyperterm.js` that's already been widely documented and talked about, specially for users who've never heard of XDG.

How about if `XDG_CONFIG_HOME` is set, we put `hyperterm.js` and `hyperterm_plugins` in it? That way it becomes a one line change, where we just add conditionally another argument to `resolve`.

Thanks!

rauchg:
To summarize: I would like to avoid breaking changes / expectations / documentation.

Okahyphen:
What happens when you require more filesystem structure for various features? Is littering  the user's `$HOME` directory really the way to go about this?

Now is the time to make these kinds of changes, while the project is still teething, because it will only get harder as you create more expectations, and write more documentation.

@sindresorhus has pointed out in #137 that Electron provides `app.getPath('userData')`, which actually makes this even easier to support. Especially when you start to consider supporting Windows.

Okahyphen:
Please note I've rebased this patch rather significantly. Please give the original post another read.

Thanks.

rauchg:
This is evolving really nicely @Okahyphen. Strongly considering it.

config-paths.js
```diff
@@ -0,0 +1,23 @@
+const { app } = require('electron');
+const { homedir } = require('os');
+const { resolve } = require('path');
+const { existsSync } = require('fs');
+
+const home = homedir();
+const root = resolve(app.getPath('appData'), 'HyperTerm');
```
sindresorhus: Nitpick, but `appData` => `userData` and you could simplify it to: `const root = app.getPath('userData');`

config-paths.js
```diff
@@ -0,0 +1,23 @@
...
+const legacyConfig = resolve(home, './.hyperterm.js');
```
sindresorhus: Another nitpick, prepending `./` to all the paths are moot. `const legacyConfig = resolve(home, '.hyperterm.js');` would be better. `path.join()` would be a better option here, since `home` is already an absolute path, so you can just join them, instead of resolving.

config-paths.js
```diff
@@ -0,0 +1,23 @@
...
+const legacy = existsSync(legacyConfig);
+
+module.exports = legacy? {
+  root: home,
+  config: legacyConfig,
+  plugins: resolve(home, './.hyperterm_plugins'),
+  localPlugins: resolve(home, './.hyperterm_plugins', './local'),
+  resolveConfigPath: (...paths) => resolve(home, ...paths)
+}:{
+  root: root,
```
sindresorhus: `root: root,` => `root,`

config-paths.js
Okahyphen: Interestingly, I tried that in dev and the application was always reporting itself as `Electron` - despite the `name` and `productName` fields in `package.json`. This was my work around, albeit not as elegant.

Is this a problem with my environment? My experience with Electron is thin.

sindresorhus: @Okahyphen That app name is always `Electron` during dev as it's the packager that changes the app name in Info.plist to what's defined in `productName` in package.json. Just look at the name in the top left of the menubar when running it with `npm test`, it says Electron, not HyperTerm. I would personally consider that a feature, as you don't mess up your own config when developing on HyperTerm.

Okahyphen: Ah, that makes sense. Will update.

config-paths.js
Okahyphen: Noted!

config-paths.js
Okahyphen: Good catch. We should be using joins on the rest of the paths as well, since XDG paths should be absolute as well, correct?

sindresorhus: Yup

Okahyphen:
Thanks to @sindresorhus for the code review. I've added a polishing commit to this, which should be an improvement.

OJFord:
Wagh! Please listen to `XDG_CONFIG_HOME` if it exists, rather than arbitrarily deciding that macOS users don't get a choice!

Or even better, allow a `HYPERTERM_CONFIG_HOME` to take precedence over platform methods.

octref:
@rauchg Any problem with this PR? If not can you consider merging it?

config-paths.js
```diff
@@ -0,0 +1,23 @@
+const { app } = require('electron');
+const { join } = require('path');
+const { homedir } = require('os');
+const { existsSync } = require('fs');
+
+const home = homedir();
+const root = app.getPath('userData');
```
OJFord:
```js
const root = process.env.XDG_CONFIG_HOME !== undefined ? join(process.env.XDG_CONFIG_HOME, 'hyper') : app.getPath('userData')
```
for macOS users' sake please!

jskeate:
I'd really like to see this merged, with a couple tweaks. If nobody else wants to, I can fix it up

config-paths.js
```diff
@@ -0,0 +1,23 @@
...
+  plugins: join(home, '.hyperterm_plugins'),
+  localPlugins: join(home, '.hyperterm_plugins', 'local'),
+  joinConfigPath: (...paths) => join(home, ...paths)
+}:{
+  root,
+  config: join(root, 'init.js'),
+  plugins: join(root, 'plugins'),
+  localPlugins: join(root, 'plugins', 'local'),
+  joinConfigPath: (...paths) => join(root, ...paths)
+};
```
jskeate: I think this is backwards. We should prioritize `$XDG_CONFIG_HOME`/`app.getPath('userData')` over the "legacy" path

albinekb: But we need to make a smooth transition for all our existing users that may have `XDG_CONFIG_HOME` set, but their config is at ~

skeate: What I mean is, check if the files exist in those places. If they have `$XDG_CONFIG_HOME` set but don't have a config for hyper there, then it just goes to the next location

OJFord: Note that `root` is defined above to use an Electron method that does not return `$XDG_CONFIG_HOME` on macOS @skeate. I know a lot of folks will really appreciate it if you can fix that up too.

caesar:
Any news with this PR? Would be really great to see it merged... Can we do anything to help?

CodeTheory:
Closing in favour of #1595.

Very sorry about the amount of time we left this open. We really appreciate the work that went into this though.

PR #171, opened 2016-07-16
feature: prevent dropped files, use path

eljefedelrodeodeljefe:
When file is dragged into context, do not navigate to file (which could be regarded a bug, as is) instead use path to file in current line (as feature)

Fixes: https://github.com/zeit/hyperterm/issues/110
Make use of https://github.com/electron/electron/pull/931

### DIFF
--- a/index.js
+++ b/index.js
@@ -55,6 +55,15 @@ app.on('ready', () => {
     });
     winCount++;
     win.loadURL(url);
+    // prevent file drops to navigate the window. Instead handle explicitly
+    // file drops as string to location of file.
+    win.webContents.on('will-navigate', (event, url) => {
+      event.preventDefault();
+      const path = url.split('file://'); // if has file:// prefix...
+      // ...send chopped url. The fallback is meant as contigency, for
+      // edge cases the author couldn't anticipate. Remove if needed.
+      rpc.emit('dropped-file', path.length > 1 ? path[1] : path[0]);
+    });
 
     const rpc = createRPC(win);
     const sessions = new Map();
--- a/app/lib/components/term.js
+++ b/app/lib/components/term.js
@@ -1,6 +1,7 @@
 /* global Blob,URL,requestAnimationFrame */
 import React from 'react';
 import hterm from '../hterm';
+import rpc from '../rpc';
 import Component from '../component';
 
 export default class Term extends Component {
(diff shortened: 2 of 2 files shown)


rauchg:
I suggest the following: instead of introducing the side effect directly in the component, dispatch an action whose name indicates what's happening.

In this case, it would be something like `dispatch(draggedFile)`.

That action can then have an effect of dispatching another action: for this new default behavior, a PTY write. 

For inspiration, look at what we do when we receive a data action. The incoming data from the pty can result in an actual write, or the navigation to that URL:

https://github.com/zeit/hyperterm/blob/master/app/lib/actions/sessions.js#L63

Similarly, in this case the default effect would be `SESSION_PTY_DATA`, but if we do it this way, plugins can later capture the original action and invoke our effect and potential others! For example, they could render the image in the browser, or something like that.

Furthermore, with this we maintain the components as simple as possible.

rauchg:
Thanks a lot for the great idea. This is something that should ship in the core for sure.

eljefedelrodeodeljefe:
Thanks for the hints. Will apply latest Monday.

developit:
@rauchg @eljefedelrodeodeljefe - this was actually the very first thing that happened to me after I downloaded and ran Hyperterm! I typed `cd` and dropped a folder from Finder into the window, but Electron just navigated to that directory (blank screen, may have been a directory listing with black text on black or something). Hope it gets merged!

rauchg:
@developit will definitely merge this

index.js
```diff
@@ -55,6 +55,15 @@ app.on('ready', () => {
     });
     winCount++;
     win.loadURL(url);
+    // prevent file drops to navigate the window. Instead handle explicitly
+    // file drops as string to location of file.
+    win.webContents.on('will-navigate', (event, url) => {
+      event.preventDefault();
+      const path = url.split('file://'); // if has file:// prefix...
```
TooTallNate: I'd recommend to use `url.parse()` here, check the `parsed.protocol === 'file:'`, and then use [`file-uri-to-path`](https://github.com/TooTallNate/file-uri-to-path) module for better Windows support.

MrRio:
Hi there, 

I've merged this with the latest master branch and tested it out. Looking good so far, there's a few issues:

- [ ] When dragging a folder or file to the window, it adds the text to the terminal, but if you press enter it doesn't actually execute/run it. 
- [ ] Files containing spaces show up as %20

Let me know if you need any help, @lordgiotto said he'd be willing to lend a hand too.

Cheers,
James

screen shot 2016-08-02 at 00 39 43

lordgiotto:
Can be closed since #559 was merged.

albinekb:
Closing due to #559 being merged with the same functionality

pammacdotnet:
Drag and drop between Finder version 10.14.4 and Hyper 3.0.0-canary.8 (canary) doesn't work (no path is printed neither added to the prompt).

Issue #145, opened 2016-07-16
NPM Error Message not sure what it means


sandysaders:
I keep getting the error message below and I'm not sure how to fix it. I have `nvm` on my computer and can run npm commands. I tried running npm commands in `hyperterm` and they worked so not sure how to fix this. (not to familiar with terminal things.) 

<img width="322" alt="screen shot 2016-07-15 at 9 23 07 pm" src="https://cloud.githubusercontent.com/assets/19610736/16891974/5d5edf50-4ad2-11e6-9e9a-6cbae640860a.png">

montogeek:
Duplicate of #106 and #89

PR #172, opened 2016-07-16
Get `npm install` to work

### LINKED ISSUE #106: We could not find the 'npm' command. Make sure it's in $PATH
I get this notification from subject.

but 

```
$ which npm
/Users/david/n/bin/npm
```

works...

Not sure if it's related, when I add "hyperpower" plugin to `.hyperterm.js` and save it, nothing really happens and 

```
$ ls .hyperterm_plugins/
local        package.json
```

Any suggestions ?

Should "hyperpower" plugin install automatically after saving the file and is this not happening because of the problem about not finding 'npm' command supposedly?

marcbachmann:
This fixes  #106 and #145

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "gaze": "1.1.0",
     "mkdirp": "0.5.1",
     "ms": "0.7.1",
+    "shell-env": "0.1.2",
     "uid2": "0.0.3"
   },
   "devDependencies": {
--- a/plugins.js
+++ b/plugins.js
@@ -8,6 +8,7 @@ const { exec } = require('child_process');
 const Config = require('electron-config');
 const ms = require('ms');
 const notify = require('./notify');
+const shellEnv = require('shell-env');
 
 // local storage
 const cache = new Config();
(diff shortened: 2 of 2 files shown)


montogeek:
@marcbachmann How did you tested it? I just compiled a new version with your changes and the problem persists

montogeek:
Also, Could you please fix the dependency version?

package.json
```diff
@@ -14,7 +14,8 @@
     "gaze": "1.1.0",
     "mkdirp": "0.5.1",
     "ms": "0.7.1",
-    "uid2": "0.0.3"
+    "uid2": "0.0.3",
+    "user-env": "^2.0.0"
```
rauchg: Can you pin this please?

rauchg:
What an elegant solution! Thank you @marcbachmann. Only one minor nit-pick.

marcbachmann:
Done. Rebased and version fixed.

montogeek:
Marvelous

marcbachmann:
Another solution would be to install `npm` as module and use the api.
That way it's also windows-compatible.

```js
const npm = require('npm')
function install (cb) {
  npm.load({}, function (err) {
    if (err) return cb(err)
    npm.commands.prune([], function (err) {
      if (err) return cb(err)
      npm.commands.install(['--production'], cb)
    })
  })
}
```

rauchg:
@marcbachmann I could be wrong but I've heard that the npm team is not recommending API usage. Is this the case @iarna @othiym23 ?

leo:
@rauchg [It is.](https://www.youtube.com/watch?v=vLcOkHhEt1g&feature=youtu.be&t=21m1s) Like @othiym23 suggests it, we should a child process instead.

marcbachmann:
I've replaced `user-env` with `shell-env` because it also has windows support and has an async function. I've also removed the prefix that wasn't needed anymore.

@rauchg this is ready to merge.

rauchg:
<3

mjgyesme:
woo hoo! nvm users rejoice!

iarna:
Indeed! We would prefer that you not use npm's internal API, we'd like to keep it subject to change. (Mostly because it's terrible, and we'd like to have the freedom to make it less terrible.)

Calling npm is the correct and supported way of integrating with npm.

If you find yourself having trouble integrating with it (eg it's output is insufficiently uniform, say) then I'd be happy to work with you all in patching it to correct that sort of thing.

rauchg:
Thanks a lot for letting us know @iarna

PR #174, opened 2016-07-16
Provide hooks to open a new tab to the same directory

hharnisc:
- Allows `cwd` parameter to be changed when creating a new directory
- Exposes the `pid` of the newly created session in the dispatch event

The plugin to try this out is called [hypercwd](https://github.com/hharnisc/hypercwd) it's published to NPM under the name `hypercwd`

This is related to #4


rauchg:
This is beyond great. Thank you so much @hharnisc

PR #176, opened 2016-07-16
Fix maximizing behaviour

CompuIves:
Currently maximizing/minimizing isn't working because of a function call to `this.rpc.emit` while `this.rpc` is null. 

This PR fixes this with also fixing some minor bugs with minimizing/maximizing.


rauchg:
Thanks for the fix. I think it would be nice if the header didn't emit directly. It'd be cool if you could handle it like we handle other effects.

For example, when you close a tab, we dispatch an action:
https://github.com/zeit/hyperterm/blob/master/app/lib/containers/header.js#L35-L45

Then actions can have side effects in `effect`: dispatch other actions or `rpc` calls.
https://github.com/zeit/hyperterm/blob/master/app/lib/actions/ui.js#L20-L35

More info here: https://hyperterm.org/#actions-and-effects

CompuIves:
Ah, I see. I will update the PR!

CompuIves:
Changed it, I'm still uncertain about one thing. I'm currently using `toggle-maximize` because in that way I can read from `BrowserWindow` if the window is maximized or not and handle according to that. I think this is not the cleanest way to do it, since it's not pure. We don't know what will happen if we run the function with the same arguments and we are relying on external logic to determine this.
A solution would be to go back to `maximize` and `unmaximize` with the tradeoff that we have to assume the user starts the terminal unmaximized. I'm more in favor of `maximize` and `unmaximize` but I first want to know your opinion of this.

CompuIves:
Updated the PR, decided to store window maximized state in the store and assume we're starting with a maximized state of `false`.

MrRio:
This works really well, can you resolve the conflicts?

CompuIves:
Resolved!

PR #179, opened 2016-07-16
Fix hypersolar plugin example, it does not exist

montogeek:


mattvagni:
👍

rauchg:
It doesn't exist _yet_ ;)

PR #180, opened 2016-07-16
Fix onWindow hook

dfrankland:
The `onWindow` method of plugins.js has no second parameter of `app`.
Switch to the `BrowserWindow` passed as the first parameter.


rauchg:
Thanks you @dfrankland!

PR #181, opened 2016-07-16
fix plugins.onWindows()

tomru:
Trying to write a plugin, I noticed `plugin.onWindow()` will never receive a window instance. `plugins.onWindows(win, app)` passes along `app` instead of `win`.


tomru:
lol, @dfrankland beat me in https://github.com/zeit/hyperterm/pull/180 

closing this

PR #185, opened 2016-07-16
menu improvements

sindresorhus:
- Add missing separator after `About`.
- Use `role` instead of deprecated `selector` for menu items
- Add `undo` and `redo` menu items. Both Terminal and iTerm have these.
- Remove indent of the report issue template. The indent ended up in the issue body, which looks weird.
- Add separator after `New Tab`. This is how iTerm does it.


rauchg:
Thanks @sindresorhus <3

PR #188, opened 2016-07-17
Example React Native implementation of hyperterm header

necolas:
Hey,

Here's a rough example of how you could use React Native to build the terminal's header. There are a few warnings in the console about style properties React Native doesn't currently know about.

You'll notice that CSS pseudo-classes (e.g. `:hover`) aren't supported. However, it's quite common to track state like hover in JavaScript anyway (as you do in `Tab`).

Something else worth mentioning is that `View` is the main building block and has a column-based flexbox layout by default. It won't accept any text or font styles; those have to be applied to a `Text` element. A helpful pattern is to create a component that encapsulates the typography concerns, e.g., `<AppText color='white' size='normal'>bash</AppText>`.

User theming could be done with JS objects. Since you'd be working with objects and not strings, you can better control exactly what custom styles you apply to the components. For example, the color but not position of text nodes could be customizable:

```
exports.decorateConfig = (config) => {
  return Object.assign({}, config, {
    borderColor: 'yellow',
    cursorColor: 'yellow',
    css: Object.assign({}, config.css, {
      tabText: {
        color: 'yellow', // this style is a valid override
        position: 'fixed' // this style is not, and you ignore it
      },
      tabsTitle: {
        color: 'yellow'
      }
    })
  });
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,9 +19,7 @@ modules are built with `electron`, run:
 $ ./scripts/install.sh
 ```
 
-Then, you want to make sure `app/dist` is populated. I recommend
-running `webpack` with `--watch` so that any changes you make
-to the app are detected.
+Then, you want to make sure `app/dist` is populated.
 
 ```bash
 $ cd app/
--- a/app/lib/components/header.js
+++ b/app/lib/components/header.js
@@ -1,16 +1,20 @@
-import React from 'react';
+import React, { Component, PropTypes } from 'react';
 import Tabs_ from './tabs';
-import Component from '../component';
 import { decorate, getTabsProps } from '../utils/plugins';
+import { StyleSheet, View } from 'react-native';
+import { shouldComponentUpdate } from 'react-addons-pure-render-mixin';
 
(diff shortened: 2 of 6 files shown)


dimorphic:
Nice initiative, but unfortunately you won't be able to spawn [child processes](https://nodejs.org/api/child_process.html#child_process_child_process) (pseudo terminal our case) on mobile devices and have this beauty of a terminal. Building the app via React Native or any other hybrid way won't work (for now).

Electron has Node.js and Chromium under the hood, that's where the Ju Ju magic comes from. :)

:+1:

dimorphic:
I was wrong, there is hope! https://github.com/jxcore/jxcore-cordova

Unfortunately seems they halted active development, but hey open source! <3 - [blog post](http://www.nubisa.com/nubisa-halting-active-development-on-jxcore-platform/)

necolas:
Huh?

necolas:
Chatted with @rauchg. Closing this example; will leave to him to port the existing code if interested.

PR #190, opened 2016-07-17
Enable preliminary Windows support

kwonoj:
This PR consists couple of changes to enable preliminary support on Windows platform, by
- replace `child_pty` to `ptyw.js`
  : original module `pty` currently have known issues with windows build (https://github.com/chjj/pty.js/issues/131), introduced `ptyw` until issue's fully resolved.
- add initial configuration to `appveyor` to trigger windows build.
  : currently only install dependenceis / bulid native modules, does not packaging it yet.
  : resues existing install script (`install.sh`), so anyone want to build locally may require *nix shell environment still.

As this PR replaces one of core module, I agree this can be risky change especially there isn't enough test coverage. Please feel freely reject if this need not to be considered at this moment.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,58 @@
+#---------------------------------#
+#      general configuration      #
+#---------------------------------#
+
+# version format
+version: 0.0.{build}-{branch}
+
+#---------------------------------#
+#    environment configuration    #
+#---------------------------------#
+
+# Build worker image (VM template)
+image: Visual Studio 2015
+
+# scripts that are called at very beginning, before repo cloning
+init:
+  # configure node-gyp environment https://github.com/nodejs/node-gyp#installation
+  - npm config set msvs_version 2015
+  - 'echo System architecture: %PLATFORM%'
+
+# environment variables
+environment:
+  nodejs_version: "6.2.2"
+
+# this is how to allow failing jobs in the matrix
+matrix:
+  fast_finish: true                     # set this flag to immediately finish build once one of the jobs fails.
+
+# build cache to preserve files/folders between builds
+cache:
+  - node_modules                        # local npm modules
+
+# scripts that run after cloning repository
+install:
+  - ps: Install-Product node $env:nodejs_version
+  - npm -g install npm@next             # install latest npm for node-gyp build issue https://github.com/nodejs/node-gyp/issues/972
+
+#---------------------------------#
+#       build configuration       #
+#---------------------------------#
+
+platform:
+  - x64
+
+# scripts to run before build
+before_build:
+  - sh scripts/install.sh
+  - npm run lint
+
+# scripts to run *after* solution is built and *before* automatic packaging occurs (web apps, NuGet packages, Azure Cloud Services)
+before_package:
+
+# scripts to run after build
+after_build:
+
+# to run your custom scripts instead of automatic MSBuild
+build_script:
+  - echo build native module completed   # avoid to trigger msbuild
--- a/package.json
+++ b/package.json
@@ -6,7 +6,6 @@
   "repository": "zeit/hyperterm",
   "description": "HTML/JS/CSS Terminal",
   "dependencies": {
-    "child_pty": "3.0.1",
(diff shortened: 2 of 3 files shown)


rauchg:
@gottox is `child_pty` not working on Windows? Based on a quick look of the code of both `pty.js` and `child_pty`, I would like to retain the latter.

Awesome stuff so far!

Gottox:
child_pty does only work on unix systems atm. There were some experiments with [winpty](https://github.com/rprichard/winpty) but as I don't own a windows system to test it hasn't come that far yet.

Gottox:
The main problem with winpty is that I refuse to ship winpty in binary.

winpty itself has insane build deps, requiring both, a [cygwin and a MinGW](https://github.com/rprichard/winpty#cygwinmsys-adapter-winptyexe) toolchain.

So for now I decided to keep it simple and only support unix systems.

Gottox:
I'm currently playing with a windows VM and nodejs. I let you know if I get it up and running.

ingro:
@Gottox looking forward to this!

Gottox:
See ongoing process here: https://github.com/Gottox/child_pty/tree/windows

mablae:
I am just totally thrilled since I if the windows build hyperterm will be able to attach to the terminal panel
inside IntelliJ IDEA IDE ?

Gottox:
This will take some time, I guess. winpty is quite different from the unix API... Especially I can't reuse child_process on windows, so a lot of stuff needs to be written. In order to get hyperterm on windows out fast, I guess switching back to pty.js is a better solution.

rauchg:
Thanks for the update @Gottox. One thing I'd like to check before we temporarily switch back to this is: chatting with @Tyriar he shared some concerns about `pty.js` leaking memory when bundled with Electron. Therefore what we need in order to merge this is to answer:

- Is this still the case?
- Is it the case for pty*w*.js?
- Are there any other differences with `child_pty` we're not contemplating?

Tyriar:
The main issue with pty.js is that it must by run in a child process on OS X or it maxes out the CPU, see https://github.com/electron/electron/issues/38

The current situation with vscode is that we do just that, but we don't use chjj/pty.js, instead we my personal fork of it Tyriar/pty.js which is a fork itself:

```
Tyriar/pty.js -> platformio/pty.js -> jeremyramin/pty.js -> chjj/pty.js
```

The purpose of the Tyriar/ fork is to gets it building on new node (which ptyw.js seems to do) and working with our build system (by removing prebuilt assets). I'm passively looking around for a maintained alternative that uses a newer version of winpty.

Gottox:
@rauchg

pty.js has some issues with binding of functions. So `this` points to unexpected objects. Other than that, I can't say much about pty.js. :)

mablae:
Hey, 
instead of having two different pty libraries from now on, it would be better to solve the issues in child_pty upstream and make it usable with windows. These issues should be solved upstream. 

There are alreay many forks of child_pty. Also some guys work on windows. 
The debt a second pty would introduce in the project should be considered.

Gottox:
Where are these windows forks your talking about? :)

Adam-Meisen:
If there's any grunt work a newcomer to the project can do to help this along, let me know.

tyrbo:
I'd love to help, but I'm not sure what's needed here.

Gottox:
@tyrbo child_pty needs windows support. There's a [window branch](https://github.com/rprichard/winpty) in the child_pty repo. the branch includes the winpty project, but I haven't figured out how to intergrate it into child_pty. The hardest part is that winpty _needs_ a custom exec method, so we cannot use child_process directly.

iamstarkov:
can anybody share status of this issue/pr?

rvion:
@iamstarkov I think windows support is coming here: https://github.com/zeit/hyper/pull/946

iamstarkov:
@rvion thanks.

this PR can be closed then? /cc @kwonoj

kwonoj:
Yes, I'm closing this on behalf of #946. Wish this could make it, but those PR seems much more ready-to-ship state.

Issue #81, opened 2016-07-12
Ability to specify which shell to use


mike-engel:
For various reasons, I have my login shell set to `/bin/bash`, but prefer to use `/usr/local/bin/fish` as my default shell. I also have a couple profiles setup in iterm right now for bash, zsh, and fish in case I need to switch to a different shell for testing or debugging.

It would be nice to, at the very least, specify which shell to run for new tabs/windows. It would be even better to be able to setup multiple profiles for quick switching.

rauchg:
Makes total sense. This shall be an option in `~/.hyperterm.js`

mike-engel:
So I'd like to make a PR for this, but I'm having a lot of trouble finding a way to set the shell/command to run when the session starts. Do you know of a good way to do this?

rauchg:
@mike-engel the shell is loaded in the `Session` class when we initialize the `pty`. You can add an extra parameter to the constructor to receive the shell, and then inspect the configuration in `index.js` to get the shell.

To get the configuration after plugins alter it, you can invoke `plugins.getDecoratedConfig()`

spudly:
So I configured the `shell` option added by the pull request, but it has no effect for me.

Does fish need to be in `/etc/shells` for this config option to work?

mike-engel:
@spudly I don't think a version of hyperterm has been release yet with this in it. Are you using a release version, or a version from `master`?

spudly:
@mike-engel, oops, I see that you're right.

I confused myself because I initially installed hyperterm by cloning the repo, so my initial hyperterm config (with `shell` key set to '') was generated by code on master, which made me think the feature was released. Now that I'm running a released version, I don't have the feature anymore.

I'll just run fish manually until this is released.

mike-engel:
@spudly Glad it was something simple. When it is released, however, it will be the path to the shell executable. i.e. `/usr/local/bin/fish` or `/bin/bash`, etc.

mkscrg:
FYI https://hyperterm.org/ lists the `"shell"` config, which doesn't currently work.

rauchg:
oops! that shouldn't have been deployed.

rauchg:
@mkscrg the good thing: we're landing that next. `0.7.0` shipping next

dotlouis:
Hi,
I just found out that this feature was shipped and the `shell` property is working in the `0.7.0`release. However I had to add it manually in my `hyperterm.js`.
There was no *default value* nor any *shell* entry.
[...]

PR #192, opened 2016-07-17
Add an option for a non-login shell to be run

mike-engel:
This adds a new key to the config object, called `shell`. This allows the user to specify a custom shell when running hyperterm. By default, however, the user's login shell will be used

Solves #81

### DIFF
--- a/config-default.js
+++ b/config-default.js
@@ -46,7 +46,11 @@ module.exports = {
       '#cc00ff',
       '#00ffff',
       '#ffffff'
-    ]
+    ],
+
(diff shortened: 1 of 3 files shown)


rauchg:
I think the default value should be `''` for auto-detection, and we can continue to rely on the `default-shell` npm package.

rauchg:
(and we should probably document that subtlety in the comments)
Amazing work again. Thank you!

mike-engel:
@rauchg oh my, I *had* the default as `''` but forgot to remove `/bin/bash` from my testing 😅 Will update here in a bit.

PR #193, opened 2016-07-17
Allow the `color` config to be an object

mike-engel:
Allows the `colors` config object to have nice names for configuring colors (as opposed to an array). It *only* covers the ANSI 16 as named colors, but allows for an array to
be used if the full color palette wants to be overridden.

### DIFF
--- a/app/lib/reducers/ui.js
+++ b/app/lib/reducers/ui.js
@@ -11,6 +11,7 @@ import {
   SESSION_SET_ACTIVE
 } from '../constants/sessions';
 import { UPDATE_AVAILABLE } from '../constants/updater';
+import { values } from '../utils/object';
 
 // TODO: populate `config-default.js` from this :)
 const initial = Immutable({
@@ -111,8 +112,18 @@ const reducer = (state = initial, action) => {
         }
 
         if (null != config.colors) {
-          if (JSON.stringify(state.colors) !== JSON.stringify(config.colors)) {
-            ret.colors = config.colors;
+          if (Array.isArray(config.colors)) {
+            const stateColors = Array.isArray(state.colors)
+              ? state.colors
+              : values(state.colors);
+
+            if (stateColors.toString() !== config.colors.toString()) {
+              ret.colors = config.colors;
+            }
+          } else {
+            if (JSON.stringify(state.colors) !== JSON.stringify(config.colors) {
+              ret.colors = config.colors;
+            }
           }
         }
 
--- a/app/lib/reducers/ui.js
+++ b/app/lib/reducers/ui.js
@@ -27,24 +27,24 @@ const initial = Immutable({
   termCSS: '',
   openAt: {},
   resizeAt: 0,
-  colors: [
-    '#000000',
-    '#ff0000',
-    '#33ff00',
-    '#ffff00',
-    '#0066ff',
-    '#cc00ff',
-    '#00ffff',
-    '#d0d0d0',
-    '#808080',
-    '#ff0000',
-    '#33ff00',
-    '#ffff00',
-    '#0066ff',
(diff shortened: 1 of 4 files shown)


rauchg:
This is amazing. Couple things missing:

- The default state should now probably be an object as well (https://github.com/zeit/hyperterm/blob/master/app/lib/reducers/ui.js#L30)
- We probably need to improve the change detection algorithm here to work with both formats? https://github.com/zeit/hyperterm/blob/master/app/lib/reducers/ui.js#L114
- It'd be awesome (not a blocker, but awesome) to document the format here https://github.com/zeit/hyperterm-website/blob/master/index.html#L533-L535

mike-engel:
@rauchg Thanks for the feedback. To answer your questions:

1. That change should have [already been in place](https://github.com/zeit/hyperterm/pull/193/files#diff-8370d1e2f4cda838221aac6e897a871eR30) with this PR. Let me know if I missed another place.
2. `JSON.stringify` should be sufficient for both objects and arrays:
```javascirpt
JSON.stringify({ a: 'hey', b: 'hello' })
// -> "{"a":"hey","b":"hello"}"
JSON.stringify(['a', 'b', 'b'])
// -> "["a","b","b"]"
```
3. I'll go ahead and make a PR for the website as well 😄

app/lib/reducers/ui.js
```diff
@@ -111,7 +111,7 @@ const reducer = (state = initial, action) => {
         }
 
         if (null != config.colors) {
-          if (state.colors.toString() !== config.colors.toString()) {
+          if (JSON.stringify(state.colors) !== JSON.stringify(config.colors) {
```
mauro-oto: @mike-engel I think there's a typo here, a missing parenthesis. Should be 
```javascript
if (JSON.stringify(state.colors) !== JSON.stringify(config.colors)) {
```
instead.

mike-engel: Looks like I didn't get that change pushed up or something. Nice catch!

app/lib/reducers/ui.js
```diff
@@ -111,7 +111,7 @@ const reducer = (state = initial, action) => {
         }
 
         if (null != config.colors) {
-          if (state.colors.toString() !== config.colors.toString()) {
+          if (JSON.stringify(state.colors) !== JSON.stringify(config.colors)) {
```
rauchg: My concern specifically was for the people who will have the outdated format in their configs, and will be compared to a different new initial state?

mike-engel: Ah yeah, ok I think I understand now. So you're concern is that if they have an array, but we're checking for an object, it will always be different, right? The check here will work, but if they have an array in their config still, it will always run the block inside here, since they will never be the same. If that's the case, I can add some logic right after work today to check if it's an array or an object. Let me know if I'm not reading that right 😄

rauchg: Yep you are! Thank you. The joys of backwards compatibility 🎉

mike-engel: It's actually not backwards compatibility since an array is still valid if the 6 x 6 cubes and grey scale need to be specified 😉 . Either way, I added some code to handle an array vs. an object. Let me know if I missed anything!

rauchg:
This is sweet!

app/lib/utils/colors.js
```diff
@@ -0,0 +1,31 @@
...
+
+export function getColorList (colors) {
+  // For backwards compatibility, return early if it's already an array
+  if (Array.isArray(colors)) {
+    return colors;
+  }
+
+  return colorList.map((colorName) => {
+    return colors[colorName]
+  })
```
TooTallNate: style: missing semicolon

PR #194, opened 2016-07-17
Remember window size and position on restart. Fix #27

### LINKED ISSUE #27: Hyperterm should remember the window size upon restart
Currently, Hyperterm always start with a set size.

## Reproduce

1. Open Hyperterm
2. Change the window size
3. Restart Hyperterm
4. See how the window still is the default size

mawie81:
This PR adds support for remembering and restoring the window size and positions for all windows on restart.
Fixes #27

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -7,6 +7,7 @@ const isDev = require('electron-is-dev');
 const AutoUpdater = require('./auto-updater');
 const toHex = require('convert-css-color-name-to-hex');
 const notify = require('./notify');
+const windowStateKeeper = require('electron-window-state');
 
 app.commandLine.appendSwitch('js-flags', '--harmony');
 
@@ -48,9 +49,15 @@ app.on('ready', () => {
 
     const [width, height] = cfg.windowSize || [540, 380];
 
+    const windowState = windowStateKeeper({
+      file: 'window-state-' + (windowSet.size + 1) + '.json'
+    });
+
     const browserDefaults = {
       width,
       height,
+      x: windowState.x,
+      y: windowState.y,
       minHeight: 190,
       minWidth: 370,
       titleBarStyle: 'hidden-inset',
@@ -67,6 +74,7 @@ app.on('ready', () => {
     const win = new BrowserWindow(browserOptions);
 
     windowSet.add(win);
+    windowState.manage(win);
     win.loadURL(url);
 
     const rpc = createRPC(win);
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "hterm-umdjs": "1.1.3",
     "json-loader": "0.5.4",
     "mousetrap": "1.6.0",
+    "electron-window-state": "3.0.3",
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -44,10 +45,18 @@ let winCount = 0;
 
 app.on('ready', () => {
   function createWindow (fn) {
+    let windowState = windowStateKeeper({
+      file: 'window-state-' + (winCount + 1) + '.json',
+      defaultWidth: 540,
```
bransorem: Can we also allow this to be overridden by `~/.hyperterm.js`?

rauchg:
This looks awesome! What's the advantage of `electron-window-state` as opposed to storing it for example in the `electron-config` we already have?

rauchg:
(other than automatically listening to the window events and syncing the state)

mawie81:
Well you already gave the answer 😉 
Used the module because everything is already in place and code changes are minimal.
What I think I can do in the future is using ```electron-config``` right inside ```electron-window-state``` 👍

ellerbrock:
@mawie81 nice you done this already! i just checked the pr's otherwise i wanted to implement the exact same feature. the standard default size really bothers me, way to small to work with for myself ... hope they bake your feature into the app soon.

index.js
```diff
@@ -47,11 +48,19 @@ app.on('window-all-closed', () => {
 
 app.on('ready', () => {
   function createWindow (fn) {
+    let windowState = windowStateKeeper({
```
albinekb: `const` ?

mawie81: Whops you are right. Just fixed that.

mawie81:
After #257 got merged I´m not sure if this does make sense anymore. The configured default window sizes would only get used once and then no more.
What do you think?

albinekb:
If you look at how iTerm2 (and terminal.app i suppose) does it, maybe it makes more sense.

From what i understand if you close all your windows, there will be nothing saved to the state, and thus we would want to load the set config values from #257?

mawie81:
I think what iTerm does is basically just saving each window position but always uses the configured rows/columns for the window size.
Since the later already kind of works using the ```windowSize``` I could just skip setting the size and only use the saved position. What do you think?

rauchg:
If `windowSize` is not user-set, I'd like to use the memorized window position (which we can store in `electron-config`). Is that what you mean?

LanderN:
Not really sure what the "Awaiting Response" label is used for, but I'm just going to assume you mean responses from users here.
I, personally, am pretty annoyed by my terminal always opening on my primary display in a tiny window while I closed it as a maximized window on my secondary display. 
So, I would love this pull request to be merged, I think it's way better than specifying a window size in the configuration file.
Anyways, keep up the good work! Loving HyperTerm so far :)

ppot:
@LanderN Check out my PR #617

leo:
https://github.com/zeit/hyper/pull/617

PR #196, opened 2016-07-17
make update feed not only OS X

Haroenv:
I'm not sure what the url for others is, since I'm not fully sure if it already exists for other platforms.

### DIFF
--- a/auto-updater.js
+++ b/auto-updater.js
@@ -3,7 +3,7 @@ const { version } = require('./package');
 const notify = require('./notify');
 const ms = require('ms');
 
-const FEED_URL = 'https://hyperterm-updates.now.sh/update/osx';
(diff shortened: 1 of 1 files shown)


rauchg:
We should probably pass it dynamically. We are using: https://github.com/GitbookIO/nuts. You can probably look there for what the parameter names would be.

rauchg:
Needs rebase

Haroenv:
for Windows, it's `/win32`, quickly looking for what it is for linux

rauchg:
@Haroenv can you rebase please?

Haroenv:
Sure! forgot about this

Haroenv:
I'm pretty sure that this PR doesn't add anything new that isn't already in that file. Closing now.

PR #197, opened 2016-07-17
Enable transparent backgroundColor

fand:
Yet `.hyperterm.js` accepts `rgba()` colors for `backgroundColor`, it can't be transparent due to the `backgroundColor` option for BrowserWindow.
We don't need the option because `backgroundColor` is passed to `document.body` here:
https://github.com/zeit/hyperterm/blob/450e337/app/lib/containers/hyperterm.js#L22

This PR removes `backgroundColor` option for BrowserWindow.
and it works like this:
hyper


rauchg:
What does startup look like now though? Is there a flicker? Or traffic lights with no background? In which case we might only want to turn off the `backgroundColor` parameter for `BrowserWindow` in the presence of an alpha channel.

Fantastic stuff btw.

dfrankland:
The `backgroundColor` option for BrowserWindow can accept an 8 digit `#AARRGGBB` CSS Level 4 hex color though. So passing in a string like `#AA000000` you can get a dark transparent background.

marcbachmann:
There will be shadow issues when the window uses a transparent background.
It's somewhat experimental.  And yes. There's a flicker because the window doesn't have any background before the webview is loaded.

So first it's white for a short time, then transparent and the window action buttons are shown and then the webview gets loaded. It's best to hide the window first and show it afterwards.

fand:
Yeah 8-digit `#AA000000` style backgroundColor worked fine on my mac.
In 764314d , succeeded updating bg color without removing `backgroundColor` option.

hy-work

But I found a case `setBackgroundColor` doesn't work.
When hyperterm was lauched with 6-digit `backgroundColor`, we can't update its color with 8-digit styles.

hy

It may be a restriction from electron spec...?? idk but it seems a bit complicated.

I'm gonna try hiding the window with my first approach.

albinekb:
@fand, maybe this can help you hide the window during loading: https://github.com/resin-io/etcher/pull/141/files

Evanlec:
Transparent backgroundColor works for me also (after re-launching hyperterm), but I could not get it to work (using an 8-digit hex code) for cursorColor -- it is very helpful to have transparent cursor, especially when using things like vim.

fand:
I tried to apply this approach https://github.com/zeit/hyperterm/pull/197#issuecomment-234457487 but hyperterm now hides its window until the prompt is ready by default.
https://github.com/zeit/hyperterm/blob/77597da/app/index.js#L95
https://github.com/zeit/hyperterm/blob/77597da/app/index.js#L126

I just reverted 764314d and it looks working fine.
But now electron's `backgroundColor` doesn't accept 8-digits color string ... ( .o.)

https://github.com/zeit/hyperterm/pull/524 may solve this problem.

leo:
Works in the latest release!

Issue #131, opened 2016-07-15
Encoding issues


Rydgel:
Strange issues with "special characters".
For example typing, `é` displays `Ã©` in the terminal.
It makes it totally unusable. Is there some configuration I'm missing?

        HyperTerm 0.4.5
        Electron 1.2.5
        darwin x64 15.5.0

lee-dohm:
Following the instructions in [a StackOverflow answer](http://stackoverflow.com/a/8161863/1954), I was able to resolve this by adding the following to my shell profile:

```shell
export LC_ALL=en_US.UTF-8
export LANG=en_US.UTF-8
```

It would be nice still to be able to configure the terminal directly.

nfcampos:
doing that didn't change anything for me

YogaPan:
+1 For this encoding issues.
I can't type traditional Chinese in hyperTerm.
It always turn into some character like "ã¨Ë".

Mac Os X 10.11.5
HyperTerm 0.5.0

Rydgel:
0.6.0 fixes that for me. What about you guys?

nfcampos:
@Rydgel for me as well, now the only characters that are not corrrectly shown are emojis, which are shown like this `�`

Dremora:
I'm on 0.6.0 but Unicode still doesn't work. Typing a Unicode character results in 2 tabs, displaying (e.g. using `ls`) represents each character as `?`.
`zsh 5.2`.

staltz:
Just reporting my issues as well. Funky encoding wizardry is going on that affects not just look and feel, but also actual execution of commands.

Using hyperterm 0.6.0, when I run Jekyll (this happened to 2 different Jekyll projects I have) on Hyperterm:
```
>  raulmariano   jekyll serve                                       
Configuration file: /Users/staltz/oss/raulmariano/_config.yml
            Source: /Users/staltz/oss/raulmariano
       Destination: /Users/staltz/oss/raulmariano/_site
 Incremental build: disabled. Enable with --incremental
      Generating... 
jekyll 3.1.6 | Error:  "\xE6" on US-ASCII
```

And then when I run it in another terminal:
```
>  raulmariano   jekyll serve
Configuration file: /Users/staltz/oss/raulmariano/_config.yml
            Source: /Users/staltz/oss/raulmariano
       Destination: /Users/staltz/oss/raulmariano/_site
 Incremental build: disabled. Enable with --incremental
      Generating... 
                    done in 1.568 seconds.
[...]

PR #201, opened 2016-07-17
Fixing problems with alt key and special characters like å, ä and ö

teemuteemu:
Should fix issues like #131


marcbachmann:
Thanks. Looks good 👍

rauchg:
This is the best thing that's happened to this project! Thank you!

patrickhaug:
On german layout, ~ doesnt show up for alt/opt + n. i can hack it there while typing alt + shift + n and deleting the unicodes before the ~.

marcbachmann:
> On german layout, ~ doesnt show up for alt/opt + n. i can hack it there while typing alt + shift + n and deleting the unicodes before the ~.

This terminal is so awesome, it makes you want to switch keyboards.

Umlauts are also still a problem. (ALT + u ... u)
Here's how iTerm handles them. I have no idea how hterm manages to swallow those inputs.
umlauts

Issue #125, opened 2016-07-15
Allow to disable sounds in configuration file


vslinko:
"Ding" sound for example.

sheerun:
Could we please change default sound indication to something else? I do lot of Vim and I'm used to press "Esc" multiple times just to be sure, and every time it makes ding.. A little flask of titlebar would be far more pleasurable, but sufficient indication.

gerchicov-bp:
@albinekb it doesn't work in both windows and osx

PR #204, opened 2016-07-17
Config option to silence bell

bekher:
Add a config option `audibleBell` and handlers to silence the bell by muting hterm's bell audio. This is achieved by toggling `this.term.bellAudio_.muted` in app/lib/components/term.js. 

We leave the bell on by default so users with out of date config files will not wonder where their bell audio went. 

Closes issue #125

### DIFF
--- a/app/config-default.js
+++ b/app/config-default.js
@@ -54,7 +54,10 @@ module.exports = {
 
     // the shell to run when spawning a new session (i.e. /usr/local/bin/fish)
     // if left empty, your system's login shell will be used by default
-    shell: ''
+    shell: '',
+
+    // enable bell ring
+    enableBell: true,
 
     // for advanced config flags please refer to https://hyperterm.org/#cfg
   },
--- a/app/lib/utils/override-event.js
+++ b/app/lib/utils/override-event.js
@@ -1,0 +1,17 @@
+const config = require('./config');
+
+let enableBell = config.getConfig().enableBell;
+
+config.subscribe(() => {
+  // listen on confg update to enable/disable bell
+  enableBell = config.getConfig().enableBell;
+});
+
+export default function shouldIgnoreEvent (data) {
+  switch(data) {
+    case '\u0007':
+        return !enableBell;
+  }
+
+  return false;
+}
--- a/lib/actions/sessions.js
+++ b/lib/actions/sessions.js
@@ -1,5 +1,6 @@
 import rpc from '../rpc';
 import getURL from '../utils/url-command';
+import shouldIgnoreEvent from '../utils/override-event';
 import { keys } from '../utils/object';
 import {
   SESSION_ADD,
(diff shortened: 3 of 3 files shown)


app/lib/components/term.js
```diff
@@ -41,6 +43,9 @@ export default class Term extends Component {
         if (cols !== this.props.cols || rows !== this.props.rows) {
           props.onResize(cols, rows);
         }
+
+      // bell on by default if not defined in the config
+      this.term.bellAudio_.muted = (props.audibleBell == false);
```
rauchg: There's probably a `pref` for this in hterm?

marcbachmann: there's `audible-bell-sound`. It accepts an empty string to disable the bell.
`'lib-resource:hterm/audio/bell'` is the default sound.

rauchg: We should use that instead like the other prefs.

rauchg:
Once this is ready & rebased I'll merge!

bekher:
Looks like I missed that pref, I didn't see it in the hterm source. @marcbachmann's suggestion works as expected, I'll rebase this tonight!

bekher:
Alright @rauchg , I've rebased this, and we now toggle the bell via the `audible-bell-sound` perf.

config-default.js
```diff
@@ -50,7 +50,10 @@ module.exports = {
 
     // the shell to run when spawning a new session (i.e. /usr/local/bin/fish)
     // if left empty, your system's login shell will be used by default
-    shell: ''
+    shell: '',
+
+    // enable audible bell
+    audibleBell: true
```
rauchg: Sorry for changing the requirements on you last minute. Do you think we should allow `false` *or* a full path to a custom sound? That way plugins can introduce sounds :D

bekher: No problem, I think a true/false switch would be intuitive, but a path would be better for future plugins. We could include a comment describing the behavior, how does this look?

```javascript
// set hterm audible bell path, default is 'lib-resource:hterm/audio/bell'
// if left empty, the audible bell will not ring
bellSoundPath: 'lib-resource:hterm/audio/bell',
```

albinekb:
Does it make sense to start grouping the settings? ex:

```
bell: {
  path: 'lib-resource:hterm/audio/bell',
  audible: true
}
```

and later you could group window-settings etc, what do you think?

Probably for another PR but why not start here 👼

config-default.js
rauchg: I think string / true / falsy is fine honestly.

indutny:
Ping 😉

marcbachmann:
@albinekb that definitely belongs into a separate pr. Otherwise this would never get merged.
@bekher is it ok for you if somebody takes over?
Or do you already have some uncommited changes?

bekher:
I can rebase this again later in the day, and if anyone wants to make any other changes after that that should be fine.

bekher:
Alright, it's rebased again @rauchg . If anyone wants to make other changes feel free to take over.

rjsteinert:
So much ding... I have to mute my computer when using hyperterm.

bekher:
Looks like I accidentally committed some old code from a few weeks ago when I was playing around with this, will fix tonigjt.

bekher:
Looks like #468 is taking care of it with a better config option. I'll close this out.

PR #205, opened 2016-07-17
Dynamically change the `font-smoothing` pref

mike-engel:
By default, hterm defaults to `font-smoothing: 'antialiased'`, which
works really well on retina displays. On non-retina displays, however,
the type looks very thin and is hard to read.

This will look at the devicePixelRatio of the device anytime the term
prefs are set, and change between `antialiased` and
`subpixel-antialiased` dynamically.

Screenshots below.
<details>
#### retina (👍)
retina

#### non-retina, before (👎)
non-retina-pre

#### non-retina, after (👍)
non-retina-post
</details>

### DIFF
--- a/app/lib/components/term.js
+++ b/app/lib/components/term.js
@@ -26,7 +26,7 @@ export default class Term extends Component {
 
     this.term.prefs_.set('font-family', props.fontFamily);
     this.term.prefs_.set('font-size', props.fontSize);
-    this.term.prefs_.set('font-smoothing', this.getFontSmoothing());
+    this.term.prefs_.set('font-smoothing', props.fontSmoothing);
     this.term.prefs_.set('cursor-color', props.cursorColor);
     this.term.prefs_.set('enable-clipboard-notice', false);
     this.term.prefs_.set('foreground-color', props.foregroundColor);
@@ -104,16 +104,6 @@ export default class Term extends Component {
     return URL.createObjectURL(blob, { type: 'text/css' });
   }
 
-  getFontSmoothing () {
-    this.devicePixelRatio = window.devicePixelRatio;
-
-    if (this.devicePixelRatio < 2) {
-      return 'subpixel-antialiased';
-    }
-
-    return 'antialiased';
-  }
-
   componentWillReceiveProps (nextProps) {
     if (this.props.url !== nextProps.url) {
       // when the url prop changes, we make sure
@@ -152,6 +142,10 @@ export default class Term extends Component {
       this.term.prefs_.set('font-family', nextProps.fontFamily);
     }
 
+    if (this.props.fontSmoothing !== nextProps.fontSmoothing) {
+      this.term.prefs_.set('font-smoothing', props.fontSmoothing);
+    }
+
     if (this.props.cursorColor !== nextProps.cursorColor) {
       this.term.prefs_.set('cursor-color', nextProps.cursorColor);
     }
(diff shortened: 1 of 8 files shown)


app/lib/components/term.js
```diff
@@ -149,6 +160,10 @@ export default class Term extends Component {
     if (this.props.customCSS !== nextProps.customCSS) {
       this.term.prefs_.set('user-css', this.getStylesheet(nextProps.customCSS));
     }
+
+    if (this.devicePixelRatio !== window.devicePixelRatio) {
+      this.term.prefs_.set('font-smoothing', this.getFontSmoothing());
```
rauchg: Can `window.devicePixelRatio` change? I think we probably need to do it only once in the ctor?

rauchg:
This is fantastic btw. Thank you!

app/lib/components/term.js
mike-engel: It can, if say, you move from a non-retina screen to a retina screen. I have a retina macbook pro for work, and a secondary monitor that's non-retina. So if I move the window between screens, this would catch that and adjust as necessary 😄

rauchg: Right, I was thinking about that, but would that transition really guarantee that new props will be received? Or should we hook into some sort of event (if any)? Maybe `resize` triggers? Can you confirm? And if so, we should document that we're relying on that effect.

developit: Electron has a [BrowserWindow event for move](https://github.com/electron/electron/blob/master/docs/api/browser-window.md#event-move) that would be triggered when switching between screens.

Or just poll the window position:

```js
let position;
setInterval( () => {
  let next = window.screenLeft+','+window.screenTop;
  if (position && next!==position) {
    console.log('moved');
  }
  position = next;
}, 500);
```

rauchg: Subscribing to move sounds like a very good idea. We can subscribe to an `rpc` event and then dispatch a redux action?

mike-engel:
Refactored to use state/move event 😄

app/lib/index.js
```diff
@@ -99,6 +100,10 @@ rpc.on('update available', ({ releaseName, releaseNotes }) => {
   store_.dispatch(updaterActions.updateAvailable(releaseName, releaseNotes));
 });
 
+rpc.on('move', () => {
+  store_.dispatch(uiActions.setFontSmoothing());
```
rauchg: To make this absolutely perfect, what I would do here is dispatch the *actual* action / intent that's ocurring (in this case, `UI_WINDOW_MOVE` action or similar, which as an `effect` dispatches that other action (so that it's overridable by plugins ;))

rauchg: Otherwise, amazing job!!!!!!!!!!!!!!!!

mike-engel: Thanks, this has been a really well-done project, and is fun to work on! I added a `UI_WINDOW_MOVE` action in a way that I think does what you were asking for. If not, let me know 😄

rauchg:
So awesome. Thank you

mike-engel:
Thanks! As always, glad to help out here 😄

PhiloNL:
Is there a way to overwrite this? I actually like the font antialiased on a regular screen as well. I've tried overwriting this in .hyperterm.js but without any result `fontSmoothing: 'antialiased'`

Thanks!

mike-engel:
@PhiloNL Have you tried adding custom css in your hyperterm config? If that doesn't work, we can possibly add another config option to overwrite the fontSmothing property.

PhiloNL:
@mike-engel thanks, that works!

omriyariv:
@PhiloNL can you please share your custom css config?

PhiloNL:
@omriyariv sure thing!

```
{
    // custom css to embed in the main window
    css: '* { -webkit-font-smoothing: antialiased }',

    termCSS: '* { -webkit-font-smoothing: antialiased }',
}
```

PR #209, opened 2016-07-17
Hide scrollbar when webview is active

### LINKED ISSUE #208: Scrollbar div overlaps webview content
image

marcbachmann:
This fixes #208


marcbachmann:
@rauchg this is ready to merge

rauchg:
Love that we already had the `: null` ready to be replaced with it :D

PR #211, opened 2016-07-17
Add support for a `registry` configuration field

developit:
For those of us using a system-wide private registry, this helps by allowing hyperterm to bypass it when installing/updates plugins.

I took a bit of a guess here as to what the recommended approach would be, just figured this would be more useful than an open issue.

### DIFF
--- a/config-default.js
+++ b/config-default.js
@@ -51,6 +51,8 @@ module.exports = {
     // the shell to run when spawning a new session (i.e. /usr/local/bin/fish)
     // if left empty, your system's login shell will be used by default
     shell: ''
+
+    // for advanced config flags please refer to https://hyperterm.org/#cfg
   },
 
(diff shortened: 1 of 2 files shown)


rauchg:
Thanks @developit! 

- `npmRegistry` would be a better name
- if you use `getDecoratedConfig` from the very `plugins.js` file you can get the config after plugins modify it (recommended)
- We need to add to `config-default.js` a line at the end of the `config: { }` object that says:
 - `// for advanced config flags please refer to https://hyperterm.org/#cfg`
- It would be nice to document the option like @mike-engel did here: https://github.com/zeit/hyperterm-website/pull/18

rauchg:
Needs rebase it seems?

developit:
Darn, I knew I couldn't get away with an in-browser PR 😋

Issue #29, opened 2016-07-03
Special characters and foreign keyboard not displayed correctly


jaimeeee:
So I use Spanish ISO keyboard (like this one: http://g01.a.alicdn.com/kf/HTB1iOQ8HVXXXXaeXpXXq6xXFXXXd/Spanish-Euro-Keyboard-Cover-Protector-for-MacBook-Air-11-11-6-Inch-European-ISO-Keyboard-Spain.jpg_640x640.jpg), and it is impossible to type some characters.

For example, "special" characters like ¿¡· are displayed like this: <00c2><00bf><00c2><00a1><00c2><00b7>, same with the ñ.

And other keyboard combinations are just impossible to type, my biggest issue is **~**, since I need to type _Alt + ñ_ but it just doesn't appear. Same issue with any letter with an accent (á, ö, â, etc).

I'm using zsh, if it matters.

rauchg:
@jaimeeee can you please test with 0.2.1? Thanks for taking the time to report this.

jaimeeee:
Still happening on 0.2.1, I get a "ding" sound when I press Alt + ñ

johanbrook:
I can also reproduce for <kbd>alt+n</kbd> (in order to do `ñ`). Getting the ding sound on 0.2.1.

Also, I cannot produce a tilde (`~`) on my (Swedish) keyboard by doing an <kbd>alt</kbd> key combo. No ding sound is played and cursor stays put.

rauchg:
Please retry with `0.3.1`.

jaimeeee:
Now on 0.3.1 and sadly it is still happening, nothing changed imo.

michalkvasnicak:
Have same problem in version `0.5.0`. Using slovak keyboard in order to product `~` pressing <kbd>alt</kbd>+<kbd>ť</kbd> is only doing ding sound.

mxstbr:
Still happening for me (newest version), am on a german keyboard and can't type `@`.

bendc:
Lots of crucial keys like `^` are actually not working for me either (Belgian AZERTY keyboard).

LoicMahieu:
Same than @bendc here. Belgian keyboard and caracters like ``` ` ``` , `^`, `~` are not working.

sullenor:
Looks like I have the same issue in `0.6.0` version with cyrillic symbols.

apacala:
Same. `£` gets displayed as `<00c2><00a3>`.

[...]

PR #214, opened 2016-07-17
[WIP] support composed characters

### LINKED ISSUE #66: can't write accented characters on a portuguese keyboard
This sentence here

> Note that composed characters (those that require multiple keystrokes) are not currently supported by this mode.

https://chromium.googlesource.com/apps/libapps/+/HEAD/nassh/doc/FAQ.md#How-do-I-enter-accented-characters

suggests this is not possible using `hterm`

any has any idea how to solve this?

nfcampos:
should close #66 and part of #29 (stuff like trying to type `~` in some international keyboards and not seeing anything should be totally fixed with this)

todo:
- [x] row that has the cursor is set to contentEditable = true
- [x] hide default blue outline and cursor for contentEditable elements
- [x] figure out why cursor is now missing its fill (it's not focused...)
- [x] ~~figure out to make it possible to insert emojis~~

this should fix most issues with international keyboards + composed characters
also makes it possible to use the native mac emoji dialog (emojis aren't still being written to the terminal for some reason, not sure why)

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -99,7 +99,21 @@ export default class Term extends Component {
   write (data) {
     requestAnimationFrame(() => {
       this.term.io.writeUTF8(data);
+      this.updateCurrentRow();
     });
+  }
+
+  updateCurrentRow () {
+    const nextCurrentRow = this.term.getCursorRow();
+
+    if (this.prevCurrentRow !== nextCurrentRow) {
+      const {rowsArray} = this.term.screen_;
+      if (this.prevCurrentRow !== undefined) {
+        rowsArray[this.prevCurrentRow].removeAttribute('contentEditable');
+      }
+      rowsArray[nextCurrentRow].contentEditable = true;
+      this.prevCurrentRow = nextCurrentRow;
+    }
   }
 
   focus () {
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -231,6 +231,10 @@ export default class Term extends Component {
       this.term.prefs_.set('font-size', nextProps.fontSize);
     }
 
+    if (this.props.customCSS !== nextProps.customCSS) {
+      this.term.prefs_.set('user-css', this.getStylesheet(nextProps.customCSS));
+    }
+
     if (this.props.foregroundColor !== nextProps.foregroundColor) {
       this.setCssVariable('--foreground-color', nextProps.foregroundColor);
     }
@@ -257,10 +261,6 @@ export default class Term extends Component {
 
     if (this.props.colors !== nextProps.colors) {
       this.term.prefs_.set('color-palette-overrides', getColorList(nextProps.colors));
-    }
-
-    if (this.props.customCSS !== nextProps.customCSS) {
-      this.term.prefs_.set('user-css', this.getStylesheet(nextProps.customCSS));
     }
 
     if (this.props.bell === 'SOUND') {
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -31,9 +31,10 @@ export default class Term extends Component {
     this.term.prefs_.set('font-smoothing', props.fontSmoothing);
     this.term.prefs_.set('cursor-color', this.validateColor(props.cursorColor, 'rgba(255,255,255,0.5)'));
     this.term.prefs_.set('enable-clipboard-notice', false);
-    this.term.prefs_.set('foreground-color', props.foregroundColor);
-    this.term.prefs_.set('background-color', props.backgroundColor);
+    this.term.prefs_.set('foreground-color', 'var(--foreground-color)');
+    this.term.prefs_.set('background-color', 'var(--background-color)');
     this.term.prefs_.set('color-palette-overrides', getColorList(props.colors));
+    this.term.prefs_.set('color-palette-overrides', props.colors);
     this.term.prefs_.set('user-css', this.getStylesheet(props.customCSS));
     this.term.prefs_.set('scrollbar-visible', false);
     this.term.prefs_.set('receive-encoding', 'raw');
@@ -167,8 +168,26 @@ export default class Term extends Component {
     return this.term.document_;
   }
 
+  setCssVariable (key, value) {
+    return this.getTermDocument().body.style.setProperty(key, value);
+  }
+
   getStylesheet (css) {
     const blob = new Blob([`
+      :root {
+        --background-color: ${this.props.backgroundColor};
+        --foreground-color: ${this.props.foregroundColor};
+      }
+
+      x-row {
+        text-shadow: 0 0 0 var(--foreground-color);
+        color: transparent;
+      }
+
+      x-row:focus {
+        outline: none;
+      }
+
       .cursor-node[focus="false"] {
         border-width: 1px !important;
       }
@@ -213,11 +232,11 @@ export default class Term extends Component {
     }
 
     if (this.props.foregroundColor !== nextProps.foregroundColor) {
-      this.term.prefs_.set('foreground-color', nextProps.foregroundColor);
+      this.setCssVariable('--foreground-color', nextProps.foregroundColor);
     }
 
     if (this.props.backgroundColor !== nextProps.backgroundColor) {
-      this.term.prefs_.set('background-color', nextProps.backgroundColor);
+      this.setCssVariable('--background-color', nextProps.backgroundColor);
     }
 
     if (this.props.fontFamily !== nextProps.fontFamily) {
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -34,7 +34,6 @@ export default class Term extends Component {
     this.term.prefs_.set('foreground-color', 'var(--foreground-color)');
     this.term.prefs_.set('background-color', 'var(--background-color)');
     this.term.prefs_.set('color-palette-overrides', getColorList(props.colors));
-    this.term.prefs_.set('color-palette-overrides', props.colors);
     this.term.prefs_.set('user-css', this.getStylesheet(props.customCSS));
     this.term.prefs_.set('scrollbar-visible', false);
     this.term.prefs_.set('receive-encoding', 'raw');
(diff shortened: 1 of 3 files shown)


app/lib/reducers/ui.js
```diff
@@ -24,7 +24,15 @@ const initial = Immutable({
   fontFamily: 'Menlo, "DejaVu Sans Mono", "Lucida Console", monospace',
   fontSizeOverride: null,
   css: '',
-  termCSS: '',
+  termCSS: '' +
```
nfcampos: @rauchg can i use an ES6 template string instead of this ugliness?
and is this the right place for this?

app/lib/reducers/ui.js
```diff
@@ -24,7 +24,15 @@ const initial = Immutable({
...
+    'x-row {' +
+    '  text-shadow: 0 0 0 #fff;' +
+    '  color: transparent;' +
+    '}' +
+    'x-row:focus {' +
+    '  outline: none;' +
```
nfcampos: this removes the ugly blue outline

app/lib/reducers/ui.js
```diff
@@ -24,7 +24,15 @@ const initial = Immutable({
   fontFamily: 'Menlo, "DejaVu Sans Mono", "Lucida Console", monospace',
   fontSizeOverride: null,
   css: '',
-  termCSS: '',
+  termCSS: '' +
+    'x-row {' +
+    '  text-shadow: 0 0 0 #fff;' +
```
nfcampos: this removes the default input cursor. not sure on how to make this respect the color set for the text @rauchg

app/lib/reducers/ui.js
rauchg: yep, well familiar with it :D

app/lib/reducers/ui.js
rauchg: I recommend you instead add it [here](https://github.com/zeit/hyperterm/blob/master/app/lib/components/term.js#L105), since it's technically part of our built-in styles, and this is user-set.

rauchg: You can certainly use template strings btw, it's natively supported by the node used by electron.

app/lib/reducers/ui.js
rauchg: You mean the color set for the text or the color set for the cursor? If the former, you have to set it to `foregroundColor`, and make sure that when `foregroundColor` changes the css gets reloaded in `componentWillReceiveProps`

nfcampos: I went with css variables for the colours, eventually I think we should move all colours to css variables, if not for anything else it makes easier to develop themes, you can change them right in the inspector

app/lib/components/term.js
```diff
@@ -52,6 +53,7 @@ export default class Term extends Component {
 
     const iframeWindow = this.getTermDocument().defaultView;
     iframeWindow.addEventListener('wheel', this.onWheel);
+    iframeWindow.addEventListener('focus', this.focus, true);
```
nfcampos: this might be too agressive (ie. `this.focus`ends up running a bunch of times) but i'm not sure of how to focus the cursor otherwise, because every keystroke focuses it on the row on which you're typing

app/lib/components/term.js
```diff
@@ -100,14 +116,33 @@ export default class Term extends Component {
...
+        outline: none;
+      }
+
       .cursor-node[focus="false"] {
         border-width: 1px !important;
       }
       ${css}
-    `]);
-    return URL.createObjectURL(blob, { type: 'text/css' });
+    `], { type: 'text/css' });
```
nfcampos: this was in the wrong place

app/lib/components/term.js
```diff
@@ -137,19 +172,19 @@ export default class Term extends Component {
...
+      this.setCssVariable('--background-color', nextProps.backgroundColor);
     }
 
     if (this.props.fontFamily !== nextProps.fontFamily) {
       this.term.prefs_.set('font-family', nextProps.fontFamily);
     }
 
     if (this.props.cursorColor !== nextProps.cursorColor) {
-      this.term.prefs_.set('cursor-color', nextProps.cursorColor);
+      this.setCssVariable('--cursor-color', nextProps.cursorColor);
```
nfcampos: might have a race condition if nextProps has both new `cursorColor` (or one of the other two colors) and `customCss`, I think the old color might stay put, shouldn't be too hard to fix though

mxstbr:
I can't wait for this to land! Anything I can help with @nfcampos?

nfcampos:
@mxstbr if you could figure out why emojis still don't show up properly that'd be awesome, i haven't managed to do that yet

weslleyaraujo:
can't wait for that as well, currently copying/pasting `~` 😅
nice work guys!

app/lib/components/term.js
nfcampos: @rauchg is this okay? or it will impact typing performance?

app/lib/components/term.js
nfcampos: fixed in the latest commit

nfcampos:
other than emoji support (but that can de done later I think) this looks ready to me!
can you guys try this out? cc @mxstbr @rauchg @weslleyaraujo and anyone else who's been having problems with typing stuff

mxstbr:
I'm not sure if it's on my end, but I pulled these changes, ran `npm run dev` in the `app/` folder, forced a rebuild, started the app and nothing is different. 😢

nfcampos:
yeah I just noticed that too... it looks like it's because if the cursor node has the focus then those extra characters don't get captured, im looking into it (I think this used to work but stopped working now that I rebased on the latest master, but i might be wrong...)
if you comment this line out https://github.com/zeit/hyperterm/pull/214/files#diff-f35cc0c2adb56ebeaab7ea051ace9ceaR56 it works for you?

nfcampos:
@mxstbr I've reverted the commit that focused the cursor node so it should now work for you if you pull hopefully

mxstbr:
Yeah, `~` works sometimes but is quite buggy. Umlauts (e.g. `ä`) still don't work. :cry:

nfcampos:
hmm, it works for me every time
screen shot 2016-07-20 at 20 53 39

nfcampos:
which is not to say this isn't buggy, I can currently identify at least three bugs:
- if you type the same accent repeatedly and no letter character that 'consumes' it in between they appear to both sides of the cursor 😱
screen shot 2016-07-20 at 20 56 47
- if you hit <kbd>cmd</kbd>+<kbd>left</kbd> a blinking cursor appears at the very beginning of the line, before the prompt
screen shot 2016-07-20 at 20 59 07
 - sometimes the cursor is filled in the right color, sometimes it's transparent and only the border has color, and i haven't figured out why it's not consistent

but I can consistently type all characters, accents, etc (everything other than emojis, on my european portuguese keyboard at least)

weslleyaraujo:
I faced a similar behavior.
If I start a command using `` and try to erase it, the first character will never disappear

screen shot 2016-07-21 at 10 43 06 am

the character is not really there, is like "ghost rendered" ... 
In general, the composed characters are working properly, just sometimes I got a small delay typing 👍

app/lib/components/term.js
rauchg: this explains a lot (of warnings)

app/lib/components/term.js
rauchg: nice!

app/lib/components/term.js
```diff
@@ -181,16 +218,12 @@ export default class Term extends Component {
...
     }
 
     if (this.props.colors !== nextProps.colors) {
       this.term.prefs_.set('color-palette-overrides', nextProps.colors);
     }
-
-    if (this.props.customCSS !== nextProps.customCSS) {
-      this.term.prefs_.set('user-css', this.getStylesheet(nextProps.customCSS));
```
rauchg: how are user additions of custom CSS via config handled now?

nfcampos: same way, I just moved this block a couple lines above to avoid a race condition if both customCSS and one of the colors changed simultaneously

rauchg: gotcha

rauchg:
Regarding emojis, curious why it's an open question? The Emoji IME shows up wherever there's focus.

nfcampos:
@rauchg regarding emojis, it was an open question because while the emoji dialog did show up emojis aren't being displayed correctly when inserted. but it seems like that is dealt with by this other PR #535 right?

nfcampos:
the main thing blocking this PR now is I think that the cursor node no longer has focus. I'm not sure how to solve that because only one element can have focus at a time and the current row needs focus to receive the composed characters input, any ideas are welcome

matheuss:
@nfcampos friendly ping ^~^

nfcampos:
@matheuss I've rebased on top of the latest master. I still haven't figured out a way to ensure composed characters *and* the cursor node has focus. Any input on that would be very welcome

rauchg:
I'm gonna give it a shot

carhartl:
I've setup my prompt to use git-radar (https://github.com/michaeldfallen/git-radar), which seems to use a unicode mathematical italic m in order to denote the master branch, e.g. _m_.

In Terminal.app (bash + zsh) the italic character renders fine, in Hyper what I see is the infamous question mark in a box:

italic

The above was setup for quick testing like so (taken from how git-radar sets this up):

```
local italic_m="$(printf '\xF0\x9D\x98\xAE')"
export PROMPT="$italic_m "
```
=> https://github.com/michaeldfallen/git-radar/blob/c67d6bcd88c9073bc072af1cb13d7bfa85cbbbb0/radar-base.sh#L97

Is this the same issue?

kauegimenes:
Any update on this?

matheuss:
Closing in favor of #1006 😄 Thank you so much @nfcampos 🙌

PR #215, opened 2016-07-17
Make CMD + K work for all commands, even tail

### LINKED ISSUE #65: Cmd-K doesn't work while tailing
Saw cmd-k landed, and for most use cases it's working nicely. However, if I'm running a `tail -f` command, cmd-k no longer works. This is usually my primary use case for it - I clear the scrollback when tailing to "mark my place", then I can scroll to the top without getting confused with prior sessions.

marcbachmann:
Fixes #65


marcbachmann:
@rauchg that one is ready

Issue #157, opened 2016-07-16
Default support for 'alt <arrow>'


aaandrew:
Looks like alt <arrow> is not supported. The expected behavior is to jump between words, and the current behavior for alt <arrow> is to output ie. (';3C' or ';3D').

eljefedelrodeodeljefe:
Can confirm. Might be related to #161

marcbachmann:
Nope. Not related. Those keyboard shortcuts just aren't mapped to the escape sequences.
`Esc + b` & `Esc + F` are working.

So I tried to write a small plugin. While the plugin system of hyperterm works fine. But sadly I always get an error from hterm. It doesn't support escape sequences: https://github.com/chromium/hterm/blob/23300c0308d9e4ea9bd784807a4e40b1fb1ac69e/js/hterm_vt.js#L1323-L1330

montogeek:
Related https://github.com/zeit/hyperterm/issues/53

marcbachmann:
I have no idea what I did the first time when it didn't work.
But I had to debug the hterm code to get it working. Will continue in #217.

PR #217, opened 2016-07-17
Add cursor actions

### LINKED ISSUE #53: Option-arrow does not act like terminal.app
When modifying commands in other terminal apps (and text editors), option-arrow acts like the `b` and `w` commands in `vi`, jumping to the next word boundary. In hyperterm, it instead prints `;3D` and makes the `^G` noise.

Note that option-delete acts as expected, deleting to the nearest word boundary.

marcbachmann:
This makes `alt + <`, `alt + >`, `alt + delete` work
Fixes #53 #157
WIP

There's one open question. How do we want to trigger those methods?
Should I add a new variable in here?https://github.com/zeit/hyperterm/blame/master/app/lib/components/terms.js#L21-L23

 Or just set it up in the ui and don't use commands and reducers?

### DIFF
--- a/app/lib/components/term.js
+++ b/app/lib/components/term.js
@@ -96,6 +96,34 @@ export default class Term extends Component {
     }
   }
 
+  moveWordLeft () {
+    this.term.onVTKeystroke('\x1bb');
+  }
+
+  moveWordRight () {
+    this.term.onVTKeystroke('\x1bf');
+  }
+
+  deleteWordLeft () {
+    this.term.onVTKeystroke('\x1b\x7f');
+  }
+
+  deleteWordRight () {
+    this.term.onVTKeystroke('\x1bd');
+  }
+
+  deleteLine () {
+    this.term.onVTKeystroke('\x1bw');
+  }
+
+  moveToStart () {
+    this.term.onVTKeystroke('\x01');
+  }
+
+  moveToEnd () {
+    this.term.onVTKeystroke('\x05');
+  }
+
   getTermDocument () {
     return this.term.document_;
   }
--- a/app/lib/containers/hyperterm.js
+++ b/app/lib/containers/hyperterm.js
@@ -49,6 +49,17 @@ class HyperTerm extends Component {
     keys.bind('command+shift+]', moveRight);
     keys.bind('command+alt+left', moveLeft);
     keys.bind('command+alt+right', moveRight);
+
+    const bound = method => { return term[method].bind(term) }
+    keys.bind('alt+left', bound('moveWordLeft'));
+    keys.bind('alt+right', bound('moveWordRight'));
+    keys.bind('alt+backspace', bound('deleteWordLeft'));
+    keys.bind('alt+del', bound('deleteWordRight'));
+    keys.bind('command+left', bound('deleteWordLeft'));
+    keys.bind('command+right', bound('deleteWordRight'));
+    keys.bind('command+backspace', bound('deleteLine'));
+    keys.bind('command+left', bound('moveToStart'));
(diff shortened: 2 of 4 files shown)


rauchg:
The only reason the `write` object exists is for extensibility, so that plugins can alter the writes all throughout the Redux pipeline. It's a bit overkill, but I think well justified for that specific reason.

Is this something we can use mousetrap to capture?
https://github.com/zeit/hyperterm/blob/master/app/lib/containers/hyperterm.js#L37

marcbachmann:
> Is this something we can use mousetrap to capture?

Yes. But I had to ignore the alt key onKeyDown in hterm: https://github.com/zeit/hyperterm/pull/217/files#diff-c48d64c2956762890a45849618f86817R32 

So now everything is working. But I'm not using `dispatch` yet.
I

app/lib/components/term.js
```diff
@@ -11,6 +32,18 @@ export default class Term extends Component {
     this.onWheel = this.onWheel.bind(this);
     this.onScrollEnter = this.onScrollEnter.bind(this);
     this.onScrollLeave = this.onScrollLeave.bind(this);
+
+    for (const entry in actions) {
```
rauchg: I'd like to avoid this sort of meta-programming since it'll have a negative effect on startup time and static analysis.

app/lib/hterm.js
```diff
@@ -29,18 +29,23 @@ hterm.Terminal.prototype.copySelectionToClipboard = function () {
...
 const oldKeyDown = hterm.Keyboard.prototype.onKeyDown_;
 hterm.Keyboard.prototype.onKeyDown_ = function (e) {
-  if (e.metaKey) {
+  if (e.metaKey || e.altKey) {
     return;
   }
   return oldKeyDown.call(this, e);
 };
 
-const oldKeyPress = hterm.Keyboard.prototype.onKeyPress_;
```
marcbachmann: Hmm. I didn't wanted to commit this. Will remove it again.

app/lib/components/term.js
marcbachmann: 👍

marcbachmann:
I've also renamed the tab shortcuts. So this would be ready to merge.

marcbachmann:
Please be patient. We're all living in different time zones. And it's 3am in CA.
And we're not working full time on that project.

rauchg:
Wonderful as always @marcbachmann 🙏

app/lib/containers/hyperterm.js
```diff
@@ -49,6 +49,17 @@ class HyperTerm extends Component {
...
+    const bound = method => { return term[method].bind(term) }
+    keys.bind('alt+left', bound('moveWordLeft'));
+    keys.bind('alt+right', bound('moveWordRight'));
+    keys.bind('alt+backspace', bound('deleteWordLeft'));
+    keys.bind('alt+del', bound('deleteWordRight'));
+    keys.bind('command+left', bound('deleteWordLeft'));
+    keys.bind('command+right', bound('deleteWordRight'));
+    keys.bind('command+backspace', bound('deleteLine'));
+    keys.bind('command+left', bound('moveToStart'));
+    keys.bind('command+right', bound('moveToEnd'));
```
rauchg: @marcbachmann did you intend for the double `command+left`/`right` bind?

marcbachmann: Oh shit. Line 58 & 59 arent't needed.
Sorry.

ekmartin:
This works just as expected, really awesome - thank you!

blackmiaool:
I love this, but how to modify the key binding?

chabou:
You can't yet 😞

PR #218, opened 2016-07-18
Add `displayTitleCwd` configuration

thomaslindstrom:
This implements a configuration I'm very fond of in `Terminal.app`: Working directory in tab title. Particularly useful when you have multiple tabs open running the same process, eg. `node`.

screen shot 2016-07-18 at 02 15 12

This could probably also be a menu item under `View` with a check box next to it if active.

### DIFF
--- a/app/lib/reducers/ui.js
+++ b/app/lib/reducers/ui.js
@@ -53,6 +53,7 @@ const initial = Immutable({
   },
   foregroundColor: '#fff',
   backgroundColor: '#000',
+  displayTitleCwd: true,
   updateVersion: null,
   updateNotes: null
 });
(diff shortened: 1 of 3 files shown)


rauchg:
This seems like it would be a perfect thing for a userland plugin. 
Let me know if there are any hooks missing.

The way I would implement this is by making sure that when [we map the `titles`](https://github.com/zeit/hyperterm/blob/master/app/lib/containers/header.js#L17) from the state for [the tabs](https://github.com/zeit/hyperterm/blob/master/app/lib/containers/header.js#L29), which are received [here](https://github.com/zeit/hyperterm/blob/master/app/lib/components/tabs.js#L27), we want to prefix them with the directories.

I wouldn't change the `title` in the state itself, since the contract right now is pretty ok (title === process title). What I would do is add a map of uids -> directories somewhere else in the state, and when you set up your own `mapHeaderState` hook, you prefix them to the titles.

Let me know if this explanation makes sense. The hyperpower guide can be useful as well: https://hyperterm.org/#hyperpower

PR #219, opened 2016-07-18
Allow configuration of working directory

dfrankland:
Add a new`workingDirectory` property to the config to allow changing
the directory of new sessions.


config-default.js
```diff
@@ -50,7 +50,11 @@ module.exports = {
 
     // the shell to run when spawning a new session (i.e. /usr/local/bin/fish)
     // if left empty, your system's login shell will be used by default
-    shell: ''
+    shell: '',
+
+    // the working directory to start in when spawning a new session (i.e. ~/projects)
```
rauchg: `e.g.` is correct here. If you can wrap `~/projects` in backticks it'd be nicer too ;)

config-default.js
```diff
@@ -50,7 +50,11 @@ module.exports = {
...
+    // if left empty, your process.env.HOME directory will be used by default
```
rauchg: backticks around `process.env.HOME` would be nice too!

index.js
```diff
@@ -4,6 +4,7 @@ const createMenu = require('./menu');
 const Session = require('./session');
 const genUid = require('uid2');
 const { resolve } = require('path');
+const untildify = require('untildify');
```
rauchg: amazing. @sindresorhus has created every single module

rauchg:
+1 other than the minor nits

index.js
dfrankland: What would we do without you @sindresorhus 🎱

dfrankland:
👍 I have no problems with nitpicks. I have made the changes requested.

developit:
@rauchg will this break [hypercwd](https://github.com/hharnisc/hypercwd)? Or will it just provide a default that is then overridden by it? (thinking it's actually the latter since hypercwd would be providing the `cwd` option via rpc)

rauchg:
Default option. Pinging @hharnisc so that he can review this.

developit:
Awesome - I've been getting some cases where new tabs open to `/`, hopefully this fixes.

hharnisc:
@rauchg I'll take a look at this in detail this evening. Thanks for the heads
up!  
  
\- Harrison  


On Jul 18 2016, at 11:53 am, Jason Miller &lt;notifications@github.com&gt;
wrote:  

> Awesome - I've been getting some cases where new tabs open to /, hopefully
this fixes.

>

> —  
> You are receiving this because you were mentioned.  
> Reply to this email directly, [view it on
GitHub](https://github.com/zeit/hyperterm/pull/219#issuecomment-233388779&r=cmVwbHkrMDAxNTJlMmZmN2FkYWU3ZDAzNzVmMWZkMjFjNzRkODliYzU3ZThkM2Q2ZWM0ZTc1OTJjZjAwMDAwMDAxMTNhNGNiYTE5MmExNjljZTA5ZTUwYTJjQHJlcGx5LmdpdGh1Yi5jb20=),
or [mute the thread](https://github.com/notifications/unsubscribe-auth
/ABUuL9PDCaUkBrMrcfb_b_dVKefMf-
aMks5qW6-hgaJpZM4JOWpQ&r=cmVwbHkrMDAxNTJlMmZmN2FkYWU3ZDAzNzVmMWZkMjFjNzRkODliYzU3ZThkM2Q2ZWM0ZTc1OTJjZjAwMDAwMDAxMTNhNGNiYTE5MmExNjljZTA5ZTUwYTJjQHJlcGx5LmdpdGh1Yi5jb20=).

hharnisc:
Giving this a spin now, here goes nothing!

hharnisc:
So this works and doesn't interfere with the `hypercwd` plugin. 👍 

It is pretty nice to be able to specify where the first terminal tab opens. Nice one @dfrankland!

I have to ask before making the call to merge this though @rauchg. Could this be a plugin instead?

If we've got a common place to store plugin specific config parameters this seems like a nice simple plugin to implement. 

No matter what though I would really like to see the cwd pass through `untildify`

```js
rpc.on('new', ({ rows = 40, cols = 100, cwd }) => {
      cwd = cwd ? untildify(cwd) : process.env.HOME;
```

Plugin would look something like this

```js
const getCwd = () => {
  // get the workingDirectory param from the config somehow
};

exports.middleware = (store) => (next) => (action) => {
  switch (action.type) {
    case 'INIT':
      store.dispatch({
        type: 'UPDATE_CWD',
        cwd: getCwd()
      });
      break;
  }
  next(action);
};

exports.reduceUI = (state, action) => {
  switch (action.type) {
    case 'UPDATE_CWD':
      return state.set('cwd', action.cwd);
  }
  return state;
};
```

hharnisc:
So if #271 gets accepted you'll only need the following for a plugin 

```js
const getCwd = () => {
  // get the workingDirectory param from the config somehow
};

exports.middleware = (store) => (next) => (action) => {
  switch (action.type) {
    case 'INIT':
      store.dispatch({
        type: 'SESSION_SET_CWD',
        cwd: getCwd()
      });
      break;
  }
  next(action);
};
```

rauchg:
@hharnisc I like that idea. What if instead you add the option for a default path to `hypercwd`? I kinda like the idea of `hypercwd` taking care of all the working directory needs :)

rauchg:
Thinking more about this, I think having that in `hypercwd` or even another plugin is a good idea. It's also a good example to have in `-website` about how plugins can read custom configuration flags from `.hyperterm.js` by the way.

bevacqua:
See [`hyperterm-working-directory`](https://github.com/bevacqua/hyperterm-working-directory) :tada:

PR #220, opened 2016-07-18
Use #ffffff instead of #fff for foreground color as the short form seems to mess up.

mauro-oto:
Hey @rauchg, 

This PR fixes https://github.com/zeit/hyperterm/issues/177 and https://github.com/zeit/hyperterm/issues/199, which looks like a dupe of the former. 

It seems that for some reason, using `#fff` as the hex value for foreground correctly makes the text white, but highlighted texts are in a barely readable gray color. Using the full form `#ffffff` or `white` instead correctly makes both text and highlighted text white. 

Before: 

screen shot 2016-07-17 at 11 21 57 pm

After: 

screen shot 2016-07-17 at 11 20 53 pm

Please check it out, thanks!

### DIFF
--- a/app/lib/reducers/ui.js
+++ b/app/lib/reducers/ui.js
@@ -51,7 +51,7 @@ const initial = Immutable({
     resize: false,
     updates: false
   },
-  foregroundColor: '#fff',
+  foregroundColor: '#ffffff',
   backgroundColor: '#000',
   updateVersion: null,
   updateNotes: null
--- a/config-default.js
+++ b/config-default.js
@@ -10,7 +10,7 @@ module.exports = {
     cursorColor: '#F81CE5',
 
     // color of the text
-    foregroundColor: '#fff',
+    foregroundColor: '#ffffff',
 
     // terminal background color
     backgroundColor: '#000',


rauchg:
Amazing. How did you find this? 

My only concern is that folks will definitely try the short form and get disappointed. Maybe we should convert them when we're about to pass `foregroundColor` to `hterm` in the `term.js` `Component`?

Thank you so much!

rauchg:
Confirmed that it works. Miraculous.

hharnisc:
Works for me too! Would have never guessed this was the issue, nice find :+1: 

Could use the [color](https://www.npmjs.com/package/color) module here to parse the color and always use the long form.

```js
import Color from 'color';

const longFormHex = Color('#333').hexString();
```

rauchg:
Really sweet, specially since it seems that we'll be using that module for the other PR. I haven't checked, but I hope the `color` module is minimal (we're very sensitive to code size for startup time).

mauro-oto:
I was getting annoyed with the grey color when using it for path
autocompletion, so I tried to track down the source of that color to no
avail. At one point I experimented with changing foreground and background
colors and I used 'white' instead of '#fff', and noticed this behavior.

I agree with using the `Color` module, I'll update this PR or make a new
one tomorrow if you want to merge this before then.

Thanks for the feedback and for your work on this awesome project!

El jul. 17, 2016 23:49, "Guillermo Rauch" <notifications@github.com>
escribió:
>
> Amazing. How did you find this?
>
> My only concern is that folks will definitely try the short form and get
disappointed. Maybe we should convert them when we're about to pass
foregroundColor to hterm in the term.js Component?

hharnisc:
@rauchg I ran `webpack --json > stats.json` and dropped that into [webpack visualizer](https://chrisbateman.github.io/webpack-visualizer/)

In it's current form it amounts to 0.68% of the project at 10.5k

screen shot 2016-07-17 at 10 42 19 pm

React (for comparison) takes up 41% at 636k

screen shot 2016-07-17 at 10 45 56 pm

rauchg:
Awesome :)

devinus:
This seems like a underlying issue elsewhere, doesn't it?

rauchg:
@devinus definitely. That's why I want to post-process the color rather than having the user be surprised when they use a shorthand form.

devinus:
@rauchg Totally, but what I meant is should we figure out why the shorthand form is causing this?

rauchg:
@devinus sure, intrigued about that as well :D

mauro-oto:
@rauchg @devinus I submitted a PR with the problem and solution in https://github.com/zeit/hyperterm/pull/272. It would make this PR obsolete.

rauchg:
Thanks!

PR #223, opened 2016-07-18
remove unnecessary comment referencing use of !important due to aphrodite

conorhastings:


rauchg:
Actually this comment is irrelevant at this point. Would you mind removing it? Thanks for bringing this to my attention!

conorhastings:
@rauchg no problem! done.

rauchg:
:)

Issue #92, opened 2016-07-15
Hide header when showing only one tab


michaelmior:
The header is not needed when only one tab is open. Would be nice if this could be hidden in this case.

ppot:
@CodeTheory This should be closed.

PR #232, opened 2016-07-18
Prevents header from rendering with one tab

### LINKED ISSUE #127: Title bar gap on Linux/Windows is awkward when there are no tabs
The height of `.tabs_nav` should be zero when there are no tabs on non-OS X operating systems.

image

danprince:
Small tweaks to ensure that when there are <= 1 tabs, the `<Header />` component renders `null` rather than `<header>`. Also changes the style in `<Terms />` to make sure that the 34px top margin is only rendered if there are enough tabs open to justify it.

Example

For everyone using TMUX rather than native tabs, this should make hyperterm a viable option. I don't actually run OSX to test it on, so it's worth checking that the removed margin doesn't cause any visual accidents there.

Fixes #127 and #92.

I've been dreaming about a web powered terminal forever. Great job!

### DIFF
--- a/app/lib/components/header.js
+++ b/app/lib/components/header.js
@@ -60,6 +60,7 @@ export default class Header extends Component {
       onClose: this.props.onCloseTab,
       onChange: this.onChangeIntent
     });
+
     return <header
       style={{ backgroundColor }}
       className={ css('header', isMac && 'headerRounded') }
--- a/app/lib/components/terms.js
+++ b/app/lib/components/terms.js
@@ -114,9 +114,10 @@ export default class Terms extends Component {
   }
 
   template (css) {
+    const isFullScreen = this.props.isFullScreen;
     return <div
       style={{ padding: this.props.padding }}
-      className={ css('terms') }>
+      className={ css('terms', isFullScreen && 'tabsHidden') }>
       { this.props.customChildrenBefore }
       {
         this.props.sessions.map((session) => {
@@ -157,12 +158,20 @@ export default class Terms extends Component {
     return {
       terms: {
         position: 'absolute',
-        marginTop: '34px',
         top: 0,
         right: 0,
         left: 0,
         bottom: 0,
-        color: '#fff'
+        color: '#fff',
+        marginTop: '34px'
+      },
+
+      tabsHidden: {
+        marginTop: 0
+      },
+
+      tabsVisible: {
+        marginTop: '34px'
       },
 
       term: {
--- a/app/lib/containers/hyperterm.js
+++ b/app/lib/containers/hyperterm.js
@@ -70,12 +70,12 @@ class HyperTerm extends Component {
   }
 
   template (css) {
-    const { isMac, customCSS, borderColor } = this.props;
+    const { isMac, customCSS, borderColor, isFullScreen } = this.props;
     return <div onClick={ this.focusActive }>
       <div
         style={{ borderColor }}
         className={ css('main', isMac && 'mainRounded') }>
-        <HeaderContainer />
+        {isFullScreen ? null : <HeaderContainer />}
         <TermsContainer ref_={this.onTermsRef} />
       </div>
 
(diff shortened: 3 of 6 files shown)


rauchg:
What do you think about tweaking `padding` in the config instead? The reason I like the consistent height is that we don't trigger pty resizes when opening new tabs :) in addition to no jump visually

danprince:
I've got the padding turned down to 0 at the moment, but there's still a 34px block of space that can't be used for tiling wm/tmux users.

I work full time on a chromebook and spend most of that time inside a tiled vim window where it really sucks to lose 5% of the height.

How about allowing tabs to be disabled with a config option instead?

rauchg:
@danprince that sounds wonderful. And if we do it right, we can turn off the title extraction and make things even more efficient for your usecase.

I think it should be an advanced option, not present in `config-default.js`, but documented in the website (https://github.com/zeit/hyperterm-website).

Thanks!

danprince:
Hey @rauchg, here's a first attempt. How's this looking?

Imgur

I added a `fullScreenMode` option to the config format, which when set to true prevents the entire `<Header />` from rendering and adds a style to get rid of that margin. Not sure that `fullScreenMode` is necessarily the best name though, maybe `disableTabs` would work better?

rauchg:
I'm wondering about the implications of `tabless` mode a bit. What should we do about the keyboard shortcuts like cmd+T?

freebroccolo:
If this is implemented, please consider allowing the user to make the electron window frameless (`frame: false`) rather than using `titleBarStyle: "hidden-inset"`.

Being able to disable the frame (on OS X at least) is nice when using a tiling window manager like [kwm](https://github.com/koekeishiya/kwm). Without it, this PR hides the tab bar but still shows the OS X traffic lights buttons and in fact these buttons are shown on top of the terminal text which is not ideal.

Here's a screenshot of what it looks like if you use this PR and also disable the frame (i.e., replace `titleBarStyle: "hidden-inset"` with `frame: false`):

shot-opt

rauchg:
@freebroccolo I think the ability to decorate the window settings passed to `createWindow` is in order as a hook?

freebroccolo:
@rauchg ah, yes a plugin hook would be a nice way to handle this. That should be sufficient I think.

rauchg:
@freebroccolo if you want to add it in a PR, `decorateBrowserOptions` would be a good name I think. Pretty much same code as `decorateConfig`

freebroccolo:
> I'm wondering about the implications of tabless mode a bit. What should we do about the keyboard shortcuts like cmd+T?

Because of this I'm not sure it's really a good idea to allow the tab bar to be disabled completely as an option. I mean, I'd rather have that feature than nothing at all but the problem with disabling it completely is the UI becomes somewhat weird and surprising.

Even if the user purposely disables tabs, they could end up in a situation where they've accidentally created a new tab but there is no visual indication of this aside from whatever happens in their shell during login. Or maybe they create some tabs to do some task and then forget that the others exist.

This could be partially addressed by adding some other kind of indicator when a tab is created or when it's switched to, maybe like a temporary overlay in one of the corners, but that means additional work and complexity to support this feature nicely.

I guess you could just disable tabs completely in tabless mode and have `cmd + T` do nothing.

I think it would be better to not disable the tab bar completely, just have it hidden unless the number of tabs is `>= 2`. I'm probably biased though because this is a configuration I use in iTerm2 😄

danprince:
@freebroccolo I prefer the hide until >= 2 option, but if @rauchg isn't keen on doing the resize and repaint then I think that "tabless" mode needs to kill the `CMD+T` and `CMD+n` options for tab creation and navigation.

I don't have OSX to test on, but hiding the titlebar sounds sensible. My goal would be to get hyper term working as a drop-in replacement for URXVT or xterm, so that I could start using it as part of a tiling wm on a daily basis. Being able to achieve all of this stuff with a single "tiling mode" or "fullscreen mode" configuration option would be brilliant.

rauchg:
@danprince definitely open to helping support the tiling WM usecase, whether in core or userland. Just need to think a bit more about the implications

danprince:
Sure, not trying to suggest otherwise. What would the userland solution look like? It'd be possible to remove the header by decorating the component, but removing that margin is going to be harder. I'd be happy to move this PR out into a "hypertile" plugin if there's a solution which doesn't involve fiddling with the core.

danprince:
Userland solution on the way: https://github.com/danprince/hypertile. Once the `decorateBrowserOptions` hook exists in core, it'll be pretty complete.

rauchg:
Awesome. Let's focus on `decorateBrowserOptions` for the time being then. Thank you so much!

PR #242, opened 2016-07-18
Remove unnecessary/odd addEventListener in Term component unmount hook

developit:
Since the node is already being removed, the frame will be destroyed.


rauchg:
:)

PR #248, opened 2016-07-18
Expose 'winCount' and 'createWindow' to plugins

CWSpear:
I'm not sure this is the direction you wanna go, but this is at least a possibility that solves https://github.com/zeit/hyperterm/issues/226.

I tested it with my visor plugin, and I can use this to create a new window when the global hotkey is activated and there are no windows to create a new one.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,9 +13,12 @@ const config = require('./config');
 config.init();
 const plugins = require('./plugins');
 
+const windowSet = new Set([]);
+
 // expose to plugins
 app.config = config;
 app.plugins = plugins;
+app.getWindows = () => new Set([...windowSet]); // return a clone
 
 if (isDev) {
   console.log('running in dev mode');
@@ -40,7 +43,6 @@ app.on('window-all-closed', () => {
   // terminal is closed
 });
 
-let winCount = 0;
 app.on('ready', () => {
   function createWindow (fn) {
     const cfg = plugins.getDecoratedConfig();
@@ -60,7 +62,7 @@ app.on('ready', () => {
       show: process.env.HYPERTERM_DEBUG || isDev
     });
 
-    winCount++;
+    windowSet.add(win);
     win.loadURL(url);
 
     const rpc = createRPC(win);
@@ -195,7 +197,7 @@ app.on('ready', () => {
 
     // the window can be closed by the browser process itself
     win.on('close', () => {
-      winCount--;
+      windowSet.delete(win);
       rpc.destroy();
       deleteSessions();
       cfgUnsubscribe();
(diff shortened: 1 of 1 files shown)


rauchg:
Can plugin authors keep their own win count ? Since you can hook into `close` event?
Exposing `createWindow` is a must though.

rauchg:
`app.createWindow` is a must IMHO. If you can document it in https://github.com/zeit/hyperterm-website, doubly appreciated, but not a blocker.

CWSpear:
I _guess_ they could, but why put the extra burden when there can be one true source<sup>TM</sup>?

I can work on some docs, tho.

CWSpear:
See https://github.com/zeit/hyperterm-website/pull/25

rauchg:
@CWSpear because exposing a count is kind of weird in my opinion. A set of window references perhaps? `getWindows`?

rauchg:
An immutable set would be nice I think :)

CWSpear:
I'm not really convinced it needs to be an Immutable! Since it's not really ever compared, Immutable-izing it doesn't really have any performance gains, but adds code/mental overhead.

But I did update it using a regular array. Does this look alright?

CWSpear:
I updated the docs to reflect these changes.

rauchg:
Thanks @CWSpear. Maybe let's make it a `Set` instead?

CWSpear:
Alright, how does _that_ look?

index.js
```diff
@@ -13,9 +13,12 @@ const config = require('./config');
 config.init();
 const plugins = require('./plugins');
 
+const windowSet = new Set([]);
+
 // expose to plugins
 app.config = config;
 app.plugins = plugins;
+app.getWindows = () => [...windowSet];
```
rauchg: Do you think returning the `Set` directly is a bad idea?

rauchg:
Almost there. Thank you so much

index.js
CWSpear: I do. If they remove/add/modify something from `Set` without cloning it, it could cause issues. We could create a _new_ `Set` (i.e. clone it) and return that if you'd like.

CWSpear: (But I feel like `Array` is the lowest common denominator that gets the job done. Most people will be more familiar with that than a `Set`. If they really want, they can easily make it a `Set` themselves in plugin-land.)

rauchg: A new `Set` sounds good. I think if we internally use a `Set`, it makes sense to expose a `Set`. Good observation. Thank you

CWSpear:
Ok, it now returns a new set (with a comment saying the goal is to "return a clone").

rauchg:
🎉

CWSpear:
Awesome! Thanks @rauchg

Issue #108, opened 2016-07-15
Default rows and cols? 


indexzero:
> _Apologies if this is already documented somewhere, but I did search around a bit and read some of the code before opening this issue._

First off: 👍👍👍 Very well done. 🐴 🏇 

 `140x30` is my preferred default terminal size and I cannot seem to find the setting in my `~/.hyperterm.js` configuration file that I should modify to change the default terminal size.

indexzero:
This is related to, but distinctly different from #27. In my own case I don't want HyperTerm to remember the window size. I want every window to always start at the same size, regardless of if I have resized it.

tamagokun:
Thinking about how to implement this, because it is a tad bit tricky with electron. The way it works is you create a new browser window and set the width/height. Then, the term itself will determine how many cols/rows it has based on its own size inside the browser window. So, here are few ways to handle this:

1. Config option for default browser size, expressed in pixels. Not very "term"-like, but it is easy to implement.

2. Try to calculate cols/rows -> pixels on window creation. This is the *ideal* implementation, though it is difficult to determine the ending size of the window. Things like `fontSize` `padding` `margin` `borders` need to be taken into consideration.

3. Calculate cols/rows -> pixels when main component mounts then resize window. This would make it much easier to calculate the target window size since we can work with the DOM, but then there is a noticeable delay from window creation to component mount which isn't exactly pretty.

tamagokun:
Thinking about option 3 some more, it would be possible to use electron's `ready-to-show` event on the window, but then there would be nothing showing until the rows/cols can be calculated.

@rauchg I haven't done a whole lot with electron, wondering if you have any insight on an implementation for having default rows/cols?

aabril:
👍

brunjick:
@tamagokun I think 1st option would be enough at this moment, so we will be able to adjust pixels ourselves

tamagokun:
going with option 1 for now. PR submitted

indexzero:
Awesome! Thanks @tamagokun 💯

jokeyrhyme:
Great work here. :)

I'd also like to voice support for defining the default size in terms of columns/rows. GNOME Terminal and the macOS Terminal both support this. I wonder how they achieve this?

Maybe we need a system service / agent that calculates the target width/height based on the selected font and size in the background, prior to hyperterm launching?

What if we calculate it and cache the value based on the font/size and columns/rows? So the first launch is slightly slower, but every launch after that is not any slower?

[...]

PR #257, opened 2016-07-18
Support config for default width/height of browser window

tamagokun:
This addresses #108 (sort of). Ideally I think rows/cols make more sense since this is a terminal application. This at least gives the user something to work with.

### DIFF
--- a/index.js
+++ b/index.js
@@ -48,10 +48,11 @@ app.on('window-all-closed', () => {
 app.on('ready', () => {
   function createWindow (fn) {
     const cfg = plugins.getDecoratedConfig();
+    const [width, height] = cfg.windowSize || [540, 380];
 
     const browserDefaults = {
-      width: 540,
-      height: 380,
+      width,
+      height,
       minHeight: 190,
       minWidth: 370,
       titleBarStyle: 'hidden-inset',


rauchg:
I like the idea of just memorizing the last size the window was open. What do you think?

tamagokun:
For anyone that creates more than one window for one-off commands or having two windows side-by-side would probably want to set a "default" for creating new windows.

I may not understand the logic fully of saving the last window size, but let's say I normally have my windows at 900x450 and then open a new window and resize it to fit in the bottom right corner of my screen for a quick command, or maybe a plugin is creating a new window for me, then that window would overwrite the last size and my next new window would be messed up and i'd have to resize it back to my preferred default. If I have defaults, I don't have to worry about resizing windows messing with my preferences.

So, thinking through the logic for window creation dimensions:

`windowSize` -> `last window size` -> fallback

So, I would actually remove `windowSize` from the default config, and have that be an optional config that users can set if they prefer to always create new windows with their preferred size.

rauchg:
Makes a lot of sense… Since it's a bit of advanced flag, maybe we should take it out of the default file? And we can document it here: https://github.com/zeit/hyperterm-website

tamagokun:
Yep, exactly. Removing it from the default config should make it work nicely with memorizing window size and cover a lot of user preferences.

tamagokun:
Done! PR over on https://github.com/zeit/hyperterm-website created as well.

rauchg:
Sorry, one last rebase 🙏

tamagokun:
all set 🎊

PR #259, opened 2016-07-18
dont bundle with webpack to get faster startup?

nfcampos:
@rauchg why are we bundling the renderer code with webpack?
by removing it the startup time goes down like 25% (on my machine at least)
am I missing anything?

(only thing missing is makign sure the bundled app runs the main js file with `NODE_ENV=production` this is maybe already true?)


rauchg:
Are you measure with `NODE_ENV=production webpack`? That performs dead code elimination and uglification, which are massive wins for startup time.

nfcampos:
ah, yeah you're right, I was measuring the time without minifying. nevermind.

PR #261, opened 2016-07-18
Move to the last tab with cmd + 9

### LINKED ISSUE #255: Make Command+9 go to the last tab
Web browsers and iTerm2 all go to the last tab (not necessarily the ninth tab) when command+9 is inputted. It would be nice to be consistent with them.
Relevant line:
https://github.com/zeit/hyperterm/blob/fa4033937e3b80554c037efd357b5fd6159372cf/app/lib/containers/hyperterm.js#L45

ekmartin:
This mimics the behavior of Chrome, Firefox, iTerm 2 and similar, by moving to the last possible tab when Command + 9 is pressed.

Fixes #255.

### DIFF
--- a/app/lib/containers/hyperterm.js
+++ b/app/lib/containers/hyperterm.js
@@ -32,6 +32,7 @@ class HyperTerm extends Component {
     const { moveTo, moveLeft, moveRight } = this.props;
     const term = this.terms.getActiveTerm();
     if (!term) return;
+    const lastIndex = this.terms.getLastTermIndex();
     const document = term.getTermDocument();
     const keys = new Mousetrap(document);
     keys.bind('command+1', moveTo.bind(this, 0));
@@ -42,7 +43,8 @@ class HyperTerm extends Component {
     keys.bind('command+6', moveTo.bind(this, 5));
     keys.bind('command+7', moveTo.bind(this, 6));
     keys.bind('command+8', moveTo.bind(this, 7));
-    keys.bind('command+9', moveTo.bind(this, 8));
+    keys.bind('command+9', moveTo.bind(this, lastIndex));
+
     keys.bind('command+shift+left', moveLeft);
     keys.bind('command+shift+right', moveRight);
     keys.bind('command+shift+[', moveLeft);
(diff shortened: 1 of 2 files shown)


app/lib/containers/hyperterm.js
```diff
@@ -42,7 +43,13 @@ class HyperTerm extends Component {
     keys.bind('command+6', moveTo.bind(this, 5));
     keys.bind('command+7', moveTo.bind(this, 6));
     keys.bind('command+8', moveTo.bind(this, 7));
-    keys.bind('command+9', moveTo.bind(this, 8));
+    keys.bind('command+9', () => {
+      const lastTerm = this.terms.getLastTerm();
+      if (lastTerm) {
```
ekmartin: I'm not sure if this check is necessary - can there be a scenario where this can be reached without at least one active term being available?

app/lib/containers/hyperterm.js
```diff
@@ -42,7 +43,13 @@ class HyperTerm extends Component {
...
+        setActiveSession(lastTerm.uid);
```
ekmartin: Should this be wrapped in an UI action creator, similar to `moveTo`/`moveLeft` and so on? Maybe to avoid calling setActiveSession if the last one is already active (although this can be checked here as well).

rauchg:
@ekmartin I would keep the container component not too logic-heavy, and instead dispatch something like `MOVE_TO_LAST`. Sounds good?

rauchg:
I'm labeling just for the sake of organization. Take your time to refactor or ask any questions you need!

app/lib/containers/hyperterm.js
albinekb: Can't you just use `moveTo`? like this
```
keys.bind('command+9', moveTo.bind(this, this.terms.getLastTermIndex())
```

```
 getLastTermIndex () {
    return this.props.sessions.length - 1;
  }
```

ekmartin:
Sorry for the delay, didn't have time to look at it before now. I've changed it to @albinekb's clean suggestion. @rauchg, if you still think it's a tad too much logic for the container I could add an action creator instead?

rauchg:
Beautiful. An action would have been only very marginally better. I like this.

PR #267, opened 2016-07-18
Make preference-opening command universal

szhu:
The `$EDITOR ~/.hyperterm.js && exit` requires that the default shell recognize variables as commands and to recognize the `&&` syntax, so it does not work with fish shell. The new command works with all common shells, including fish.

Old behavior with `fish`:

screen shot 2016-07-18 at 4 05 52 pm


rauchg:
Brilliant. Thank you

PR #271, opened 2016-07-19
Add reducer to set cwd

hharnisc:
This allows the `cwd` to be set with the dispatcher in a common way. This commit is proactive in that it's anticipating plugins are going to want to set the `cwd`.

An example of dispatch action that would trigger a `cwd` change

```
const cwd = '/';
store.dispatch({
     type: 'SESSION_SET_CWD',
     cwd
});
```


rauchg:
Super smart! Thank you

PR #272, opened 2016-07-19
Override the buggy hexToRGB implementation in hterm.

mauro-oto:
Hey @rauchg, 

This PR fixes the issue we've seen with colors and discussed in #220. The problem is in https://github.com/dbkaplun/hterm-umdjs/blob/master/dist/index.js#L343-L376. 

Instead of converting the hex "FF" to decimal (255), it was converting just "F", which resulted in 15, and the gray `rgb(15, 15, 15)` color instead of `rgb(255, 255, 255)`. The shorthand was not being expanded properly. 

Let me know if I should move the code elsewhere from hterm or if we should solve this problem some other way. 

Please check it out, thanks!

### DIFF
--- a/app/lib/hterm.js
+++ b/app/lib/hterm.js
@@ -86,5 +86,37 @@ hterm.Terminal.prototype.clearPreserveCursorRow = function () {
   this.scrollPort_.redraw_();
 };
 
+// fixes a bug in hterm, where the shorthand hex
+// is not properly converted to rgb
+lib.colors.hexToRGB = function(arg) {
+  var hex16 = lib.colors.re_.hex16;
+  var hex24 = lib.colors.re_.hex24;
+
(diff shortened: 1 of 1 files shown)


rauchg:
This is the best possible fix you could have come up with. Thank you so much. We should notify the chromium folks! I would drop them a note here: https://groups.google.com/a/chromium.org/forum/#!forum/chromium-hterm

dbkaplun:
Opened a ticket upstream: https://groups.google.com/a/chromium.org/d/msg/chromium-hterm/U1JR0Pfi43Y/Or7ucUbGAQAJ

mauro-oto:
@dbkaplun thanks! I submitted a patch here: https://chromium-review.googlesource.com/#/c/362160/

PR #273, opened 2016-07-19
add homebrew cask install method

ZakariaRidouh:


rauchg:
Superb. I'm also down for adding it to the website under `# Installation`

Issue #277, opened 2016-07-19
Command to install hyperterm plugins


mikeal:
It's kind of clunky to install plugins, it would be great if there were just a hypterm command line to do that kinda stuff.

```
hyperterm install hypertheme
```

mxstbr:
`hpm` by @matheuss is trying to solve that, not sure if that should be in core? https://github.com/matheuss/hpm

matheuss:
I think it would be very cool if `hpm` came along with HyperTerm 😊 Atom, for instance, ships `apm` with the app.

mxstbr:
@matheuss submitted a PR for that (#282), let's see what @rauchg has to say about that!

rauchg:
I'm interested, but I'd like to understand the clunkiness more. Right now you edit a file, you save it, it hot-reloads.

matheuss:
Well, is faster and easier to install a package with a little command 😅 When I first thought of writing `hpm`, I thought "HyperTerm needs something like [`apm`](https://github.com/atom/apm)" 

![hpm](https://cloud.githubusercontent.com/assets/4721750/16961000/02403c48-4dc2-11e6-8a83-a1412f3ef019.gif)

booc0mtaco:
Although it can be assumed that anyone using hyperterm probably understands how to edit the file, something like `hpm` would:

- provide an external mechanism to check for package validity (or search a la `brew search`)
- make scripting a bit easier for dotfile repositories
- drastically reduce the likelihood of typos corrupting the config file entirely during (un-/re-)install

Just a few additional thoughts

matheuss:
Just an update: now we have more commands 😊 
```
❯ hpm --help
[...]
    s, search <query>      Search for plugins on npm
    ls-remote              List plugins available on npm
    d, docs <plugin>       Open the npm page of the <plguin>
```

<img src="https://raw.githubusercontent.com/matheuss/hpm/master/screenshot.gif?v=2" width="629">

tunnckoCore:
In anyway, nowhere in the docs or site is said how to install it. I just see `~/.hyper_plugins` folder, but how packages goes there? Should i just `npm install` the plugins globally or what?

matheuss:
You can add nom package names to the `plugins` array on `.hyper.js` 👌

chabou:
It is planned to ship `hpm` with Hyper

SpaceNoodles01:
After i install a plugin, what i must do to show up what actually the plugin does?

[...]

PR #282, opened 2016-07-19
Install hpm by default

mxstbr:
This is an easy solution to issue #277. Installs hyperterm package manager (hpm) by @matheuss after installing hyperterm.

Further discussion possibly needed, maybe this should be in core?

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     }
   },
   "scripts": {
+    "postinstall": "npm install -g hpm",
     "start": "electron index",
     "lint": "eslint *.js"
   }


amilajack:
Nice parallel to atom

mathieudutour:
Hum when you download the release directly, there is no `npm install` done so `hpm` won't be installed

mxstbr:
True, that's kinda annoying. I wonder how Atom does this with `apm`?

amilajack:
`apm` isn't installed when running `npm I -g hpm`. This assumes that the user has `npm` installed globally, and that isn't a requirement to install atom. Run `which apm` or `which apm-beta` and you'll see the executable in the `/usr/local/bin/apm-beta` directory while modules installed with npm are (for nvm users at least) in the `~/.nvm/versions/node/${node_version}/bin/`. The binary files seem to have symbolic links to their corresponding javascript located in `../lib/node_modules/${module_folder}/${file}`. Ex:

```
npm -> ../lib/node_modules/npm/bin/npm-cli.js
```

amilajack:
Also note that `apm` isn't listed when running  npm list -g --depth 0.

matheuss:
I think that the best thing we can do is something like:
```sh
command -v npm >/dev/null 2>&1 ||  echo >&2 "You must install npm etc etc";
```
If the user does not have node and/or npm, he will not be able to install any plugins anyway.

matheuss:
`apm` comes bundled with atom: 
```
❯ ~ which apm | xargs ls -l
/usr/local/bin/apm -> /Applications/Atom.app/Contents/Resources/app/apm/node_modules/.bin/apm
```
I think it might be a good approach for us. Then `hpm` can check for npm and provide instructions (install `nvm` etc) if it is not installed 😊

rauchg:
Yeah, I think it's looking pretty great. I need to do more testing, so we'll reserve this for the next release.

rauchg:
Things we discussed that are outstanding concerns for this:
- How do we install it in a location that we know is in $PATH?
- How do we keep it updated? Ideally we pin it, and with each new release of HyperTerm we can bump it, and it updates `hpm` automatically
- If we can't find a location in $PATH we can write to, how do we tell the user? (ie: notification)
 - Where in the menu do we display "Install hpm` in that scenario?
- How do we tell users when we install it that they can now run `hpm` if needed?
  - Maybe we have an onboarding screen with a little html file that we also use to explain the configuration system?

sindresorhus:
> How do we install it in a location that we know is in $PATH?

`$ which npm`

matheuss:
Also: we need `node`. I'm not completely sure, but I think that HyperTerm can run without `node`, can it?

lordgiotto:
We can use the node executable bundled in electron I guess.

rauchg:
@sindresorhus we anticipate scenarios where npm is not installed, so we also have to consider that

sindresorhus:
Good point. You already have the user's $PATH from the `shell-env` module though, so you can just use any path defined there.

sheerun:
Why not bundling npm along hyperterm? No need to use system one.

(also when bundling npm, you can fix the version, e.g. to 3.x)

sheerun:
I've submitted #381 that embeds npm, but I believe hyperterm shoudn't install hpm externall, but rather install it as dependency, and link it to system [similarly](https://github.com/atom/atom/blob/master/spec/command-installer-spec.coffee) to [atom](https://github.com/atom/atom/blob/master/src/main-process/squirrel-update.coffee#L33-L55).

sheerun:
PR #383 on to of global hotkey functionality, includes ability to embed default plugins into hyperterm. I think this should be primary way to embedding hpm. One thing remaining is to create system links for `hmr` binary upon installation, but unfortunately I won't have time to implement it.

matheuss:
+1 for shipping `npm` with the app

leo:
@matheuss https://github.com/zeit/hyperterm/pull/172#issuecomment-233174361

MindTooth:
Guess this can be closed now with 2.0.0-canary9?

https://github.com/zeit/hyper/commit/5700690e0bf70d004c0b2ae35c01a8e8e2db3a20

leo:
🎉 Big applause to @chabou! 👏 👏

PR #286, opened 2016-07-19
Add middle click to close tab functionality

itswil:
On clicking a tab, "switch to tab" if the mouse is left-clicked, otherwise "close the tab" if the mouse is middle clicked

### DIFF
--- a/app/lib/components/tab.js
+++ b/app/lib/components/tab.js
@@ -6,6 +6,7 @@ export default class Tab extends Component {
     super();
     this.hover = this.hover.bind(this);
     this.blur = this.blur.bind(this);
+    this.handleClick = this.handleClick.bind(this);
     this.state = {
       hovered: false
     };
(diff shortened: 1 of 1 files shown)


app/lib/components/tab.js
```diff
@@ -41,7 +49,7 @@ export default class Tab extends Component {
             isLast && 'textLast',
             isActive && 'textActive'
           ) }
-          onClick={ this.props.isActive ? null : this.props.onSelect }>
+          onClick={ this.handleClick.bind(this) }>
```
rauchg: Binding every render has a performance penalty. Maybe create a bound version of `handleClick` in the constructor?

itswil:
`handleClick` now bound in constructor (not `render()`)

app/lib/components/tab.js
```diff
@@ -19,6 +20,14 @@ export default class Tab extends Component {
     this.setState({ hovered: false });
   }
 
+  handleClick (event) {
+    if (event.nativeEvent.which === 1) {
+      this.props.isActive ? null : this.props.onSelect();
+    } else if (event.nativeEvent.which === 2) {
```
marcbachmann: I'd assign that to a variable, so it's more descriptive.
e.g. `const isMiddleClick = event.nativeEvent.which === 2`

marcbachmann: Same with the other variable `const isLeftClick = event.nativeEvent.which === 1`

rauchg: +1

itswil:
Mouse click native events now assigned to descriptive variables

marcbachmann:
Nice 👍

itswil:
If you use BetterTouchTool (or equivalent) you could map a middle click to a gesture 😄

marcbachmann:
👍 Everything is working.
@rauchg that's ready to merge.

PR #291, opened 2016-07-19
Fix installing plugins that use `node-gyp`

dfrankland:
Sets the proper environmental variables to install plugins that have
specific needs for compilation e.g. `nodegit`.

I came across this issue with a plugin I recently created [`hyperterm-sync-settings`](https://github.com/dfrankland/hyperterm-sync-settings) which would build dependencies from source improperly. This change fixes that.

### DIFF
--- a/plugins.js
+++ b/plugins.js
@@ -194,7 +194,7 @@ function install (fn) {
     let registry = exports.getDecoratedConfig().npmRegistry;
     if (registry) env.NPM_CONFIG_REGISTRY = registry;
     env.npm_config_runtime = 'electron';
-    env.npm_config_target = '1.2.5';
+    env.npm_config_target = require('./package.json').devDependencies['electron-prebuilt'];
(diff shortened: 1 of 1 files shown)


plugins.js
```diff
@@ -193,6 +193,9 @@ function install (fn) {
   shellEnv().then((env) => {
     let registry = exports.getDecoratedConfig().npmRegistry;
     if (registry) env.NPM_CONFIG_REGISTRY = registry;
+    env.npm_config_runtime = 'electron';
+    env.npm_config_target = '1.2.5';
```
rauchg: Maybe we can read the setting from our own build to use a matching electron version?

rauchg:
Good stuff @dfrankland

plugins.js
dfrankland: Can do, it'll be easy to read from the `package.json`. I'll update the pull soon.

dfrankland:
Alright, should be good to go 👍

rauchg:
👏

Issue #97, opened 2016-07-15
URL Handling - Unable to Open localhost Webpage


joemccann:
When I type `nodesource.com` it opens our the webpage as one would expect.

When I type `http://0.0.0.0:3000` I get `zsh: no such file or directory: http://0.0.0.0:3000`.

I dug into the source code for the way URLs are handled and feel like the culprit may be there?  RegEx related?

bnb:
Definitely seems like localhost pages would be `A+` for locally-running Node apps and development, not to mention just about any other web platform that you'd be developing locally.

rauchg:
Yep, gotta improve the regex!

rauchg:
Good catch @joemccann

joemccann:
You bet @rauchg 

Love this project!

7373Lacym:
FYI If you need this now it looks like ```www.localhost``` works.

Issue #98, opened 2016-07-15
Can't open http://localhost:3000


langolf:
It seems we cannot open url without domain zone. But I really need :)

Also hyperterm crashes when I try add `.com` in the end
`http://localhost:3000.com`

kinduff:
👍 Able to reproduce. Unable to open localhost y crashes with dot com at the end.

kinduff:
This is a duplicate of #97, the `.com` at the end should be also fixed, not sure if that bug is part of the same issue.

iamstarkov:
just ran `open http://localhost:3000` and that works perfectly fine

langolf:
@iamstarkov We are talking about opening inside hyperterm as normal urls. Hyperterm already can do this with normal urls.

iamstarkov:
@langolf what do you mean?

langolf:
@iamstarkov you can open any url inside hyperterm window like a browser
https://monosnap.com/file/jzAKLD1oI3wPAi5U3Go7r17HivvOTk

iamstarkov:
how did you do this?

langolf:
@iamstarkov just type `http://someurl` or `www.someurl` in shell

iamstarkov:
thats neat, how do use it?

langolf:
I use vim for development. So it's very convenient to have "browser" in next-tab. E.g. I don't need to cmd-tab to browser to check markup.

kopollo:
try from here [http://localhost.support/:3000](http://localhost.support/:3000)

PR #305, opened 2016-07-20
allow opening of files in webview

### LINKED ISSUE #304: https sites load as http
One of the cool features of this app is that URLs automatically open a webview. The problem is that if the URL is https it will load it as http. For most websites, you will not be able to tell since they redirect to https anyway.

cuth:
If a URL is entered as a command, that URL is loaded in a webview. This PR will allow the file protocol `file://` to also be considered as a URL. This will allow you to open images, videos, music and more right inside the webview.

By adding non-capturing groups to the `url-regex.js` file, this made the `url-command.js` simpler.

This also fixes #304

### DIFF
--- a/app/lib/utils/url-command.js
+++ b/app/lib/utils/url-command.js
@@ -4,47 +4,23 @@ export const domainRegex = /\b((?=[a-z0-9-]{1,63}\.)(xn--)?[a-z0-9]+(-[a-z0-9]+)
 
 export default function isUrlCommand (shell, data) {
   const matcher = regex[shell];
-  if (undefined === matcher) return null;
-  let url, i;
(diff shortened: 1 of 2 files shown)


rauchg:
😍

Thanks @cuth. You might also want to take a look at tackling the issue related to IPs and ports ;)

cuth:
This should also fix #97 and #98 because any string that begins with `http://`, `https://`, `file://` or `//` is recognized as a URL and doesn't use the domain regex test.

Also, just note that the [default background color the webview is black](https://github.com/zeit/hyperterm/blob/0.6.0/app/lib/components/term.js#L172) so if all you have on the webpage is black text, you will not see anything.

PR #310, opened 2016-07-20
Adds plugin hook for decorating the electron browser options

danprince:
Creates a new plugin hook `decorateBrowserOptions` as discussed in #232. It works just like `decorateConfig` but for the browser options that Electron expects in the [`BrowserWindow`][1] constructor.

Designed to be used [like this][2].

Default options are specified inline, but I guess they could be moved out like `config-default.js` if necessary.

[1]: https://github.com/electron/electron/blob/master/docs/api/browser-window.md#new-browserwindowoptions
[2]: https://github.com/danprince/hypertile/blob/master/index.js#L22-L25

### DIFF
--- a/index.js
+++ b/index.js
@@ -46,9 +46,8 @@ app.on('window-all-closed', () => {
 app.on('ready', () => {
   function createWindow (fn) {
     const cfg = plugins.getDecoratedConfig();
-    const opts = plugins.getDecoratedBrowserOptions();
 
-    const browserOptions = Object.assign({
+    const browserDefaults = {
       width: 540,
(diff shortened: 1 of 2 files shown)


rauchg:
Usually the decorators get the existing options, and further enhance them (hence "decoration").
You should pass the object to `getDecoratedBrowserOptions`
Thank you!

danprince:
Makes sense. Changed it so that `getDecoratedBrowserOptions` receives the defaults as an argument.

rauchg:
Beautiful!

PR #314, opened 2016-07-20
Specify bash to run command in

kyleholzinger:
Could be useful for people who use a non-POSIX compatible shell, such as fish. I think it's common enough that this could be useful, and since it's Mac only, I think it's safe to say that the user has bash installed

### DIFF
--- a/app/lib/actions/ui.js
+++ b/app/lib/actions/ui.js
@@ -145,7 +145,7 @@ export function showPreferences () {
               uid,
               ['echo Attempting to open ~/.hyperterm.js with your \$EDITOR',
                'echo If it fails, open it manually with your favorite editor!',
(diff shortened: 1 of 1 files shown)


rauchg:
@kyleholzinger did this command not work on fish? I believe it's been tested on both bash and zsh already.

kyleholzinger:
Nope, you can't use variables as commands in fish

kyleholzinger:
Oh, in my downlaoded version of hyperterm it tries to do `$EDITOR ~/.hyperterm.js`, but in master it's `exec env $EDITOR ~/.hyperterm.js`. Is the download from the site just lagging behind?

marcbachmann:
> Is the download from the site just lagging behind?

Yes, master doesn't get released automatically.

PR #319, opened 2016-07-20
Update dependencies

freebroccolo:
This PR updates most of the dependencies except for `child_pty`.

I added `eslint-plugin-react` to the root package because otherwise eslint would sometimes complain about not being able to find the plugin when running it manually.

On my system, building with `child_pty` using version `v3.0.3` or `v3.0.4` caused `CTRL-C` to stop working. However, `CTRL-C` does work properly again if you build from master. Since [Gottox/child_pty@f416b22](https://github.com/Gottox/child_pty/commit/f416b22a3a004b036316c6708c878115bf7117d1) is the only post-`v3.0.4` commit, it appears that some termios flags (probably `ISIG`, maybe others) are not being set properly.

I did try to fix that by calling `setattr` after the pty is created with `child_pty.spawn()` in `session.js` but still couldn't get it to work. In any case, the termios stuff has been removed after `v3.0.4` so it's probably not worth bothering with.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -6,7 +6,7 @@
   "repository": "zeit/hyperterm",
   "dependencies": {
     "aphrodite-simple": "0.4.1",
-    "hterm-umdjs": "1.1.2",
+    "hterm-umdjs": "1.1.3",
     "json-loader": "0.5.4",
     "mousetrap": "1.6.0",
     "ms": "0.7.1",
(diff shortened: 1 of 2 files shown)


rauchg:
CCing @gottox to he knows

rauchg:
Thank you!

PR #321, opened 2016-07-20
Fix linter issues

freebroccolo:
This PR does the following:

- Adds `.eslintignore` files
- Switches to `eslint .` instead of `eslint *.js`
- Fixes some linter issues that were not being reported without `eslint .`


rauchg:
Superb.

PR #326, opened 2016-07-20
Disable pinch-zoom functionality

### LINKED ISSUE #156: Zoom In/Out Bugs
So with MacOS you can use two fingers on your touchpad to zoom in/out. With hyperterm you're able to zoom in using this gesture, but you're not able to zoom out. Which leaves you with something like:

screen shot 2016-07-16 at 1 08 45 am

To replicate: have a windowed terminal, (note that zooming in/out works on fullscreen) then try to zoom in/out on anywhere BESIDES the top-left corner (for some reason it works at this spot even when windowed).

Also a side note: the default terminal doesn't support pinch to zoom.

marcbachmann:
This fixes #156

We don't have yet a specific config file/location for the renderer.
Should I keep it in the index.js file?


rauchg:
Yep that sounds great.

PR #327, opened 2016-07-20
Use node6 es2015 features; remove extra transforms

freebroccolo:
This PR removes some unnecessary babel transforms. Since the current electron ships with node 6 it already supports over 95% of the es2015 features out of the box (details [here](http://node.green/) and [here](https://kangax.github.io/compat-table/es6/)).

I've added the `--harmony` flag in the app to enable a few of the remaining non-defaulted es2015 features in node 6 (mostly related to `Function.name` and `Symbol`). But if you'd rather not use `--harmony` for some reason it probably won't hurt since these last few features are rarely used.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -26,7 +26,6 @@
     "babel-core": "6.10.4",
     "babel-eslint": "6.1.2",
     "babel-loader": "6.2.4",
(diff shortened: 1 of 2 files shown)


rauchg:
:O!

rauchg:
@freebroccolo you're the best 🎉

PR #328, opened 2016-07-20
Hook scripts

freebroccolo:
This PR does the following:

- Add `husky` as a development dependency
- Set hook scripts for `prepush` and `prepublish` to run `npm test`
- Set `npm test` to at least call `npm run lint` for now
- Fix a linter issue in `menu.js`

The hook scripts for git (e.g., `prepush`) will be setup when the user runs `npm install`.

With the `prepush` hook, git will run `npm test` prior to pushing to a repo. Could also do `precommit` but it's overkill and can make complex rebase operations more difficult.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -73,6 +73,7 @@
   "scripts": {
     "dev": "webpack --watch",
     "lint": "eslint .",
-    "build": "NODE_ENV=production webpack"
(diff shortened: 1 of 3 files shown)


amilajack:
Good idea

rauchg:
Nice!

PR #329, opened 2016-07-20
Handling file opening

lordgiotto:
Hi,
this is a pull request based on @kirb pull request #276 

Now that #307 is merged, we can drop files on the dock icon, but we have to handle them app side.

@Kirb, in his commits, has created 'open file' event, openFile function and escapeShellArgument function.
___
Shown below I'll describe all the changes I've done to fix all the issues:
- 486081f adds the property `focusTime` to `win` object that store the last process.uptime() when it was focused. It also add the method `getLastFocusedWindow` to `app` object that let get the last focused window.
- 6bb3bd3  changes the init process: I removed the `requestSession()` inside `init`in actions/index.js and replaced it with the `'init'` event emit. The creation of the session is now responsability of the callback passed to `createWindow` function. If no callback is passed, it will be setted as `rpm.emit('session add req')`, like before. _Doing this way permits you to choose how to create the session on init_ and fix the problem of double session on startup if a file is opened.
- b1861c3 adds the property `windowCallback` to app object that will be called instead of createWindow callback and deleted right after. This permits to handle the file drop before that `'ready'` event occurs.
- 51f5d97 improves a bit how @kirb designed the actual file opening: if it's directory or it's an executable file it will be pushed to data with a return; otherwise it will be pushed to data without the return.

____

I hope you approve the style and the design, otherwise let me know and I'll be happy to fix it.


index.js
```diff
@@ -75,7 +85,9 @@ app.on('ready', () => {
 
     rpc.on('init', () => {
       win.show();
-      if (fn) fn(win);
+      if (!fn) fn = (win) => win.rpc.emit('session add req');
+      (app.windowCallback || fn)(win);
```
rauchg: This is a bit obscure. Comments here would be nice.

rauchg:
Looking so awesome so far!

index.js
lordgiotto: Do you mean comment in the code or an explanation here?
___
The first line (line 88) sets `createWindow` callback to `win.prc.emit('session add req')` if no callback is set: this means that by default `createWindow` will create a new session, except if a callback is passed (in our specific case the callback setted by `'open-file'` app event will fire `'open file'` rpc event, and `'session add req'` will not be fired).
___
The second line (line 89) execute `app.windowCallback` (the callback setted before `createWindow` is even defined, in our specific case in `'open-file'` app event, that fires before the `'ready'` event) or the `fn` `createWindow` callback if `app.windowCallback` is not defined. 

**`(app.windowCallback || fn)(win);`** 
is exactly the same of writing 
**`app.windowCallback ? app.windowCallback(win) : fn(win);`**
___
Hope to have clarified the code, it's not so easy to explain :)

app/lib/utils/file.js
```diff
@@ -0,0 +1,10 @@
+// Taken from https://github.com/kevva/executable and modified
+export function isExecutable (fileStat) {
+	if (process.platform === 'win32') {
+		return true;
+	}
```
rauchg: Our indent level is two spaces.

index.js
```diff
@@ -199,6 +211,10 @@ app.on('ready', () => {
       }
     });
 
+    win.on('focus', () => {
+      win.focusTime = process.uptime();
```
rauchg: I think a comment here (in the code) of why we store the focus time as opposed to keeping a variable around would be nice! (I assume it has to do with windows quitting etc)

index.js
```diff
@@ -250,3 +266,15 @@ function initSession (opts, fn) {
     fn(uid, new Session(opts));
   });
 }
+
+app.on('open-file', (event, path) => {
+  let lastWindow = app.getLastFocusedWindow();
```
rauchg: This can be a `const`. `getLastFocusedWindow` doesn't need to necessary live in `app`

app/lib/utils/file.js
lordgiotto: Oh right, sorry. Fixed! :P

index.js
lordgiotto: I decided to use this approach because it's easier and cleaner in my opinion: to handle windows closed and removed, we should store not only the last focued win, but also the previous ones to fallback in case of window closing. Doing this with process.uptime() way allow us to not have to handle window closing creating complex logic to order, delete and update a _focusedWindows_ array.

index.js
lordgiotto: Fixed the const.
I added getLastFocusedWindow to app in order to expose it to plugins developers: it could be useful also to them i guess :)

lordgiotto:
Added various comments and done small fixes.

app/lib/actions/index.js
```diff
@@ -1,7 +1,5 @@
-import { requestSession } from './sessions';
+import rpc from '../rpc';
 
 export function init () {
-  return (dispatch) => {
-    dispatch(requestSession());
-  };
+  rpc.emit('init');
```
rauchg: so, before we were delaying `init` until we got a session going. The reason for this is that we don't want to `show` the window until the user has something to do pretty much (i.e: a prompt). Why was this changed?

lordgiotto: With my PR it's the same, but the request session creation is relying on createWindow callback: 
in master on `init()` is called `requestSessions()` that fire `rpc.emit('init')` if `const initial = null == sessions.activeUid;`; 
in my PR on `init()` is fired `rpc.emit('init')`, that create the session through the createWindow callback `rpc.emit('session add req')`, so essentially is the same but `initial`check is not needed.

I've done this because a blank session spawned on createWindow is not always the wanted behaviour, like in our case where we want to call `rpc.emit('open file', { path })` in place of a blank session. Otherwise two session would be created on createWindow event.

app/lib/utils/file.js
```diff
@@ -0,0 +1,10 @@
+// Taken from https://github.com/kevva/executable and modified
```
sindresorhus: Modified how and why?

app/lib/utils/string.js
```diff
@@ -0,0 +1,3 @@
+export function escapeShellArgument (text) {
```
sindresorhus: Pretty sure this exists as an existing module on npm.

app/lib/utils/file.js
lordgiotto: I take only the part that check the file mode, not the part that make the game.statSync (to not do it twice), and removed the part that check owner and group (maybe user wants to execute a command to discover he has no permissions)

app/lib/utils/string.js
lordgiotto: I haven't find a module that only checks the octal permissions without doing the fs.stat of a file, but maybe I missed it :)

app/lib/utils/file.js
sindresorhus: Can you write this in the comment and ideally open an issue on `executable` about making it more configurable. Would be nice to be able to just depend on that module instead of copy-paste.

app/lib/utils/string.js
sindresorhus: I think you're commenting on the wrong thing here. This is about shell arguments.

lordgiotto: Oh sorry! I'll look for a module on npm :)

sindresorhus: https://npms.io/ ;)

app/lib/utils/file.js
lordgiotto: Good point, I'll do it :)

app/lib/utils/string.js
lordgiotto: Removed escapeShellArgument and instead added php-escape-shell module from npm :)

app/lib/utils/file.js
lordgiotto: Added detailed comment to explain the situation, fixed the code and opened a pull request to https://github.com/kevva/executable in order to eventually use it directly in the future.

app/lib/actions/ui.js
```diff
@@ -1,6 +1,8 @@
+import { php_escapeshellcmd } from 'php-escape-shell';
```
albinekb: Name this `escapeShellCmd` or at least `phpEscapeShell`

lordgiotto: I don't like it either, believe me, I would prefer that the author had choose a different name. But why set an alias? There is no name conflicts and the original name is self-explicative. Keeping the original name simply helps code clarity imho, that's why i kept it.

albinekb: We don't use snake_case in javascript 😸

lordgiotto: I know, I follow the code style standards too: the author of php-shell-escape doesn't.
Ok, I'll alias it, I just explained why I kept the original name.

albinekb: Sorry, thought it was a copy-pasta error! Didn't check the code of php-shell-escape

lordgiotto:
Rebased on master after reorg; fixed some lint errors; added error handling in openFile inside lib/actions/ui.js.

lordgiotto:
@rauchg @CodeTheory do you still have some concerns about this PR? It's "ready" from 5 days, maybe you wish that I change or explain something?
Thank you :)

CodeTheory:
@lordgiotto Nope! I just didn't see that it was completed from the comments. Sorry!

lib/actions/ui.js
```diff
@@ -166,3 +171,32 @@ export function windowMove () {
...
+      type: UI_OPEN_FILE,
+      effect () {
+        stat(path, (err, stats) => {
+          if (!err) {
+            // We need to use 'php-escape-shell' property this way
+            // until this eslint issue will be fixed:
+            // https://github.com/eslint/eslint/issues/6755
+            let command = shellEscape.php_escapeshellcmd(path);
+            if (stats.isDirectory()) {
+              command = 'cd ' + command + '\n';
```
ekmartin: template string? :)

package.json
```diff
@@ -11,6 +11,7 @@
     "mousetrap": "1.6.0",
     "ms": "0.7.1",
     "object-values": "1.0.0",
+    "php-escape-shell": "^1.0.0",
```
ekmartin: This should probably be pinned to match the rest of the dependencies?

app/index.js
```diff
@@ -20,6 +20,18 @@ app.config = config;
 app.plugins = plugins;
 app.getWindows = () => new Set([...windowSet]); // return a clone
 
+// function to retrive the last focused winow in windowSet;
+// added to app object in order to expose it to plugins.
+app.getLastFocusedWindow = () => {
+  let lastWindow = null;
```
ekmartin: Could do something like:

```js
app.getLastFocusedWindow = windowSet.reduce((lastWindow, window) => (
  window.focusTime > lastWindow.focusTime ? window : lastWindow
);
```

app/index.js
```diff
@@ -20,6 +20,18 @@ app.config = config;
 app.plugins = plugins;
 app.getWindows = () => new Set([...windowSet]); // return a clone
 
+// function to retrive the last focused winow in windowSet;
```
albinekb: typo `focused win*d*ow` 🙈

app/index.js
lordgiotto: Wow, reduce is perfect in this case, never used it, thank you!
However we can't do that because windowSet is a Set, not an array. But i could convert it to an array with `Array.from(windowSet);` :)
Now I'm going to sleep, but tomorrow I'll do that!

app/index.js
lordgiotto: lol, thanks :)

lib/actions/ui.js
lordgiotto: Right, since i still use es5 i'm not accustomed to use it, sorry. I'll change it tomorrow :)

app/index.js
```diff
@@ -205,6 +228,13 @@ app.on('ready', () => {
       }
     });
 
+    // Keep track of focus time of every window, to figure out
+    // which one of the existing window is the last focused.
+    // Works nicely even if a window is closed and removed.
+    win.on('focus', () => {
```
albinekb: Probably better to do it like this:

```
app.on('browser-window-blur', (event, browserWindow) => {
app.getLastFocusedWindow = () => browserWindow
})
````
? 👼

File: app/index.js (hunk 11 above)
Comment (ekmartin) [inline app/index.js]: Ah yeah, didn't think about that 🙃

File: lib/actions/ui.js [hunk 14]
```diff
@@ -166,3 +171,32 @@ export function windowMove () {
...
   };
 }
+
+export function openFile (path) {
+  return (dispatch, getState) => {
+    dispatch({
+      type: UI_OPEN_FILE,
+      effect () {
+        stat(path, (err, stats) => {
+          if (!err) {
```
ekmartin: Maybe this error should be handled somehow too? Not sure how, but a console.log or notification is probably a good start.

lib/utils/file.js
```diff
@@ -0,0 +1,17 @@
+// Based on https://github.com/kevva/executable
+// Since this module doesn't expose the function to check stat mode only,
+// his logic is pasted here.
+// Opened an issue and a pull request about it,
+// to maybe switch to module in the future:
+// Issue: https://github.com/kevva/executable/issues/9
+// PR: https://github.com/kevva/executable/pull/10
+
+export function isExecutable (fileStat) {
+  if (process.platform === 'win32') return true;
```
albinekb: Why does this return true in windows? 😱 
what will happen if you drop in a .txt for example?

app/index.js
lordgiotto: Or `browser-window-focus`, but I've some concerns about this solution: I not sure it would handle closed window as well as uptime solution. What should happen if I've two opened windows, i close one without focusing the other? Or if I launch an hyperterm instance without focusing the window? (not all window manager behave the same throughout different systems).
On the other hand process.uptime check is absolutely rock solid (I can't figure out a situation where it could fail): if a window is closed the previous one will be returned by `app.getLastFocusedWindow` for sure.
And adding a simple integer property to win object is not a big deal :)

albinekb: Didn't think about that, great 👍

lib/utils/file.js
lordgiotto: Now opening a txt on windows would simply return 'Bad command...', but you can open an executable and it would start automatically. Otherwise, if returning false, both opening a txt or an executable would insert the path without adding the "\n".
I don't know, maybe you're right and return always false is a better approach? :)
But, since mode check should not work on win32 platform (as far as i know), i've to return a fixed value.

lib/actions/ui.js
lordgiotto: I handled this error only to be sure that an error is never thrown, but it's really unlikely that you could open a file that doesn't exists, and if you find a way to do so I can't immagine what you should expect :P 
However, how could I send a notification to the user? Should I import utils/notify.js or there is an error event i could emit?

package.json
lordgiotto: Sorry, i don't understand: is pinned dependency a greenkeeper.io thing or you mean set it with the exact version without the `^`? I apologize for the ignorance about this :)

albinekb: dependencies are pinned because there are currently no tests, so a package *could* break hyperterm without us knowing
https://github.com/zeit/hyperterm/pull/513

ekmartin: Yeah, just meant setting it without the `^` :)

albinekb: Sorry i read that wrong 😂 
but yeah, without the `^`

lib/utils/file.js
albinekb: Don't know how the windows implementation will work, if it will use the new windows-bash or something, maybe just wait till we can run hyperterm on windows and take care of that then 👍  I just thought it was weird that it always returned true in windows (which it also does in the [original implementation](https://github.com/kevva/executable/blob/master/index.js#L7))

lib/actions/ui.js
ekmartin: For plugin errors and similar both `console.error` and `notify` are used, like here: https://github.com/zeit/hyperterm/blob/master/lib/utils/plugins.js#L75-L76

lib/utils/file.js
rauchg: This is a bit odd to me as well.

lordgiotto: I agree it's odd, it's a design decision at this point. 

Starting from the idea that when "executable" file is opened, a new line (`\n`) will be appended to the path, we have to define what an executable file is. 

Since windows support is not still defined, I decided to keep the original implementation of "executable" npm module and return a fixed value. Now we alway have possible false positive results, but we can change it to have always false negative results.

In a *nix environment an executable file is a file with a specific mode for a specific user/group/admin; but on windows it's not so clear.
It could be a file with an extension that appears inside [PATHEXT env variable](http://environmentvariables.org/PathExt), but I don't know if the shell we'll use on windows will be capable of execute .EXE, .BAT, etc.

However, maybe we should avoid to use "executable" external module and keep this logic inside our code.

@rauchg @albinekb What do you think about that?

lib/actions/ui.js
lordgiotto: Now error follow the same error reporting convention you suggested me, thanks :)

app/index.js
lordgiotto: Fixed 😆

app/index.js
lordgiotto: Implemented Array.prototype.reduce() method for getLastFocusedWindow, thank you again for the great advice 😄 I had only to handle empty windowSet case to avoid typeErrors.

package.json
lordgiotto: Dep pinned :)

lib/actions/ui.js
lordgiotto: Done!

lib/utils/file.js
lordgiotto: I checked this thing twice on a windows VM and I can confirm that, on win32 platform, the file mode is the same for files with different permissions. So we have to return a fixed value or consider to use the file extension as distinctive.

image

But, 'till win support will be effective, we could support at least *nix system with this feature :)

albinekb: Personally, i think the default in hyperterm should be what people expect, and in both iTerm2 and Terminal.app, nothing special happens if you drop an executable file.

I think that part is more suited for a plugin 🍦

lordgiotto: So you suggest to completely remove the check of executability and keep only two behavious?
- `cd path` and return if it's a folder
- insert the path without return in the other cases?

albinekb: `cd <drop in file/folder location>` and adding a return is a nice feature 🎉 

Yes, thats my $0.05, but I won't stop you 👻

Maybe the magic is an awesome feature a lot of people will like, but if you ask me, I think it should be a plugin 📌 😊

I don't often drop files to the terminal, but when i do it's mostly stuff from chrome like images to use with `dd` 👼

lordgiotto: Consider that it will work not only for drop file onto the dock icon, but also for `open -a '/Applications/Hyperterm.app' 'path/to/a/folder/or/a/file'`. 

So the **folder** feature will be usefull for things like 'atom-terminal' (a plugin for atom that opens the project folder in the terminal) or termHere (an osx application that open the terminal the current finder terminal), since both uses `open -a` under the hood.

Likewise the **executable** may be useful for the same reason, but if you have concerns about that we could reconsider it in the future.

I think that the important thing of this PR is to make file dropping onto the dock icon and `open -a` doing something, because it's a very common and shared behaviour. Enhance this feature could be done on user's feedback basis also in the future :)

albinekb: Sounds good @lordgiotto 🎉 🤗

lordgiotto: Great! Thanks for the feedbacks @albinekb 😄 

So, if @rauchg approves, I remove all this part of executability check and I keep only the base behaviour we talked about:
- `cd ${path}\n` if it's a folder
- `path` without return for any file

rauchg: Seems a bit inconsistent for an arbitrary reason? I would expect both to be printed in the same way.

lordgiotto: I don't understand, sorry. You mean folders should not be `cd`ed?
The behaviour of iTerm.app in this case is to `cd` if it's a folder, or append the return to the path if it's a file (without checking if it's executable or not like I did).

Why it's inconsistent to you?

lordgiotto:
Hi, i'm writing this to explain (IMHO) what this PR is, why it's needed and to discuss with you the best way to implement this behaviour in hyperterm.
After 12 days, 19 commits and 68 messages, it seems to *stale* and it would be a pitty to have lost so much time and energy all of us.
____
**TL;DR;** this PR is not intended only to handle files dropped onto the dock icon, but also handle `open -a Hyperterm.app` files. This actions is widely used on Mac among different application to integrate with the terminal (for instance TermHere for Finder or some atom plugins uses to launch terminal with a specific cwd). We have only to choose how to handle those files.
____
**Why it's important? (IMHO)**
`open-file` event is a specific OSX electron event to handle files opened in two ways:
-  when a file/folder is dropped onto the application icon in the dock
- when the command 'open -a Application' is called
If the first behaviour is useful but not necessary, I think that the second one is very important since a lot of applications use it under the hood to start the terminal with specific conditions.

For instance, in my workflow, it's used by two application: TermHere, a finder extension that let you open the terminal at the folder you're browsing; atom with his plugins to launch the terminal directly at the folder of the project you're editing.

____
**How it should work?**
Before thinking of which should be the best behaviour on hypothetical basis, let's take a look how the official Terminal.app anche iTerm.app works:

|  | Terminal.app | iTerm.app |
|----------------------|-------------------------------|-----------------------------------------------|
| folder | Opens terminal in that folder | Open terminal and cd to that folder |
| file  _(executable)_ | Execute it and exit. | Execute it and exit. `path; exit;` |
| file  _(any other)_ | Do nothing. | "Execute" it and exit, no metter the file type. |

So, for folders it seems straightforward that the expected behaviour could be to launch the terminal with that folder as cwd; talking about files is not so clear. 
The actual PR implementation mimics the behaviour of default Terminal.app, except that non executable files are handled anyway and their path is printed in the command line and not executed. We could also consider to avoid to handle files completely, or execute all of them like iTerm.app does.
My choice would be to keep the actual implementation, but this point is a matter of design and preferences, so @rauchg is the best person shed light on it.
____
**And what about Linux and Windows?**
Since `open-file` is a specific macOS event, some other work will be necessary for linux and windows platforms. But, once this thing is implemented, it will be enough to get the first `process.argv` and pass it to the same ui action we're using to handle macOS opened files (`openFile()` ui action) to achieve the same result.
____
**Conclusion**
This long post is intended only to say that, in my opinion, this is a very important feature and a shared expected behaviour of a terminal emulator, used everyday by a lot o people.
I'm willing to make all the necessary changes, fixes and discussions: please give me feedbacks about it, I really want to help to push forward this project :)

MrRio:
Thanks for the message @lordgiotto – that all makes sense. I'll check out a copy of this branch to test and make sure it's all good to merge.

MrRio:
Hey,

This looks really good. Found a few issues, happy to help fix if needed!

- [ ] Fix executing files with spaces, if I drag "run me.sh" on to the icon in the dock, it fails.
- [ ] Fix opening directories with spaces in too.
- [ ] Dragging files onto the window itself just replaces the browser window with that file, it should have the same behaviour as dragging on to the dock icon (happy for this to be a different Pull Request)

Many thanks,
James

lordgiotto:
Thank for the feedback and the review @MrRio 😄 
Now here is night, but tomorrow i'll work on these issues!

Talking about the third point, i think #171 is pursuing exactly that goal.

MrRio:
Cool! I've opened a PR against your branch, if you merge that then this one is good to go :)

MrRio:
Actually, I can just merge that in myself. 

Looks great, and looking forward to #171

lordgiotto:
Thank you @MrRio 🎉 I'll help with #171 for sure :)
As soon as the new version of eslint will be released and updated here, i'll open a new PR to fix the "style hack" described on line 192 of lib/actions/ui.js :)

index.js
CWSpear: I realize this is way late, but what does `open file` do that doesn't want `termgroup add req` (previously `session add req`) to be fired?

Any other plugin authors that want to rely on `createWindow` to open a "regular window," needs to know they need to call `win.rpc.emit('termgroup add req')`--which isn't intuitive--otherwise their window appears "broken" (it looks like an empty window). I haven't used Hyper in a while, but it appears that this broke a few things because of the new requirement of calling `win.rpc.emit('termgroup add req')` (which has already changed once).

Hopefully there is some way that we can address both use-cases more cleanly.

PR #333, opened 2016-07-21
Update rounded header radius

### LINKED ISSUE #332: Rounded header border-radius is slightly off
You can see it more clearly if you change tab themes (https://github.com/insanityfarm/hyperterm-mactabs) or change the rounded header tab manually. There's a couple pixels of bleed through from the background.

image

This appears to be caused by the fact that `mainRounded` has a `5px` border radius while `headerRounded` is a `6px` border radius. I believe this `1px` addition is an error, and should have actually been a `1px` subtraction to account for the 1px border width of the window.

Adjusted to `4px`:

image

blakeembrey:
Closes #332

### DIFF
--- a/app/lib/components/header.js
+++ b/app/lib/components/header.js
@@ -82,8 +82,8 @@ export default class Header extends Component {
       },
 
       headerRounded: {
-        borderTopLeftRadius: '6px',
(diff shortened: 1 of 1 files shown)


rauchg:
😯😍

PR #341, opened 2016-07-21
Linux Support

c0b41:


package.json
```diff
@@ -18,6 +18,7 @@
     "uid2": "0.0.3"
   },
   "devDependencies": {
+    "electron-installer-debian": "^0.3.0",
```
rauchg: Can you pin please?

index.js
```diff
@@ -78,7 +78,7 @@ app.on('ready', () => {
       if (fn) fn(win);
 
       // auto updates
-      if (!isDev) {
+      if (!isDev && process.platform !== 'linux') {
```
rauchg: Is there a way we can have the auto updater at least *notify* the user?

rauchg:
Thank you so much! Pretty straightforward!

rauchg:
By the way, can `./scripts/package.sh` generate linux binaries as well? It'd be great to incorporate that

index.js
c0b41: idk, but not need linux [autoupdater](http://electron.atom.io/docs/api/auto-updater/#linux)

package.json
c0b41: Fixed

avindra:
This isn't really Linux support, this is debian-based Linux support. Generic linux binaries or an rpm package would be suitable to a wider range of distributions.

E.g., here is Atom's packaged rpm file for their 1.8.0 release:

https://github.com/atom/atom/releases/download/v1.8.0/atom.x86_64.rpm

rauchg:
We already build `deb` and `AppImage`:
https://github.com/zeit/hyperterm/releases/tag/v0.7.1

avindra:
I wasn't aware of the `AppImage` format, works fine. Thanks :+1:

Issue #206, opened 2016-07-17
Closing terminal via Linux window controls leaves instances running


WoodyWoodsta:
When I close by using the window controls in Linux, the window exits but Hyperterm plus the instances associated with electron remain running. Only quitting via the menu option "Quit" correctly terminates the instance.

PR #343, opened 2016-07-21
Fix quit problem

c0b41:
Ref #206


rauchg:
We normally use `{` and `}` but this is sweet! Linux support here we go!

rauchg:
btw those comments were really useful IMHO

c0b41:
@rauchg oh sorry, may fault..

PR #350, opened 2016-07-21
add `TERM_PROGRAM` and `TERM_PROGRAM_VERSION` env vars

TooTallNate:
MacOS's Terminal.app sets these same env variables.
They're useful to be able to sniff out the type of
terminal emulator and do different settings in your
bashrc / vimrc / etc. configuration files.

So for example, `TERM_PROGRAM` will be "HyperTerm",
and `TERM_PROGRAM_VERSION` will be the value from `package.json/version`.

### DIFF
--- a/session.js
+++ b/session.js
@@ -2,7 +2,7 @@ const { app } = require('electron');
 const { EventEmitter } = require('events');
 const { exec } = require('child_process');
 const defaultShell = require('default-shell');
-const { productName, version } = require('../package');
+const { productName, version } = require('./package');
 
 let spawn;
 try {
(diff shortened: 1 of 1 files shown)


rauchg:
I think `TERM_PROGRAM` should be `.toLowerCase`d. Needs rebase. Thanks tootall

TooTallNate:
@rauchg I'm down, but for comparison sake, Terminal.app has mixed case characters:

```
TERM_PROGRAM=Apple_Terminal
```

rauchg:
Great counter point :)

PR #351, opened 2016-07-21
fix lint

TooTallNate:
Otherwise the app won't boot locally 😕

### DIFF
--- a/auto-updater.js
+++ b/auto-updater.js
@@ -1,6 +1,5 @@
 const { autoUpdater } = require('electron');
 const { version } = require('./package');
-const notify = require('./notify');
 const ms = require('ms');
 
 const FEED_URL = 'https://hyperterm-updates.now.sh/update/osx';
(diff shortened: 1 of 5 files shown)


auto-updater.js
```diff
@@ -1,6 +1,5 @@
 const { autoUpdater } = require('electron');
 const { version } = require('./package');
-const notify = require('./notify');
```
Morantron: I think you still need to require this module, since it adds listeners to the `app` object.

rauchg: ugh :\

rauchg: (if that's the case, that is)

freebroccolo: You can add '// eslint-disable-line no-unused-vars' to get eslint to ignore it.

TooTallNate: Updated.

rauchg: I'd love to continue to investigate whether we really need to have this lingering require or not.

mciparelli:
```js
const { version } = require('./package');
require('./notify');
const ms = require('ms');
```
...

auto-updater.js
mciparelli:
```js
const { version } = require('./package');
require('./notify');
const ms = require('ms');
```
...

PR #352, opened 2016-07-21
Add support for multiple cursor shapes

rpunkfu:
Hi!

Just a little PR to add support for multiple cursor shapes used by hterm, let me know what do you think :)

demo

### DIFF
--- a/app/lib/components/term.js
+++ b/app/lib/components/term.js
@@ -47,6 +47,9 @@ export default class Term extends Component {
           props.onResize(cols, rows);
         }
       };
+
+      // this.term.CursorNode_ is available at this point.
+      this.term.setCursorShape(props.cursorShape);
     };
     this.term.decorate(this.refs.term);
     this.term.installKeyboard();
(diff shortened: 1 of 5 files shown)


rauchg:
woah that's so beautiful. Care documenting these in `hyperterm-website` repo?

rauchg:
Also, we need to add it to `config-default` since I suspect some people would love to customize it (specifically, I *love* the looks of `BEAM`)

rpunkfu:
Sure, will do! Beam is my favorite too! :)

rpunkfu:
Which one should be default, block?

rauchg:
`// `BEAM` for | shape, `UNDERLINE` for _, `BLOCK` for █

rpunkfu:
I mean, which one should be set by default in `config-default.js`? :)

rpunkfu:
I added `BLOCK` by default ;)

rauchg:
image

PR #353, opened 2016-07-21
fixing syntax error 'missing parenthesis' on ui.js when running 'npm …

rssilva:
…run dev'


rauchg:
cc @mike-engel

mike-engel:
Yikes! I need to file a bug with whatever atom plugin is doing this to my code. Thanks for the fix @rssilva!

PR #354, opened 2016-07-21
add LANG env variable

### LINKED ISSUE #346: LANG environment variable
```bash
env | grep LANG
```

shows `LANG=en_US.UTF-8` on Terminal and iTerm, in OSX. It shows nothing in HyperTerm.

I'm guessing the correct default `LANG` value depends on the user's language, and isn't completely trivial to detect. Not having it set adds a bit of friction. For example, `mitmproxy` fails to run with the following error message.

```
Error: mitmproxy requires a UTF console environment.
Set your LANG enviroment variable to something like en_US.UTF-8
```

On a separate note, thank you very much for writing and open-sourcing hyperterm! It is insanely cool! ❤️

TooTallNate:
Fixes #346.

``` bash
$ env | grep LANG
LANG=en-US
```

### DIFF
--- a/session.js
+++ b/session.js
@@ -1,3 +1,4 @@
+const { app } = require('electron');
 const { EventEmitter } = require('events');
 const { exec } = require('child_process');
 const defaultShell = require('default-shell');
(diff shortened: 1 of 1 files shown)


pwnall:
Thank you very much for fixing this so quickly for me!!

PR #356, opened 2016-07-21
Added cask update instructions to README.md

arnif:
I setup cask loong time ago and hadn't updated cask since then. Had to update cask before I was able to install hyperterm. Also not the first one to encounter this: https://github.com/zeit/hyperterm/issues/295 :)


rauchg:
good find :)

rauchg:
Also, we've been ditching `$` from command instructions lately, in case you want to help out with another PR

arnif:
Sure thing: https://github.com/zeit/hyperterm/pull/358 :)

PR #366, opened 2016-07-22
Make the cursor opaque by default so that the text beneath it remains legible

jamox:
I've also made a POC implementation as HyperTerm theme: https://github.com/jamox/opaque-cursor but I feel this would make it a sensible default as well.
 
Before: 
Before

After:
screen shot 2016-07-22 at 16 15 29

### DIFF
--- a/app/lib/components/term.js
+++ b/app/lib/components/term.js
@@ -138,6 +138,9 @@ export default class Term extends Component {
       .cursor-node[focus="false"] {
         border-width: 1px !important;
       }
+      .cursor-node[focus="true"] {
+        opacity: 0.5 !important;
+      }
       ${css}
     `]);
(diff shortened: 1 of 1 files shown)


matheuss:
Don't you think that the iTerm2 way is a better approach? 😅
screen shot 2016-07-22 at 10 21 20 am

jamox:
Not sure which I like the best (well my preference matters only a little)

However, with the two themes I've tested, solarized-{light,dark} I've found the opaquity to work better than e.g. `mix-blend-mode`.

Either of these solutions, however, beats the current hiding cursor 😄

hharnisc:
Good news! This was merged with #76

jamox:
Awesome.

PR #369, opened 2016-07-22
Fix OS X keybindings

ekmartin:
For some strange reason it seems like the imported UI action creators were overriding the mapped props with the same names (`moveTo` etc.), [here](https://github.com/zeit/hyperterm/blob/d47e9e43d64ca5300e0df667625ab17a4f703044/app/lib/containers/hyperterm.js#L32). 

This used to work flawlessly previously, so kind of confusing. Would love if someone would try reproducing so it's not just me doing something weird (Node 6.2.0, OS X 10.11.5).

### DIFF
--- a/app/lib/containers/hyperterm.js
+++ b/app/lib/containers/hyperterm.js
@@ -4,7 +4,7 @@ import TermsContainer from './terms';
 import NotificationsContainer from './notifications';
 import Component from '../component';
 import Mousetrap from 'mousetrap';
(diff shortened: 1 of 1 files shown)


mauro-oto:
@ekmartin yes, I was gonna report this. Cmd+1, Cmd+2, etc. stopped working in `master` for me, as well as Cmd+Shift+] and Cmd+Shift+[ 

:+1:

rauchg:
So grateful that you're all testing `master`!

Issue #355, opened 2016-07-21
decorateEnv


rauchg:
Would be nice to have a hook in the same vein as `decorateConfig` to extend the `env` passed to the pty.
See: https://github.com/zeit/hyperterm/pull/354/files#diff-e3160a2ef9b1781b8a44576cd311c948L27

albinekb:
Close?

PR #370, opened 2016-07-22
Add decorateEnv to the extensions API

ekmartin:
See #355. This opens for plugins like this:
```js
exports.decorateEnv = env => (
  Object.assign({}, env, {
    COOL_VARIABLE: 'pretty cool'
  })
);
```

@rauchg: If this is what you had in mind I can send a PR to hyperterm-website as well, updating the extensions API list.


plugins.js
```diff
@@ -16,6 +17,13 @@ const cache = new Config();
 // modules path
 const path = resolve(homedir(), '.hyperterm_plugins');
 const localPath = resolve(homedir(), '.hyperterm_plugins', 'local');
+const availableExtensions = [
+  'onApp', 'onWindow', 'onUnload', 'middleware',
+  'reduceUI', 'reduceSessions', 'decorateMenu',
+  'decorateTerm', 'decorateHyperTerm', 'decorateTab',
+  'decorateNotification', 'decorateNotifications',
+  'decorateTabs', 'decorateConfig', 'decorateEnv'
+];
```
rauchg: This is best suited as a `Set` that we can call `has` on :)

plugins.js
```diff
@@ -282,48 +284,44 @@ exports.onWindow = function (win) {
...
-  });
-  return decorated;
+exports.getDecoratedConfig = function () {
+  const baseConfig = config.getConfig();
+  return decorateObject(baseConfig, 'decorateConfig');
 };
 
+exports.getDecoratedBrowserOptions = function (defaults) {
+  return decorateObject(defaults, 'decorateBrowserOptions');
+};
```
rauchg: great refactor is great

session.js
```diff
@@ -26,12 +25,7 @@ module.exports = class Session extends EventEmitter {
       columns,
       rows,
       cwd,
-      env: Object.assign({}, process.env, {
-        LANG: app.getLocale(),
-        TERM: 'xterm-256color',
-        TERM_PROGRAM: productName,
-        TERM_PROGRAM_VERSION: version
-      })
+      env: getDecoratedEnv()
```
rauchg: what I would do is pass that object here:

ie:

```js
decorateEnv({
  LANG: app.getLocale(),
  TERM: 'xterm-256color',
  TERM_PROGRAM: productName,
  TERM_PROGRAM_VERSION: version
})
```

plugins.js
```diff
@@ -282,48 +284,44 @@ exports.onWindow = function (win) {
...
-      const res = plugin.decorateConfig(decorated);
-      if (res && 'object' === typeof res) {
-        decorated = res;
-      } else {
-        notify('Plugin error!', `"${plugin._name}": invalid return type for \`decorateConfig\``);
-      }
-    }
+exports.getDecoratedEnv = function () {
+  const baseEnv = Object.assign({}, process.env, {
+    LANG: app.getLocale(),
```
sindresorhus: `LANG: app.getLocale()` => `LANG: app.getLocale().replace('-', '_')`

Since `app.getLocale()` returns the locale with `-` (`en-US`) while `echo $LANG` in a shell is with underscore: `en_US`

rauchg: Thanks @sindresorhus. Amazing catch

plugins.js
```diff
@@ -239,14 +247,8 @@ function requirePlugins () {
...
       mod = require(path);
-
-      if (!mod || (!mod.onApp && !mod.onWindow && !mod.onUnload &&
-        !mod.middleware && !mod.reduceUI && !mod.reduceSessions &&
-        !mod.decorateConfig && !mod.decorateMenu &&
-        !mod.decorateTerm && !mod.decorateHyperTerm &&
-        !mod.decorateTab && !mod.decorateNotification &&
-        !mod.decorateNotifications && !mod.decorateTabs &&
-        !mod.decorateConfig)) {
+      const exposed = mod && availableExtensions.some(name => !!mod[name]);
```
sindresorhus: The `!!` is moot as `some` works on truthy/falsy values.

plugins.js
ekmartin: Good feedback, fixed in 160e759ba290318c80236182423b883b0186fa9c 👍

plugins.js
ekmartin: A Set makes sense, but I'm not sure if I understand where you'd call `has` on it? Something like `Object.keys(mod).some(name => availableExtensions.has(name));` or were you thinking of a different place altogether?

plugins.js
ekmartin: Yep, updated 🐈

plugins.js
rauchg: that's good

ekmartin: Updated 🙂

ekmartin:
Rebased :)

PR #377, opened 2016-07-23
fix shell property value

cuth:
When using the new shell option from #192, the `shell` prop is not correctly being set on the session object. This will correctly set the prop.


rauchg:
Great catch 💃

PR #380, opened 2016-07-24
Check if session exists before exiting

timneutkens:
As talked about on the zeit slack channel. When all windows are closed and you press `cmd + w` you get the following error:
screen shot 2016-07-24 at 09 56 49

This extra check to see if the session is still there fixes it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -142,7 +142,7 @@ app.on('ready', () => {
 
     rpc.on('exit', ({ uid }) => {
       const session = sessions.get(uid);
-      session ? session.exit() : false;
+      session && session.exit();
     });
 
     rpc.on('unmaximize', () => {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -141,7 +141,8 @@ app.on('ready', () => {
     });
 
     rpc.on('exit', ({ uid }) => {
-      sessions.get(uid).exit();
+      const session = sessions.get(uid);
+      session ? session.exit() : false;
```
demoneaux: Why not `session && session.exit()` if you want to be less verbose (since you’re not using `if`)?

timneutkens: Fair point. Changing now 👍

rauchg:
Thanks!

PR #381, opened 2016-07-24
Ship yarn with hyper

sheerun:
This is a step towards #282 and makes sure hyperterm always
installs with the same npm version it shipped with.

It also drops dependency on system's node.


rauchg:
That's really sweet. I wonder how much bigger it makes our compressed builds / startup time ?

leo:
I'm not sure why you're ignoring [this](https://github.com/zeit/hyperterm/pull/172#issuecomment-233174361).

sheerun:
It doesn't mater as you embed npm, and have can fix working version if you need.

Spawning a new process simply doesn't work because there's no node installed in electron app..

rauchg:
I would love for @othiym23 to re-visit that decision. It's kind of a bummer that we need to ask the user to install node to use `npm` as an API :(

sindresorhus:
> Spawning a new process simply doesn't work because there's no node installed in electron app..

You can. Just spawn `process.execPath` with the [`ELECTRON_RUN_AS_NODE`](https://github.com/electron/electron/blob/master/docs/api/environment-variables.md#electron_run_as_node) environment variable, and it will use the embedded Node.js.

```js
childProcess.execSync(process.execPath, ['npm', 'install'], {env: {ELECTRON_RUN_AS_NODE: true}});
```

sheerun:
I tried to implement it as @sindresorhus proposed, but for some reason electron refuses to exit when running npm with it. Here's how to reproduce:

```bash
export CWD=$(pwd) && (cd ~/.hyperterm_plugins && ELECTRON_RUN_AS_NODE=1 $CWD/node_modules/electron-prebuilt/dist/Electron.app/Contents/MacOS/Electron $CWD/node_modules/.bin/npm install --verbose)
```

For now it's blocking me, so any help would be appreciated..

rauchg:
@sheerun seems like something we should be reporting to the electron team. This is looking very good btw, thanks @sindresorhus

rauchg:
@sheerun I'd love to merge this. Do you have any news about whether the issue with exiting was fixed? Or any workarounds?

sheerun:
The issue to fix is here: https://github.com/electron/electron/issues/4944

With newest electron it kinda works, but takes 30 seconds to exit.

The best I can do is to rebase, catch.

sheerun:
@leo @rauchg I've replaced npm with yarn and it runs like wonder. It's good to merge.

There's issue with CI build as `xo` doesn't recognize `queue` package as dependency (it is). Maybe that's something @sindresorhus could check..

matheuss:
@sheerun `queue` is on the wrong `package.json` 😅 You need to move it to `app/package.json` 😄

sheerun:
@matheuss Is `yarn` on correct one?

sheerun:
Anyway, I fixed the issue with build. Ready for review and merge..

sheerun:
@matheuss This is no longer in-progress. This is done.

matheuss:
@sheerun thanks for the ping 😄 

> Is yarn on correct one?

Nope, you need to move it to the other one 👌 

Could you please rebase to fix the merge conflicts? I'll make sure to discuss it with the team 👌

sheerun:
@matheuss @rauchg Done. I hope I've rebased it the last time... :)

rauchg:
This is fantastic!

sheerun:
there are conflicts again.... Can I ask you rebase and merge it by yourself?

rauchg:
Yes. Sorry for the trouble

timneutkens:
@sheerun After building this branch using `npm run pack` I'm seeing an issue related to
```
  const electronPath = resolve(
    __dirname, '..', 'node_modules', 'electron',
    readFileSync(resolve(__dirname, '..', 'node_modules', 'electron', 'path.txt')).toString().trim()
  );
```

path.txt can't be found. Can you give me some insights into this?

sheerun:
@timneutkens electron-prebuit should install it in postinstall hook: https://github.com/electron-userland/electron-prebuilt/blob/master/install.js#L39

It uses this file even for executing itself, so this file must be present: https://github.com/electron-userland/electron-prebuilt/blob/master/index.js

sheerun:
any progress?

albinekb:
Yarn has a `--json` option:

kapture 2017-06-10 at 4 52 04

maybe that's a better way to run it?

albinekb:
I think this is the better way to deal with this 👌  Update the deps (yarn 0.17 is old 🗡 ) and then we need to test it.

sheerun:
With all respect, I've already reabased and updated this PR per request
like 5 times. I think this time it's up to you to update it.

chabou:
Rebased and yarn version changed to 0.21.3: Since 0.22.0 it gives me errors:
```
App threw an error during load
TypeError: Cannot read property 'from' of undefined
    at Object.<anonymous> (/Users/chabou/Documents/Projets/hyper/node_modules/readable-stream/node_modules/safe-buffer/index.js:3:11)
    at Object.<anonymous> (/Users/chabou/Documents/Projets/hyper/node_modules/readable-stream/node_modules/safe-buffer/index.js:60:3)
    at Module._compile (/Users/chabou/Documents/Projets/hyper/node_modules/v8-compile-cache/v8-compile-cache.js:178:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (/Users/chabou/Documents/Projets/hyper/node_modules/v8-compile-cache/v8-compile-cache.js:159:20)
    at Object.<anonymous> (/Users/chabou/Documents/Projets/hyper/node_modules/readable-stream/lib/_stream_readable.js:32:14)

```

chabou:
Same error than @timneutkens when using built version.
There isn't neither `path.txt` file nor `node_modules` directory:
`Uncaught Exception:
Error: ENOENT: no such file or directory, open '/Users/chabou/Documents/Projets/hyper/dist/mac/Hyper.app/Contents/Resources/node_modules/electron/path.txt'`

chabou:
I'm trying to debug this problem with `path.txt` file but there is an important thing to note. 
This PR can't work if we have `asar` activated: `node_modules` directory will be packed in  `app.asar` archive 😞

chabou:
This PR do not work by design.
`path.txt` can't be used to this purpose. It almost [constant paths](https://github.com/electron-userland/electron-prebuilt/blob/master/install.js#L51-L63) that are only valid in dev mode. In built mode, we do not have electron dependency, we need to execute Hyper itself (as node_engine), not Electron.

I tried to use `asarUnpack` to extract only `./node_modules/yarn`. But it is a nightmare to extract yarn's dependencies too. We can `asarUnpack` entire `node_modules` directory but it quite the same than disabling Asar.

chabou:
This PR now works in dev and packaged mode.
The only downside is to `asarUnpack` `app/node_modules` directory

We need to test it on Windows and Linux

chabou:
`Yarn` is now embedded as a standalone bin, directly taken from its [release page](https://github.com/yarnpkg/yarn/releases/tag/v0.24.6). It works like a charm even with last version v0.24.6. Only this file is `asarUnpacked` 🎉 
Many thanks to @albinekb for his help.

Still need to be tested on Windows and Linux.

ppot:
@chabou I will let you finish this because it will be cleaner for me in my third part.

chabou:
It is finished! It only needs some testing on Windows/Linux! But I'm struggling with Win10 installation on BootCamp 😞

albinekb:
the app/bin/yarn-standalone.js should be built in the `build` step, it shouldn't be checked into git

chabou:
Why? I consider `yarn` as an external program. If it was written in C++, we wouldn't want to compile it by ourself. What are the pros to not have this external dependency "compiled" and versioned in our git?

chabou:
Problem with production build on Windows solved (bad `yarn.lock` file).
Tested on Mac and Windows (dev and prod build).

Need a test on Linux

chabou:
Tested.

Ready to be merged

app/plugins.js
```diff
@@ -228,46 +223,47 @@ function toDependencies(plugins) {
...
-      execOptions.shell = shell;
-    }
+  function yarn(args, cb) {
+    spawnQueue.push(end => {
+      const fullcmd = [process.execPath, yarnPath].concat(args).join(' ');
+      console.log('Launching yarn:', fullcmd);
+
+      cp.exec(fullcmd, {
+        cwd: path,
+        env,
```
ppot: What happend when `isDev === true`

app/plugins.js
```diff
@@ -228,46 +223,47 @@ function toDependencies(plugins) {
...
+        shell: true,
+        timeout: 1000 * 60 * 5,
```
ppot: add comment for `// 5min`

app/plugins.js
```diff
@@ -228,46 +223,47 @@ function toDependencies(plugins) {
...
-
-    // https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback
-    // node.js requires command line parsing should be compatible with cmd.exe on Windows, should able to accept `/d /s /c`
-    // but most custom shell doesn't. Instead, falls back to default shell
-    if (process.platform !== 'win32') {
-      execOptions.shell = shell;
-    }
+  function yarn(args, cb) {
+    spawnQueue.push(end => {
+      const fullcmd = [process.execPath, yarnPath].concat(args).join(' ');
```
ppot: `fullcmd` ?  Simply use `cmd` or `command`

app/plugins.js
```diff
@@ -228,46 +223,47 @@ function toDependencies(plugins) {
 }
 
 function install(fn) {
-  const {shell: cfgShell, npmRegistry} = exports.getDecoratedConfig();
+  const yarnPath = resolve(__dirname, '..', 'bin', 'yarn-standalone.js');
```
ppot: `yarnPath` is not used elsewhere then inside `yarn` function. Better to move it inside function scope.

app/plugins.js
```diff
@@ -228,46 +223,47 @@ function toDependencies(plugins) {
...
+      console.log('Launching yarn:', fullcmd);
+
+      cp.exec(fullcmd, {
+        cwd: path,
+        env,
+        shell: true,
+        timeout: 1000 * 60 * 5,
+        stdio: ['ignore', 'ignore', 'inherit']
+      }, err => {
+        console.log('Done!');
```
ppot: Be more specific. What is `Done!` ? Plugins installation? Plugins update?

app/plugins.js
```diff
@@ -1,12 +1,13 @@
...
-
-const {app, dialog} = require('electron');
+const cp = require('child_process');
 const {sync: mkdirpSync} = require('mkdirp');
 const Config = require('electron-config');
 const ms = require('ms');
-const shellEnv = require('shell-env');
+const queue = require('queue');
+
+const spawnQueue = queue({concurrency: 1});
```
ppot: `spawnQueue` or `yarnQueue`

app/plugins.js
chabou: This production flag is for `yarn`. See https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-production
No side effect to Hyper.

app/plugins.js
chabou: `spawnQueue` is better.

app/plugins.js
chabou: We don't need this log.

app/plugins.js
chabou: `ms('5m')` will be better. Thank you to have pointed this

app/plugins.js
chabou: I will

app/plugins.js
chabou: I will move it (`env`const too)

chabou:
We should test plugins update too.

chabou:
Plugins update is ok!

app/plugins.js
ppot: ok!

Stanzilla:
1. Close Hyper
2. Edit config file by hand to add a new plugin
3. Launch Hyper

For me it first errors that it can't load the new plugin, then the plugin "update" kicks in, which then installs the new plugin and then everything is fine. I just think that first error should be something we should avoid.

chabou:
@Stanzilla this error is currently shown without my PR.
It could be better handled but it should be another PR.

chabou:
FTR: 
`yarn` v0.24.6 is embedded and [extraResources feature](https://github.com/electron-userland/electron-builder/wiki/Options#config--platformspecificbuildoptions) of `electron-builder` is used to ship it in our bundle.

PR #383, opened 2016-07-24
Allow for specifying global hotkey for toggling visibility

sheerun:
I've implemented it as a default plugin for hyperterm. My presented suggestion is to put default hyperterm plugins into `@hyperterm` folder and namespace modules with `@hyperterm` scope. You might consider claiming this scope on npm.

As for hotkey behavior, it is the same as in iTerm2:

- If any window is hidden, show all
- If all windows are shown, hide all
- Preserve focused window and z-order when showing / hidding them.

The hotkey is configurable with "hotkey" option.

### DIFF
--- a/@hyperterm/hotkey/LICENSE.md
+++ b/@hyperterm/hotkey/LICENSE.md
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2016 Cameron Spear
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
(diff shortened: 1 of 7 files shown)


sheerun:
One annoying thing that will be hard to fix: when hidding iTerm2 with lgobal hotkey, previously focused app (e.g. chrome) gets refocused. This implementation just hides Hyperterm without re-focusing.

This can be fixed later, I think

rauchg:
I'd love this to be shipped separately. @italoacasas is working also on a new plugin listing page where we can highlight our recommendations such as this one, `hypercwd` etc.

image

soutar:
@rauchg @sheerun I published a package to do this today after installing HyperTerm and missing iTerm's hotkey feature. I'd love feedback from you two if you have time.

https://github.com/soutar/hyperterm-summon

sheerun:
@soutar Works wonderfully, thanks. I'd prefer "hotkey" configuration option, but otherwise ok.

rauchg:
@soutar amazing

soutar:
@rauchg @sheerun Cheers for having a look both of you. Regarding hotkey/shortcut config options, I'm not particularly tied to the naming so if you want to PR the change I'll happily merge.

PR #386, opened 2016-07-24
Reorg

rauchg:
In 7 easy(?) steps I moved to `electron-builder`. Benefits:

- Outputs multiple platforms easily
- No more hand-written scripts
- Good directory conventions 
 - Main electron code goes inside `./app`. Only `./app` is included in the build, which is why we target the webpack build there and we moved `index.html` inside it.
 - Browser code is now more accessible (inside `./lib/`)
- Way easier time explaining how to get up and running to newcomers ✨

Thanks @sindresorhus for the suggestion


lordgiotto:
Great :) Unfortunately it revert PR #307.
Please, in package.json, inside build, can you add "extend-info" property that points to Info.plist inside ./static (or inside ./build if you think it's better to move it)?
Otherwise I could do the pull request myself if you prefer :)

Thank you very much!

develar:
I see that you disabled asar. It is not recommended because windows io is very poor. JFYI: I will fix https://github.com/electron-userland/electron-builder/issues/602 soon and open PR to enable asar.

PR #387, opened 2016-07-24
plugins: fix wrong require path

ekmartin:
Seems like the path wasn't updated when stuff was moved around 😀


ekmartin:
Realized that this wouldn't work after the app is packed 😐

PR #388, opened 2016-07-24
Updated README.md to use npm scripts

timneutkens:
Updated README.md to use npm scripts in favor of always running the same version of electron while developing.


rauchg:
Awesome @timneutkens

PR #390, opened 2016-07-25
Update dependencies

freebroccolo:
:arrow_up: electron-builder@5.16.0
:arrow_up: electron-prebuilt@1.3.0
:arrow_up: husky@0.11.6

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,14 +32,14 @@
     "babel-loader": "6.2.4",
     "babel-preset-es2015-native-modules": "6.9.0",
     "babel-preset-react": "6.11.1",
-    "electron-builder": "5.14.2",
-    "electron-prebuilt": "1.2.8",
+    "electron-builder": "5.16.0",
+    "electron-prebuilt": "1.3.0",
     "eslint": "3.1.1",
     "eslint-config-standard": "5.3.5",
     "eslint-plugin-promise": "2.0.0",
     "eslint-plugin-react": "5.2.2",
     "eslint-plugin-standard": "2.0.0",
-    "husky": "0.11.4",
+    "husky": "0.11.6",
     "webpack": "2.1.0-beta.15"
   },
   "eslintConfig": {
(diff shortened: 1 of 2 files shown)


Charlotteis:
@rauchg might I recommend greenkeeper.io to take care of your dependencies automatically? It sends PR's automatically to update your package.json dependencies ☺️

leo:
We'll enable Greenkeeper very soon! Probably this week, I think.

freebroccolo:
@rauchg Looks like this is failing because `app/package.json` is [missing some fields](https://travis-ci.org/zeit/hyperterm/jobs/147263286#L384).

package.json
```diff
@@ -32,14 +32,14 @@
     "babel-loader": "6.2.4",
     "babel-preset-es2015-native-modules": "6.9.0",
     "babel-preset-react": "6.11.1",
-    "electron-builder": "5.14.2",
-    "electron-prebuilt": "1.2.8",
+    "electron-builder": "5.16.0",
+    "electron-prebuilt": "1.3.0",
```
rauchg: when we update this we need to bump the hard-coded one in `app/plugins.js`

rauchg: 😫

freebroccolo: Hmm, I'll have to take a look. Can the one in `app/plugins` be computed from the `package.json` somehow instead?

freebroccolo: I've gone ahead and bumped the version in `app/plugins.js` too.

rauchg: @freebroccolo that'd be amazing but it's difficult because `../package` doesn't get shipped. `electron-builder` only ships `app/`. We'd have to copy `package.json` into `package-dev.json` and place it inside `app/` at build time or something like that. Doesn't sound great. Maybe there's another way of obtaining that version number in runtime

leo:
Greenkeeper is enabled now, btw!

PR #391, opened 2016-07-25
babel-plugin-transform-es2015-modules-commonjs is not in devDependencies

stephenkingsley:
when I run `npm run dev`, `babel-plugin-transform-es2015-modules-commonjs` has not find. so update devDependencies in package.json file.

### DIFF
--- a/lib/containers/hyperterm.js
+++ b/lib/containers/hyperterm.js
@@ -93,7 +93,7 @@ class HyperTerm extends Component {
         <HeaderContainer />
         <TermsContainer ref_={this.onTermsRef} />
       </div>
-      
+
       <NotificationsContainer />
       <style dangerouslySetInnerHTML={{ __html: customCSS }} />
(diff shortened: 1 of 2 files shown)


rauchg:
actually this shouldn't be needed?

rauchg:
The reason is that the bundle we use should already include everything you need. Try running `npm install` again? And perhaps `npm cache clean`?

stephenkingsley:
> The reason is that the bundle we use should already include everything you need. Try running npm install again? And perhaps npm cache clean?

I git clone, and then `npm i`. when I run dev, it warning me can not find `babel-plugin-transform-es2015-modules-commonjs`. maybe is the npm's problem.

albinekb:
```
git status && node --version && npm --version && npm ls
```

stephenkingsley:
@albinekb what is the problem? can you tell me?

PR #393, opened 2016-07-25
🔥 Use uuid

albinekb:
Use [uuid](https://www.npmjs.com/package/uuid) instead of [uid2](https://www.npmjs.com/package/uid2)

`uuid` is a more mature package without callbacks and has no need for error handling!

🎉


rauchg:
Thanks!

PR #395, opened 2016-07-25
Only try to match url if it's needed

albinekb:
- Only try to match an URL when the enter key is sent

Maybe there's some better way to check when a command is being run, but before this the getURL function was called on every keypress


rauchg:
we should probably extensively benchmark to find the best way :)

albinekb:
Indeed!

Issue #67, opened 2016-07-07
Warn upon quit


rauchg:

alireza-ahmadi:
I think the scenario should be something like this:

- If there is 0 or 1 tab, the window will quit immediately.
- If there is at least two tabs, a dialog button(using the Electron's default dialog) will be shown, the dialog would have a message like "All sessions will be closed" and two buttons, "OK" and "Cancel". If user selects "Cancel", nothing will happen and if selects "OK", the application closes.
- We can also define a boolean variable named `quitPropmt` in `.hyperterm.js` config file and allow users to disable functionality.

If this scenario is accepted, I would implement it in the next few days.

d3viant0ne:
If you are going to do this, might i suggest making this a configurable option. There are definitely times where the "Are you sure you want to do this" thing is more aggravating as useful and doing this based on config value shouldn't add a great deal of work. Default it to on if you want but imo there is value in being able to turn it off.

delgermurun:
Please warn also before close tab if there is continuing/unfinished process on this tab.

chabou:
I can't figure out how to know that a process is running without a deeper shell integration

willklein:
@YuaShizuki This is still an open issue. There is a plugin to check out for now: https://github.com/zachflower/hyper-confirm

rob:
So is the plugin mentioned above still the best way to do this, or will Hyper ever support this natively? :confused:

j-f1:
I like how Chrome (or it an extension?) makes me hold down Cmd+Q to quit.

reformedot:
Any update on this?

Stanzilla:
It's up for grabs if you want to make a PR for it? Otherwise we don't have any regular devs on the project so it's depending on external contributions now.

PR #403, opened 2016-07-25
Warn before quitting with sessions open

### LINKED ISSUE #399: Confirm closing
It would be great to have some variation or both (in app or via plugins) of this iTerm features:
- Confirm closing multiple sessions
- Confirm "Quit iTerm2 (⌘+Q) command"

Both of those prevent accidental closing of app (and loosing work if you're not using some multiplexer).

albinekb:
Fixes #399, #67

Im putting this up here to discuss if we should put an option in the menu like chrome  
image

Or if we just should just have a config option for it in `hyperterm.js`

If we want to have it so that the checkbox changes your settings, we need to do like [hpm](https://github.com/matheuss/hpm/blob/master/hyperterm.js) to read and change hyperterm.js

However, with that approach there can be 🐎 race conditions, so i think we should wait with that.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -1,4 +1,4 @@
-const { app, BrowserWindow, shell, Menu } = require('electron');
+const { app, BrowserWindow, shell, Menu, dialog } = require('electron');
 const createRPC = require('./rpc');
 const createMenu = require('./menu');
 const uuid = require('uuid');
@@ -36,6 +36,29 @@ console.log('electron will open', url);
 app.on('window-all-closed', () => {
   if (process.platform !== 'darwin') {
     app.quit();
+  }
+});
+
+let quitIsConfirmed = false;
+
+app.on('before-quit', (event) => {
+  const windows = BrowserWindow.getAllWindows().filter(win => win.sessions);
+
+  if (!quitIsConfirmed && Boolean(windows.length)) {
+    event.preventDefault();
+
+    dialog.showMessageBox({
+      type: 'question',
+      buttons: ['OK', 'Cancel'],
+      defaultId: 0,
+      title: 'Quit HyperTerm?',
+      message: 'All sessions will be closed.'
+    }, (index) => {
+      if (index === 0) {
+        quitIsConfirmed = true;
+        app.quit();
+      }
+    });
   }
 });
 


kjg:
I'd rather have the application just save state and restore upon reopening instead, (and maybe make session restore the toggleable option)

albinekb:
@kjg What if you run commands that is long-running and don't want to accidentally terminate it?

Why not have both? 👼

kjg:
That's a good point. The way mac OS Terminal.app handles this is by showing a dialog warning about the currently running process ONLY when trying to quit or close the tab in which a process is running. I'd definitely like to see HyperTerm do the same. No preference needed for that, and no warning when there are no processes running.

albinekb:
That's really nice, great UX! I'll check and see what i can do 👼

kjg:
I can provide a screen shot or gif of the terminal.app behavior if that will help. Let me know, Thanks!

albinekb:
The problem is how to know if a session is currently running a process, anyone know how to? 👼

rauchg:
One way of doing so would be to ascertain that the current process title is different from the basename of the shell.

If the shell is `/bin/bash` and the current process title is `ps` then you're "running a process".

kjg:
There just has to be a better way than checking the title. If there are no hooks into if a process is running in a session, I think the best course of action is to add one and then take advantage of it.

albinekb:
@kjg yeah we got the pid to the process so i'm currently looking into how to use that

rauchg:
@kjg there also has to be a better way of obtaining the title than this hack I wrote:
https://github.com/zeit/hyperterm/blob/master/app/session.js#L79

kjg:
Ah yeah, there does seem like there should be a better way for getting the title than that haha. I couldn't figure it out quickly though.

albinekb:
I'm looking at https://github.com/Gottox/child_pty to see how hard it would be to make it come straight from there instead.

kjg:
Are https://github.com/zeit/hyperterm/blob/2dec8a0da292a7b2017bc9c5e4767bb04bf69378/lib/actions/sessions.js#L155 or https://github.com/zeit/hyperterm/blob/2dec8a0da292a7b2017bc9c5e4767bb04bf69378/lib/actions/sessions.js#L163 related at all?

app/index.js
```diff
@@ -39,6 +39,29 @@ app.on('window-all-closed', () => {
   }
 });
 
+let quitIsConfirmed = false;
+
+app.on('before-quit', (event) => {
+  const windows = BrowserWindow.getAllWindows().filter(win => win.sessions);
+
+  if (!quitIsConfirmed && Boolean(windows.length)) {
```
TooTallNate: `Boolean` seems clever here. Prefer `windows.length > 0`.

app/index.js
```diff
@@ -39,6 +39,29 @@ app.on('window-all-closed', () => {
   }
 });
 
+let quitIsConfirmed = false;
+
+app.on('before-quit', (event) => {
+  const windows = BrowserWindow.getAllWindows().filter(win => win.sessions);
```
rauchg: I believe we're keeping a `Set` of windows (a copy of which we return from `getWindows`). Maybe we don't need `getAllWindows()`?

albinekb: Smart! I'm currently working on fixing the terminal input issues, and to finish this one i need a way to know if a terminal is currently running something

>The way mac OS Terminal.app handles this is by showing a dialog warning about the currently running process ONLY when trying to quit or close the tab in which a process is running. I'd definitely like to see HyperTerm do the same. No preference needed for that, and no warning when there are no processes running.

will take time to research ways to do that as soon as I'm done with the input fixes :)

albinekb:
Closing this now, I will submit an updated version during #hacktoberfest 🎉

jonaswindey:
any news on this?

albinekb:
Totally forgot this @jonaswindey 😄

Issue #401, opened 2016-07-25
Javascript error popped up while playing with tabs


c4milo:
![screen shot 2016-07-25 at 4 05 04 pm](https://cloud.githubusercontent.com/assets/41681/17115541/caa54534-5281-11e6-973d-8728386d29cf.png)

CodeTheory:
Do you have any more information that could help identify the issue?

Here's some suggestions for things you could provide to help:
- What were you doing for this error to occur specifically?
- What can others to to replicate the issue?
- Did you look into it at all or have any experience on where the error could be coming from?

c4milo:
>What were you doing for this error to occur specifically?

I just opened a bunch of tabs, then close them all using `Cmd + w`, the last once got closed but the hyperterm window remained. The error happens when I try to write anything from that point. Although, it does not happen consistently which leads me to believe it could be some sort of race condition.

>What can others to to replicate the issue?

I just described it in the answer to the first question

>Did you look into it at all or have any experience on where the error could be coming from?

no

rauchg:
Pretty sure this is fixed in `0.7.0`

rauchg:
Actually no, that's a different bug :P

tleunen:
Just had a similar "JS error" when running my tests for an app (not 100% repro tho :/)

Version 0.7.1 (0.7.1.36)
```
Uncaught Exception:
Error: watch /Users/tleunen/.node-spawn-wrap-32116-47b5d6584f88 ENOENT
    at exports._errnoException (util.js:949:11)
    at FSWatcher.start (fs.js:1433:19)
    at Object.fs.watch (fs.js:1460:11)
    at Gaze._watchDir (/Applications/HyperTerm.app/Contents/Resources/app/node_modules/gaze/lib/gaze.js:288:30)
    at /Applications/HyperTerm.app/Contents/Resources/app/node_modules/gaze/lib/gaze.js:357:10
    at iterate (/Applications/HyperTerm.app/Contents/Resources/app/node_modules/gaze/lib/helper.js:69:5)
    at /Applications/HyperTerm.app/Contents/Resources/app/node_modules/gaze/lib/helper.js:78:11
    at /Applications/HyperTerm.app/Contents/Resources/app/node_modules/gaze/lib/gaze.js:422:5
    at iterate (/Applications/HyperTerm.app/Contents/Resources/app/node_modules/gaze/lib/helper.js:69:5)
[...]

PR #404, opened 2016-07-25
Handle lost session on rpc data

albinekb:
Solves #401, check if the session exists like when running `exit`

However, i think there's something that's not done correctly when closing tabs that makes this happen, will investigate more. Possibly race-condition?

If you open a couple of tabs and then close them all quickly it logs `session not found by` from `rpc.on('exit')..`, maybe the close is happening more than once if you do it fast, before it actually got time to close.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -159,7 +159,13 @@ app.on('ready', () => {
     });
 
     rpc.on('data', ({ uid, data }) => {
-      sessions.get(uid).write(data);
+      const session = sessions.get(uid);
+
+      if (session) {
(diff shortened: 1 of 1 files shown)


rauchg:
There's definitely a race condition. I think we need to solve it properly. We should probably use a redux logger to inspect the state over time. The problem is that I was only able to reproduce this "at high speed". I would turn on the redux logger and then the problem would go away. 

tl;DR: https://en.wikipedia.org/wiki/Heisenbug

PR #405, opened 2016-07-26
Update shell after config change and notify the user

mike-engel:
Currently, if you update the shell, it does not take effect until a full reload. This updated the config when it changes, which will update the shell on a new session.

This also sends the user a notification to open a new tab or window to see the shell change.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -6,6 +6,7 @@ const { resolve } = require('path');
 const isDev = require('electron-is-dev');
 const AutoUpdater = require('./auto-updater');
 const toHex = require('convert-css-color-name-to-hex');
+const notify = require('./notify');
 
 // set up config
 const config = require('./config');
@@ -41,7 +42,8 @@ app.on('window-all-closed', () => {
 
 app.on('ready', () => {
   function createWindow (fn) {
-    const cfg = plugins.getDecoratedConfig();
+    let cfg = plugins.getDecoratedConfig();
+
     const [width, height] = cfg.windowSize || [540, 380];
 
     const browserDefaults = {
@@ -70,7 +72,18 @@ app.on('ready', () => {
 
     // config changes
     const cfgUnsubscribe = config.subscribe(() => {
+      const cfg_ = plugins.getDecoratedConfig();
+
       win.webContents.send('config change');
+
+      if (cfg_.shell !== cfg.shell) {
+        notify(
+          'Shell configuration changed!',
+          'Open a new tab or window to start using the new shell'
+        );
+      }
+
+      cfg = cfg_;
     });
 
     rpc.on('init', () => {


app/index.js
```diff
@@ -41,7 +42,8 @@ app.on('window-all-closed', () => {
 
 app.on('ready', () => {
   function createWindow (fn) {
-    const cfg = plugins.getDecoratedConfig();
+    let cfg = plugins.getDecoratedConfig();
```
albinekb: `const cfg`

app/index.js
```diff
@@ -70,7 +72,18 @@ app.on('ready', () => {
 
     // config changes
     const cfgUnsubscribe = config.subscribe(() => {
+      const cfg_ = plugins.getDecoratedConfig();
```
albinekb: rename to something like `oldCfg`

app/index.js
mike-engel: Needs to be `let` since it's gets reassigned when the config changes, on [line 86](https://github.com/zeit/hyperterm/pull/405/files#diff-2018087f584c4398b5c3a23fc0e5f9dbR86)

albinekb: Oh, sorry!

app/index.js
mike-engel: This is the new config, and is also temporary. This, I believe, follows the existing style for temporary variables using `_` at the end.

albinekb: Maybe even do 
`const oldCfg = Object.assign({}, cfg)`

albinekb: Sorry, i'm a bit tired 💃

mike-engel: No worries! Most mornings are like that for me 😄

app/index.js
```diff
@@ -70,7 +72,18 @@ app.on('ready', () => {
...
       win.webContents.send('config change');
+
+      if (cfg_.shell !== cfg.shell) {
+        notify(
+          'Shell configuration changed!',
+          'Open a new tab or window to start using the new shell'
+        );
+      }
+
+      cfg = cfg_;
```
albinekb:
```
cfg = Object.assign({ shell: cfg.shell }, cfg_)
``` ?

That way the url-opener will still work

albinekb:
Nice 👌

app/index.js
mike-engel: I'm tempted to leave this as is, as it will update the config for *any* new session. Right now, config updates only affect the react app, but not the node/electron side. This will update the config that gets used for any new session/window.

Also, this is updating the shell for *new* sessions only, not current ones. Any existing sessions will continue to use the one originally loaded when the session was created.

mike-engel: re: edit2: `shell` is derived from `cfg` during the `initSession` phase

albinekb: Cool!
What do you think of doing this check here https://github.com/zeit/hyperterm/blob/master/app/config.js#L34 instead?

That way it will only be checked once per load instead of once per window?
(possible that I'm completely wrong)

PR #406, opened 2016-07-26
Server now extends EventEmitter instead of creating and EventEmitter instance

jaywunder:


jaywunder:
Wait I apologize, let me update this PR with the most recent version of master

jaywunder:
I saw the Server class was creating an instance of EventEmitter in the constructor.  Using my PR the Server class inherits from EventEmitter.  A benefit of this is `Server instanceof EventEmitter` will be `true`, also inheritance means you don't have to reimplement all the methods you want to have from EventEmitter

Issue #433, opened 2016-07-26
[visual] Tab-bar margin on linux 


albinekb:
As seen here https://github.com/zeit/hyperterm/issues/431#issuecomment-235380436
![image](https://cloud.githubusercontent.com/assets/2480569/17152665/144668b0-5379-11e6-8024-049e525d2859.png)

The tab-bar still has margin for the traffic lights

matheuss:
Related: #5

rauchg:
Yeah, I was thinking that we might want to ship a completely different tab style considering no semaphore lights :D

PR #442, opened 2016-07-26
Fix linux tab margin

c0b41:
#433

### DIFF
--- a/lib/components/tab.js
+++ b/lib/components/tab.js
@@ -85,7 +85,7 @@ export default class Tab extends Component {
       },
 
       first: {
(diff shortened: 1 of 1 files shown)


albinekb:
This will probably need to happen on windows too, http://electron.atom.io/docs/api/browser-window/

Maybe it's better to only set titleBarStyle on macOS and then use that to decide the margin?
https://github.com/zeit/hyperterm/blob/master/app/index.js#L52

rauchg:
It should be `76` if it's explicitly `darwin` as opposed to non-`linux`

PR #445, opened 2016-07-26
Remove duplicated shortcut bindings

marcbachmann:


rauchg:
btw I think this feature is not working for me 😢 
I get `^A`:

image

Issue #269, opened 2016-07-19
Don't close hyperterm when one tab is remaining


oscarmcm:
CMD + W when only have one tab dont close hyperterm.

Demo: http://g.recordit.co/B3aFzVPXyh.gif

![image](https://cloud.githubusercontent.com/assets/2553459/16934317/534b7630-4d12-11e6-9193-92f26fea3861.png)

albinekb:
Hello @oscarmcm, thanks for reporting this issue! 🙌

This is fixed in `master` and will make it in to the next release 🎉

Issue #123, opened 2016-07-15
Shell > Close is ambiguous


Tyriar:
Cannot tell if it's closing the window (all tabs) or a single tab.

ekmartin:
This is a good point, the two close buttons' names are pretty confusing.

ericmarkmartin:
Might it be less confusing if `Close` was renamed to `Close Tab`?

PR #448, opened 2016-07-27
Fix close process

### LINKED ISSUE #206: Closing terminal via Linux window controls leaves instances running

c0b41:
#343 after more, close #206 

maybe some #269 #123


rauchg:
It'll be cool to test that we're not regressing on Mac.

freebroccolo:
It looks like these commits related to "close process" have caused some issue on OS X. See #467.

PR #450, opened 2016-07-27
Fix Close process 

### LINKED ISSUE #206: Closing terminal via Linux window controls leaves instances running

c0b41:
#343 after more, close #206

maybe some #269 #123

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -212,7 +212,7 @@ app.on('ready', () => {
       deleteSessions();
       cfgUnsubscribe();
       pluginsUnsubscribe();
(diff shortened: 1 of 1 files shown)


rauchg:
Maybe a comment is in order above it?

rauchg:
also is `win.close` or `app.quit` what you want?

rauchg:
Thanks btw!

c0b41:
@rauchg yeah app.quit as well but win.close call window-all-closed event, I tried at the same time everything is fine

PR #468, opened 2016-07-27
Add config for bell

### LINKED ISSUE #125: Allow to disable sounds in configuration file

albinekb:
Fix #125

### DIFF
--- a/app/config-default.js
+++ b/app/config-default.js
@@ -57,14 +57,13 @@ module.exports = {
     shell: '',
 
     // for environment variables
-    env: {}
+    env: {},
 
     // set to false for no bell
     bell: 'SOUND'
 
     // URL to custom bell
     // bellSoundURL: 'http://example.com/bell.mp3',
-
 
     // for advanced config flags please refer to https://hyperterm.org/#cfg
   },
--- a/app/config-default.js
+++ b/app/config-default.js
@@ -59,6 +59,13 @@ module.exports = {
     // for environment variables
     env: {}
 
+    // set to false for no bell
+    bell: 'SOUND'
+
+    // URL to custom bell
+    // bellSoundURL: 'http://example.com/bell.mp3',
+
+
     // for advanced config flags please refer to https://hyperterm.org/#cfg
   },
 
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -200,6 +200,12 @@ export default class Term extends Component {
     if (this.props.customCSS !== nextProps.customCSS) {
       this.term.prefs_.set('user-css', this.getStylesheet(nextProps.customCSS));
     }
+
+    if (this.props.bell === 'SOUND') {
+      this.term.prefs_.set('audible-bell-sound', this.props.bellSoundURL);
+    } else {
+      this.term.prefs_.set('audible-bell-sound', '');
+    }
   }
 
   componentWillUnmount () {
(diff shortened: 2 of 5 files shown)


app/config-default.js
```diff
@@ -54,7 +54,13 @@ module.exports = {
 
     // the shell to run when spawning a new session (i.e. /usr/local/bin/fish)
     // if left empty, your system's login shell will be used by default
-    shell: ''
+    shell: '',
+
+    // URL to custom bell
+    // bellSoundURL: 'http://example.com/bell.mp3',
```
rauchg: This comment / flag should be bellow `bell`, not before it 👼

lib/components/term.js
```diff
@@ -200,6 +200,12 @@ export default class Term extends Component {
     if (this.props.customCSS !== nextProps.customCSS) {
       this.term.prefs_.set('user-css', this.getStylesheet(nextProps.customCSS));
     }
+
+    if (this.props.bell === 'SOUND') {
+      this.term.prefs_.set('audible-bell-sound', this.props.bellSoundURL);
+    } else {
+      this.term.prefs_.set('audible-bell-sound', '');
+    }
```
rauchg: You also need to add it to the `constructor`

lib/reducers/ui.js
```diff
@@ -124,6 +127,14 @@ const reducer = (state = initial, action) => {
           ret.termCSS = config.termCSS;
         }
 
+        if (allowedBells.includes(config.bell)) {
```
rauchg: how about `null`. Why is it not a `Set` and `.has` here?

albinekb: When would it be `null`? Yeah I'll change it to a `Set` :)

lib/components/term.js
albinekb: Where?

rauchg: Where we initialize the term and its preferences. `componentWillReceiveProps` only gets invoked when `props` change, not on initial render. So the sound pref wouldn't be honored unless you change the config.

albinekb: Oh, added that https://github.com/zeit/hyperterm/pull/468/files#diff-268710614c2274f611be667a5f72bb13R18

MrRio:
Should this be true/false rather that 'SOUND' or false?

efegurkan:
@MrRio normally you would have, something like visual flash, bell, beep etc. And you would want them to work together so a boolean value would be much more applicable.

But in case there is other options for sound like beep or some other sound, "SOUND" approach would be better.

albinekb:
There will be 'VISUAL' or 'BELL' aswell :)

rauchg:
LGTM

indutny:
Let's land this :)

albinekb:
Rebased, should really have merged master so that the commits/comments would have stayed 👍  will do that next time.

MrRio:
This works really nicely. Tried with external MP3 URL 👍

ericmarkmartin:
I feel like this should be added to the docs on http://hyper.is.

nagman:
it's not working for me. I put bell: false and I still have this f*** sound every second (I use a lot the escape key).

chabou:
@nagman Please follow this issue https://github.com/zeit/hyper/issues/1613

guidobouman:
Hey @albinekb, what about the `VISUAL` option? :D

PR #471, opened 2016-07-27
Use babel for minification

freebroccolo:
This PR does the following:

1. Remove UglifyJS
2. Simplify webpack settings
3. Run `app/dist/bundle.js` through `babel` with minification options
4. Enable harmony flags

The result of (3) is that `bundle.js` is stripped of comments, whitespace, etc.

Size of `bundle.js`:
```
uglify: 528K
babel: 784K
```

The `babel` configuration doesn't do name mangling so it's not as aggressive as uglify but there are a number of transformations that could be applied that might make up some of the difference. See [here](https://github.com/boopathi/babel-minify). Still, minifying with basic `babel` options is not bad.

On my system, init seems to be around `10ms` slower on average. Again, some of the other transforms might make up the difference. It would be nice if someone else could confirm this.

@rauchg I tried using the `harmony` branch for uglify and enabling the babel `es2015` transforms individually but it seems like it only has very basic support for es6 still so it's not worth bothering with. I think it's better to move away from uglify in the long run because it seems unlikely they'll have es6 support any time soon. It sounds like the babel developers are working on a babel-based solution which will probably be the best bet.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -8,6 +8,8 @@ const AutoUpdater = require('./auto-updater');
 const toHex = require('convert-css-color-name-to-hex');
 const notify = require('./notify');
 
+app.commandLine.appendSwitch('js-flags', '--harmony');
+
 // set up config
 const config = require('./config');
 config.init();
--- a/package.json
+++ b/package.json
@@ -33,7 +33,6 @@
     "babel-core": "6.11.4",
     "babel-eslint": "6.1.2",
     "babel-loader": "6.2.4",
-    "babel-preset-es2015-native-modules": "6.9.2",
     "babel-preset-react": "6.11.1",
     "electron-builder": "5.17.0",
     "electron-prebuilt": "1.3.1",
@@ -76,7 +75,6 @@
   },
   "babel": {
     "presets": [
-      "es2015-native-modules",
       "react"
     ]
   },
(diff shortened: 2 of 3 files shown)


rauchg:
@freebroccolo can you calculate a regression in the "scripts parsing" portion of the timeline for each bundle?

rauchg:
btw this simplification looks awesome.

freebroccolo:
@rauchg I'm not sure of a more precise way to measure just the parsing but here's some detail on the timings:

`npm run pack`:

```
babel
-----
14.67s user 2.11s system 110% cpu 15.154 total
14.68s user 2.13s system 111% cpu 15.125 total
14.69s user 2.13s system 110% cpu 15.160 total
14.64s user 2.11s system 110% cpu 15.098 total

uglify
------
15.03s user 1.97s system 111% cpu 15.292 total
15.14s user 1.99s system 111% cpu 15.378 total
15.12s user 1.99s system 111% cpu 15.378 total
15.11s user 2.01s system 111% cpu 15.360 total
```

`npm start` (reported init time in dev console):
```
babel
-----
205.35000000000002
197.12500000000003
201.14999999999998
208.10000000000002
199.86
193.70999999999998

uglify
------
199.27499999999998
193.39500000000004
203.65500000000006
194.66499999999996
191.34500000000006
199.09000000000003
```

babel summary:
babel-summary-opt

uglify summary:
uglify-summary-opt

babel call-tree:
babel-call-tree-opt

uglify call-tree:
uglify-call-tree-opt

rauchg:
I was referring to the "Scripting" piece of the pie chart. That said, it's entirely possible that we make up deficiencies since the transpiled output tends to be slower than native. So I'll merge this.

rauchg:
Thanks you 👏

samccone:
This may be useful
https://github.com/samccone/The-cost-of-transpiling-es2015-in-2016

:)

PR #487, opened 2016-07-28
changed regex for most forms of localhost

### LINKED ISSUE #98: Can't open http://localhost:3000

7373Lacym:
This will allow users to open most forms of localhost using hyperterm.

I am new to regex and used this http://stackoverflow.com/questions/8426171/what-regex-will-match-all-loopback-addresses

Any feedback is much appreciated

Tested on OSX.
Fixes #98


leo:
❤️

PR #488, opened 2016-07-28
Add comment for PR #478

chabou:


leo:
Thanks for clarifying it!

PR #489, opened 2016-07-28
Fix #467 and #206; quit app on window close

### LINKED ISSUE #206: Closing terminal via Linux window controls leaves instances running

### LINKED ISSUE #467: HyperTerm crashes on close on OS X
The commits ed2c810c9ff7c50884a4cc1d3c215d02deede4ad … 72059a7d93f7d18456f8bdf4efcc9ad6c0a084d8 cause HyperTerm to crash with `SIGSEGV` on OS X when HyperTerm is closed. If I rebase on master (2dec8a0da292a7b2017bc9c5e4767bb04bf69378) dropping only these commits the problem doesn't occur.

freebroccolo:
This PR fixes the crash on close on OS X (#467) that was introduced by #450 but still fixes #206.

@rauchg @albinekb @c0b41 It seems the problem was that the `window-all-closed` event was never being triggered. (I looked through the electron issues and didn't see anything about that so I'm not sure if it's just the current behavior or something wrong on our end.)

I've tested this on OS X `10.11.6` and on Ubuntu `16.04`. The only events triggered on window close are `close` and `closed`. Calling `app.quit()` on `closed` seems to work fine.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -36,12 +36,6 @@ const url = 'file://' + resolve(
 
 console.log('electron will open', url);
 
-app.on('window-all-closed', () => {
-  if (process.platform !== 'darwin') {
-    app.quit();
-  }
-});
-
 app.on('ready', () => {
   function createWindow (fn) {
     let cfg = plugins.getDecoratedConfig();
@@ -227,8 +221,11 @@ app.on('ready', () => {
       deleteSessions();
       cfgUnsubscribe();
       pluginsUnsubscribe();
-      if (windowSet.size === 0) {
-        win.close();
+    });
+
+    win.on('closed', () => {
+      if (process.platform !== 'darwin') {
+        app.quit();
       }
     });
   }


c0b41:
@freebroccolo idk but this fix all window closes? maybe close event okey?

freebroccolo:
If someone wants to verify that `window-all-closed` is not being triggered, you can add this to `app/index.js`:

```
// for app
const _appEmit = app.emit.bind(app);
app.emit = (event, ...args) => {
  console.log(event);
  _appEmit(event, ...args);
};

// for win
const _winEmit = win.emit.bind(win);
win.emit = (event, ...args) => {
  console.log(event);
  _winEmit(event, ...args);
};
```

rauchg:
This looks really good to me. Needs rebase.

freebroccolo:
@rauchg done.

PR #492, opened 2016-07-29
Prevent hovering on tabs from causing text flicker

### LINKED ISSUE #436: Hover tabs makes font flicker
font

Possibly something to do with font-smoothing or GPU rendering (`transform: translateZ(0);`) ?,
Also makes the prompt text flicker but not as much.

Only present when i have more than one tab open

mike-engel:
For some reason, geometricPrecision was causing text to flicker on non-retina screens when hovering over the tabs. Fixes #436

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -15,6 +15,12 @@
         text-rendering: optimizeLegibility;
         box-sizing: border-box;
       }
+
+      @media (min-resolution: 2dppx) {
+        * {
+          text-rendering: geometricPrecision;
+        }
+      }
     </style>
   </head>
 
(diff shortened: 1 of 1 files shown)


leo:
The only problem with this fix is that it makes the text thicker on retina devices (before on top and after at the bottom):

screen shot 2016-07-29 at 09 21 33

mike-engel:
Huh, I had tested it on my retina device and didn't see that. I'll add a media query then!

albinekb:
This one will close #436

mike-engel:
@rauchg @leo updated the inline stylesheet to use a pixel ratio media query for retina displays

Issue #505, opened 2016-07-29
Close button icon asset is missing


insanityfarm:
Just noticed that the tab close button "X" icon no longer exists. The developer tools console contains this error:

`file:///Applications/HyperTerm.app/Contents/Resources/app/assets/icons.svg#close Failed to load resource: net::ERR_FILE_NOT_FOUND`

I opened the app package contents in Finder and confirmed that I can't find icons.svg anywhere. There is no assets directory inside Hyperterm.app/Contents/Resources/app/.

leo:
Fixed in https://github.com/zeit/hyperterm/commit/2a13f3a38f686131d479947e7a36e64ef1f2bd76

insanityfarm:
Sorry to reopen, but did I miss something here? I could've sworn this was fixed, but I have the latest version and it's missing the asset again. Same issue as I initially reported.

ekmartin:
It hasn't been released yet @insanityfarm, should be fixed in the next version. If you're eager to try it out you can create your own `HyperTerm.app` by cloning and running `npm install && npm run pack`. More info here: https://github.com/zeit/hyperterm#contribute

insanityfarm:
Ah, thank you. I had misread Github and thought it was already released.

PR #508, opened 2016-07-29
Correct css blob type

albinekb:
- Correctly copy assets (fix for https://github.com/zeit/hyperterm/commit/2a13f3a38f686131d479947e7a36e64ef1f2bd76)
- Use correct way to set blob type
#505 
🐱

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -139,8 +139,8 @@ export default class Term extends Component {
         border-width: 1px !important;
       }
       ${css}
-    `]);
-    return URL.createObjectURL(blob, { type: 'text/css' });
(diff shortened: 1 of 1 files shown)


leo:
For me, https://github.com/zeit/hyperterm/commit/2a13f3a38f686131d479947e7a36e64ef1f2bd76 fixed the problem with the close icon. The whole "assets" directory gets copied to "app/dist". So I'm not sure why you've added https://github.com/zeit/hyperterm/pull/508/commits/24f5c4b6316031ce3158cb07899ed2285ebfcb79.

albinekb:
Oh yeah it worked.. 📦

PR #509, opened 2016-07-29
GitHub templates for issues and pull-requests

iamstarkov:
some sane templates (borrowed from [ava](https://github.com/avajs/ava/tree/master/.github) and adjusted). Seems fine to start.

Also maybe we need consider adding more detailed contributing guidelines plus maybe code of conduct, but its up to discussion

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,50 @@
+<!-- CLICK "Preview" FOR INSTRUCTIONS IN A MORE READABLE FORMAT -->
+
+## Prerequisites
+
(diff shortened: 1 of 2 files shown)


rauchg:
Awesome! Needs rebase tho.

rauchg:
Also, I'd like to make it a lot briefer and focus more on bugs that are likely to occur with HyperTerm
- The whole template should ideally fit in the height of the issue `<textarea>`
- We should continue to recommend #hyperterm slack for realtime support
- They're not likely to have tests, we can get rid of that
- The bonus points should actually be to include a gif, for which we can link to licecap if the user doesn't know how to record one.

iamstarkov:
sorry, do not know to phrase it that short

PR #513, opened 2016-07-29
Use semver for devDependencies

albinekb:
This will make @greenkeeperio-bot send less PR's

https://blog.greenkeeper.io/announcing-real-time-dependency-break-detection-for-greenkeeper-4f7558c10d77#.rae5cx64w

because there are no tests, i only updated devDependencies, 
however we should maybe add `npm run dist` to the tests? to make sure nothing broke?

Good/bad idea?


rauchg:
This is not a bad compromise. If we see unexpected errors come up in the future due to this we can revert it.

PR #515, opened 2016-07-30
make eslint work again :P

rauchg:


rauchg:
Merging this directly as it's making `git push` not work :\

PR #517, opened 2016-07-30
Fallback to nano if no editor is set

MrRio:
If $EDITOR is not set, changing the config file fails. We should default to nano. 

screen shot 2016-07-30 at 14 35 30

Reasons to use nano:

* It's eveywhere.
* It's less confusing for beginners, the instructions for saving the file are at the bottom.
* Vim fans will have already set their $EDITOR config, or this will convince them to.
* crontab on a fresh Ubuntu install does the same thing, presumably for the reasons listed above.

I'm personally a vim fan, but this just fixes an issue new users a likely to run into.

### DIFF
--- a/lib/actions/ui.js
+++ b/lib/actions/ui.js
@@ -137,6 +137,11 @@ export function moveTo (i) {
 }
 
 export function showPreferences () {
+
+  var editorFallback = 'nano';
+  if (process.platform === 'win32') {
+    editorFallback = 'notepad';
+  }
   return (dispatch, getState) => {
     dispatch({
       type: UI_SHOW_PREFERENCES,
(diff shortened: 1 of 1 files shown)


CodeTheory:
Does Windows have `nano` installed by default? What happens with them?
With active efforts to support Windows, we shouldn't overlook it.

MrRio:
I'll boot up a VM now

MrRio:
The current command only works if bash is installed.

Does the current version of hyperterm run on Windows?

MrRio:
screen shot 2016-07-30 at 15 24 06

I can't seem to get it to work. 

Notepad is probably a safe bet though, so once Bash is working and hooked up to the terminal for Windows this patch will work.

screen shot 2016-07-30 at 15 26 13

CodeTheory:
A fallback like you've implemented is perfect in my opinion. I'm not sure of the current progress for windows, but I know there are active interests and efforts to get it to work there so catering to that interest is a good option in my opinion.

lib/actions/ui.js
```diff
@@ -137,6 +137,10 @@ export function moveTo (i) {
 }
 
 export function showPreferences () {
+  var editorFallback = 'nano';
```
ekmartin: `const editorFallback = process.platform === 'win32' ? 'notepad' : 'nano';`
or use `let` instead of `var` if you'd rather have the if statement.

rauchg:
I'll merge this for now but it's likely that we'll resort to `node-open` if `EDITOR` is not set.

PR #519, opened 2016-07-30
Add support for opening the terminal on the correct display

MrRio:
See this issue https://github.com/zeit/hyperterm/issues/481

Opening a new window will now open it ever so slightly diagonally across from the current active window. This is the same behaviour as Terminal.app. I've got for the height of the header as the value.

screen shot 2016-07-30 at 17 46 20

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -44,7 +44,7 @@ app.on('ready', () => {
 
     let startX = 50;
     let startY = 50;
-    
+
     // Open the new window roughly the height of the header away from the
     // previous window. This also ensures in multi monitor setups that the
     // new terminal is on the correct screen.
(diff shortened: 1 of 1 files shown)


rauchg:
This is brilliant 👏. What happens if you go off-screen? Should we reset the `y` back to the starting point and shift the `x`?

rauchg:
@MrRio I apologize for the poor braille character width calculation so far :P

MrRio:
@rauchg hah, it's hardly the feature at the top of every terminal emulator's list. I'm pretty sure we can fix it by kicking the renderer into the same mode that creates <span>s for colored chars and some CSS.

Yeah the X and Y should reset, I'll knock that up

MrRio:
This works well for me, let me know :)

MrRio:
Build broken due to https://github.com/zeit/hyperterm/pull/533

PR #522, opened 2016-07-30
Revert "Use babel for minification (#471)"

rauchg:
Unfortunately on my Macbook 12 the difference between parsing a 1.5mb JS bundle and a 400kb one is very significant (in the order of 60-100ms consistently added to bootup time).

While I think less transpilation is always good, this is an example of when it's not. Since Uglify doesn't work with modern ES6 features, giving it up is very costly in parsing and interpretation time.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -8,8 +8,6 @@ const AutoUpdater = require('./auto-updater');
 const toHex = require('convert-css-color-name-to-hex');
 const notify = require('./notify');
 
-app.commandLine.appendSwitch('js-flags', '--harmony');
-
 // set up config
 const config = require('./config');
 config.init();
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "babel-core": "^6.11.4",
     "babel-eslint": "^6.1.2",
     "babel-loader": "^6.2.4",
+    "babel-preset-es2015-native-modules": "^6.9.2",
     "babel-preset-react": "^6.11.1",
     "copy-webpack-plugin": "^3.0.1",
     "electron-builder": "^5.19.1",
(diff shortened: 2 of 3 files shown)


rauchg:
cc @freebroccolo

freebroccolo:
@rauchg Can you double check the `app/dist/bundle.js` size after `npm run pack`? The size reported by webpack is before `babel` is run so it's not minified yet. The minified size with `babel` should end up around `780K`. That's still bigger than the ~`530K` from uglify though.

Also, are you seeing that much of a difference with the default hyperterm config? One thing I noticed is that a few of the more popular extensions are adding quite a bit to start up time. I've seen > `20ms` for some of them so I try to test with the default config just to be sure.

With the default config on a 2012 macbook retina I'm still only seeing around `10ms` difference.

One thing you could also try is adding these plugins to the babel configuration:

```
    "plugins": [
      "transform-member-expression-literals",
      "transform-merge-sibling-variables",
      "transform-minify-booleans",
      "transform-property-literals",
      "transform-remove-debugger",
      "transform-simplify-comparison-operators",
      "transform-undefined-to-void"
    ]
```

I think that saved another `60K` or so last time I tried. I think it may have been faster too but it started to get to the point where it was hard to see a difference on average between the `uglify` and `babel` version without a more precise testing set up.

If you still think it's worth switching back to uglify that's okay though. However, I think I've seen @hzoo say somewhere that the babel team is close to releasing their own minifier and I would expect that being based on babel it would be a much better long term solution.

rauchg:
I see. That's a very good point!

rauchg:
(I was running `NODE_ENV=production webpack` so I assumed optimizations were being enabled)

PR #524, opened 2016-07-30
Add window bgcolor updates

rauchg:
- Electron's window gets a backgroundColor now with the alpha value in front, as per the documentation (https://github.com/electron/electron/blob/master/docs/api/browser-window.md#new-browserwindowoptions)
- Simplify `Header` and `Term` to no longer require a background color at all. Everything is transparent and relies on the main setting of the window
- Improve borders in the tabs to not rely on setting a border of a different color for the active tab. Also restore the margin on mac that some other PR broke

This PR is more of a "future proofing". The main reason is that alpha values seem broken in Electron, and `setBackgroundColor` is broken (doesn't do anything for me). So we still have to set a background in the body when the configuration changes. This can have the negative effect of adding a secondary layer of alpha (transparency on top of transparency).

The other reason is that there seem to be a lot of rendering quirks in Chromium with transparent backgrounds.


rauchg:
Glitches notwithstanding, it looks pretty cool:

screen shot 2016-07-30 at 1 36 47 pm

app/utils/to-electron-background-color.js
```diff
@@ -0,0 +1,14 @@
+const Color = require('color');
+
+// returns a background color that's in hex
+// format including the alpha channel (e.g.: `#00000050`)
+// input can be any css value (rgb, hsl, string…)
+module.exports = function toElectronBackgroundColor (bgColor) {
+  const color = Color(bgColor);
+  if (1 !== color.alpha()) {
+    // (╯°□°）╯︵ ┻━┻
```
MrRio: haha :)

rauchg:
I'm gonna merge since I think it's pretty sweet. Gonna break a few themes but it's the right way to go

PR #533, opened 2016-07-31
Fix build -- Don't force install of next version of electron builder

MrRio:
We'll just use the version specified in packages.json


MrRio:
This issue with Travis has resolved itself without this PR.

Issue #424, opened 2016-07-26
Add `env` config


rauchg:
It'd be cool if we let the user merge custom ENV easily from the config.

Code should look like ~:

```js
Object.assign({}, process.env, {
      LANG: app.getLocale().replace('-', '_') + '.UTF-8',
      TERM: 'xterm-256color',
      TERM_PROGRAM: productName,
      TERM_PROGRAM_VERSION: version
    }, envFromConfig || {})
```

timneutkens:
@CodeTheory seems like this can be closed 👍

CodeTheory:
Yup, looks like it!

PR #534, opened 2016-07-31
Add support for environment variables in config

zhaorz:
#424 Allows users to specify environment variables in their `.hyperterm.js` under a field named `env`.


MrRio:
Yeah this looks good, tested locally. Can you update with the latest commits from master?

zhaorz:
@MrRio Merged the commits over from master, should be good to go

PR #547, opened 2016-08-02
[WIP] Add proper window support

MrRio:
Adds window switching and title support to the main menu, and the dock menu.

zukzs1tpoc

screen shot 2016-08-02 at 02 51 56

Adds:
* Switching the window title when the tabs focus changes
* Add the window titles to the Window menu
* Add the titles to the dock menu
* Allow making a new window from the dock

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -97,7 +97,6 @@ app.on('ready', () => {
     const browserOptions = plugins.getDecoratedBrowserOptions(browserDefaults);
 
     const win = new BrowserWindow(browserOptions);
-
     windowSet.add(win);
 
     win.loadURL(url);
@@ -159,6 +158,7 @@ app.on('ready', () => {
         });
 
         session.on('title', (title) => {
+          win.setTitle(title);
           rpc.emit('session title', { uid, title });
         });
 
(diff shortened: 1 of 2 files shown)


app/menu.js
```diff
@@ -209,6 +209,7 @@ module.exports = function createMenu ({ createWindow, updatePlugins }) {
     },
     {
       label: 'Window',
```
sindresorhus: `label` is not needed when specifying `role`.

app/index.js
```diff
@@ -309,6 +310,16 @@ app.on('ready', () => {
       }
     }));
 
+    // If we're on Mac make a Dock Menu
+    if (process.platform === 'darwin') {
+      const {app, Menu} = require('electron');
+
```
rauchg: extra `\n`

app/index.js
```diff
@@ -309,6 +310,16 @@ app.on('ready', () => {
       }
     }));
 
+    // If we're on Mac make a Dock Menu
+    if (process.platform === 'darwin') {
+      const {app, Menu} = require('electron');
```
rauchg: `standard` needs spaces when de-structuring IIRC. Did the `git push` linting not complain?

app/menu.js
```diff
@@ -208,7 +208,7 @@ module.exports = function createMenu ({ createWindow, updatePlugins }) {
       ]
     },
     {
-      label: 'Window',
+      role: 'window',
```
rauchg: nice!

rauchg:
other than those nits, LGTM!

app/index.js
MrRio: The linter didn't complain no, I think the rule it enforces it that you keep it consistent throughout that part.

PR #412, opened 2016-07-26
Add syntactic support for tags 🏷

matheuss:
There's already technical support since you can `hyperpower#beta` in your `~/.hyperterm.js`, but syntactic support with proper documentation is better ☺️

PR #549, opened 2016-08-02
Add the very first test 🎉

matheuss:
Now we have tests running on macOS and Linux! 🎉 🙏

This PR also implements what #412 proposed – but in a better way.

I'll only merge this with the 👍 of @rauchg and @leo since it's a PR that introduces how the testing will be done 😅

### DIFF
--- a/test/app/plugins.js
+++ b/test/app/plugins.js
@@ -9,13 +9,14 @@ describe('plugins', function () {
                        '@org1/project4#1.0.0', '@org2/project5@alpha',
                        '@org3/project6'];
 
-      const result = _toDependencies({plugins: plugins});
-      result.should.have.property('project1', 'latest');
-      result.should.have.property('project2', '1.0.0');
-      result.should.have.property('project3', 'beta');
-      result.should.have.property('@org1/project4', '1.0.0');
-      result.should.have.property('@org2/project5', 'alpha');
-      result.should.have.property('@org3/project6', 'latest');
+      _toDependencies({plugins: plugins}).should.be.eql({
+        'project1': 'latest',
+        'project2': '1.0.0',
+        'project3': 'beta',
+        '@org1/project4': '1.0.0',
+        '@org2/project5': 'alpha',
+        '@org3/project6': 'latest'
+      });
     });
   });
 });
--- a/test/app/plugins.js
+++ b/test/app/plugins.js
@@ -1,0 +1,21 @@
+require('../setup');
+
+const {_toDependencies} = require('../../app/plugins');
+
+describe('plugins', function () {
+  describe('#toDependencies()', function () {
+    it('should convert dependencies form hyperterm\'s format to npm\'s', function () {
+      const plugins = ['project1', 'project2#1.0.0', 'project3@beta',
+                       '@org1/project4#1.0.0', '@org2/project5@alpha',
+                       '@org3/project6'];
+
+      const result = _toDependencies({plugins: plugins});
+      result.should.have.property('project1', 'latest');
+      result.should.have.property('project2', '1.0.0');
+      result.should.have.property('project3', 'beta');
+      result.should.have.property('@org1/project4', '1.0.0');
+      result.should.have.property('@org2/project5', 'alpha');
+      result.should.have.property('@org3/project6', 'latest');
+    });
+  });
+});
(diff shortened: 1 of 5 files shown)


leo:
Thanks! 😊 Is it already possible to use ava instead of mocha?

matheuss:
Very unfortunately, no 😢 After some extensive talk with @sindresorhus and @rauchg, we concluded that AVA can't do the job here – for now. So, LGTM with Mocha for now :metal:

test/app/plugins.js
```diff
@@ -0,0 +1,21 @@
...
+                       '@org1/project4#1.0.0', '@org2/project5@alpha',
+                       '@org3/project6'];
+
+      const result = _toDependencies({plugins: plugins});
+      result.should.have.property('project1', 'latest');
+      result.should.have.property('project2', '1.0.0');
+      result.should.have.property('project3', 'beta');
+      result.should.have.property('@org1/project4', '1.0.0');
+      result.should.have.property('@org2/project5', 'alpha');
+      result.should.have.property('@org3/project6', 'latest');
```
gustavnikolaj: I know that this is kind of an annoying drive-by-suggestion, so sorry in advance; this is just such a good example of why I like the unexpected assertion framework.

The above assertions could be folded into a single one:

```js
expect(_toDependencies({ plugins: plugins }), 'to satisfy', {
  'project1': 'latest',
  'project2': '1.0.0',
  'project3': 'beta',
  '@org1/project4': '1.0.0',
  '@org2/project5': 'alpha',
  '@org3/project6': 'latest'
});
```

The `to satisfy` assertion is like equals - except that it doesn't complain about extra keys. So the following won't fail:

```js
expect({ foo: 'bar', qux: 'baz' }, 'to satisfy', { foo: 'bar' });
```

Rolling multiple assertions, like in this example with should, into a single one, means that the error can give you a lot more context and generally a better idea about what is wrong. The focus of the unexpected assertion library is to provide good enough tools that you can fold these assertions together and to give the best possible output when the assertions fail.

```js
> expect({ foo: 'bar', baz: 'qux' }, 'to satisfy', { foo: 'qux' });

expected { foo: 'bar', baz: 'qux' } to satisfy { foo: 'qux' }

{
  foo: 'bar', // should equal 'qux'
              //
              // bar
              // qux
  baz: 'qux'
}
```

http://unexpected.js.org/

(I work/worked with the authors)

matheuss: Since I have no expertise with Mocha, I just used the assertion framework that the docs suggested 😅 
The `to satisfy` thing seems really cool – I'll definitely take a look at Unexpected 🙂 Thanks!

gustavnikolaj: `to satisfy` has a whole bag of tricks beyond this, but even in it's simplest form it's really useful :) The documentation site has a lot of examples and [a list of all the assertions](http://unexpected.js.org/assertions/any/to-be/) in core, plus [a list of plugins](http://unexpected.js.org/plugins/) (most of which also has documentation sites) that might come in useful when doing more specialized things.

I'm on the [unexpected gitter channel](https://gitter.im/unexpectedjs/unexpected) and I'd love to help you out if you have any questions :-)

rauchg:
LGTM! Would love to move to `ava` when it becomes an option. I like the idea of having a single test lib for everything

matheuss:
Cool ☺️ I'll rebase/merge later today and then start writing tests for everything ☺️

fearphage:
> After some extensive talk with @sindresorhus and @rauchg, we concluded that AVA can't do the job here – for now.

Out of curiosity, what is it that ava cannot do?

matheuss:
To test a file that `require('electron')`, you need to run it **inside** electron, which means you have to `electron test.js`. But AVA [does not let you do this](https://github.com/avajs/ava/blob/204f2be2e8f174f698a56d504a0a6c76a1e0b60f/lib/process-adapter.js#L17) :/

I was supposed to open an issue in AVA about this, but I forgot 😹 Thanks for helping me remember  😜

sindresorhus:
AVA forks a new Node.js process for test files. When you run it using Electron you want AVA to fork the test file process using Electron too, but Electron changes the behavior of `childProcess.fork()` and instead force spawns Electron in Node.js mode, where `require('electron')` is not available. I have opened an issue on Electron, but they don't seem very interested in adding any kind of workaround for it: https://github.com/electron/electron/issues/6656

PR #559, opened 2016-08-04
File dropped into terminal window adds path to command line

lordgiotto:
Since #171 seems to be abandoned, I reimplement his idea, but changing how data is added to session (using actions instead writing directly to term component).

Changes I made:
- 177265f: Added `file-uri-to-path` depencency
- 060ffc2: added rpc event `session data send`, bound to `sendSessionData()` action. `session data` is not suitable to send data in this case 'cause it doesn't add data to session, but only to state.
- e2f13af: if `sendSessionData()` is called without uid (which means that `rpc.emit('session data send')` is emitted without uid), the activeUid is used instead: data will be sent to the active session.
- c212ef2: webContents `will-navigate` event is prevented if a file is dropped onto the terminal window, and his path is sent to the active session (emitting `session data send` rpc event like described in the commit 060ffc2)

Let me know what do you think about these changes, thanks :)

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -3,6 +3,8 @@ const createRPC = require('./rpc');
 const createMenu = require('./menu');
 const uuid = require('uuid');
 const { resolve } = require('path');
+const { parse: parseUrl } = require('url');
+const fileUriToPath = require('file-uri-to-path');
 const isDev = require('electron-is-dev');
 const AutoUpdater = require('./auto-updater');
 const toHex = require('convert-css-color-name-to-hex');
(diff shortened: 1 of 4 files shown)


MrRio:
This is great, works a treat! 👍 Even with spaces in the names

pablopunk:
This doesn't work anymore 🤔

Vico1993:
Hey I opened a new pull request who fix this issue on the last version of Hyper ( tested on MacOs and Windows 10 ) 
#3690

PR #560, opened 2016-08-04
Fix session exiting

chabou:
Here a bug I found:
- Open a new tab
- launch `vi`
- close tab with `vi` running
- Exception thrown : sessions.js?3551:40 Uncaught TypeError: Cannot match against 'undefined' or 'null'.

Exception occurs here : 
`````javascript
export function addSessionData (uid, data) {
  return function (dispatch, getState) {
    dispatch({
      type: SESSION_ADD_DATA,
      data,
      effect () {
        const { shell } = getState().sessions.sessions[uid]; // <------- session doesn't exist anymore

        const enterKey = Boolean(data.match(/\n/));
        const url = enterKey ? getURL(shell, data) : null;
````

Why are trying to add some session data on a deleted session ? This is because `vi` send some lines on exit.

I found that session in Electron was not properly exited/cleaned.
Only session.exit() was called (this function call only this.destroy())
Listeners was not removed and session was not deleted from `sessions` collection.

I found another place where session was not deleted from `sessions` collection.

Maybe I missed or misunderstood something.


chabou:
Ok I misunderstood some flow (deleting from sessions collection on exit session event).
I'll correct this later.

chabou:
Corrections done.
I think it is better to stop data propagation from pty at source, using `ended` flag in Session object.

rauchg:
wow good find :)

leo:
@chabou Please rebase! 😊

leo:
❤️

Issue #527, opened 2016-07-30
Fails to display anything


austinwade:
Hello all,

I am running Hyperterm 0.7.1.36 on OSX El Capitan 10.11.6.

When I start Hyperterm, the app opens but does not display anything within the console:

![](https://i.gyazo.com/610c27b134ad7ba541414bb1368559ad.png)

This problem has occurred since the first time I opened the app after upgrading to version 0.7.

I tried clearing the Hyperterm directory within _~/Library/Application Support/_ along with a restart, unfortunately to no luck.

Could someone help me? Thank you so much!

albinekb:
Hello @austinwade, are you able to join the HyperTerm slack? I think it would be easier to debug that way 🚑 
[![Slack Channel](https://zeit-slackin.now.sh/badge.svg)](https://zeit.chat/)

If not, can you try pressing `alt+cmd+i` or go to `View` -> `Toggle Developer Tools` in the menubar?

 👋

albinekb:
Chatted with @austinwade and found out what the problem was: https://github.com/zeit/hyperterm/blob/master/lib/components/term.js#L31

If you set `cursorColor` to anything that [color](https://www.npmjs.com/package/color) can't parse, the error is not handled and results in a broken hyperterm 😭

jwhitmarsh:
I'd love to start contributing to this project, and this seems like a good place to start - there isn't much in the README in regards to contributing, so can I just crack on is there anyone who could point out some gotchas for me?

PR #562, opened 2016-08-04
Validate cursorColor value and apply default if it fails (#527)

jwhitmarsh:
Added `validateColor()` method that takes a color to validate and an alternative if validation fails (which has a default value of `rgb(255,255,255)`). If validation does fail a `notification` is raised. 

There is some risk that a developer could pass an invalid color as the alternative. Initially I was testing for that, but it causes the notification to be raised for the alternative, not the user-defined color, which would be confusing. For the sake of simplicity I've removed that second layer of validation, but can add it back in if people think it should be there. 

The PR is for #527 so I've only applied the validation to setting cursorColor, but I don't see why it shouldn't be used in all the configurable color options? Is there any reason [color](https://www.npmjs.com/package/color) is only being used for setting cursorColor?

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,7 +1,12 @@
+# build output
+dist
+build
 app/dist
+
+# dependencies
+node_modules
 app/node_modules
+
+# other
 app/static
 assets
-build
-dist
-node_modules


lib/components/term.js
```diff
@@ -143,6 +144,17 @@ export default class Term extends Component {
     return URL.createObjectURL(blob);
   }
 
+  validateColor (color, alternative = 'rgb(255,255,255)') {
+    let validColor = alternative;
+    try {
+      validColor = Color(color).rgbString();
```
marcbachmann: There's no need for the `validColor` declaration. You can directly use `return Color(color).rgbString()` 😉

albinekb:
Code looks good, but I'm thinking that we would like to validate the whole config at the same place. This is a good start. Maybe a more common-validation of stuff like css/other config validation should be in another PR? Thoughts?

jwhitmarsh:
For sure - I can work on that in a separate PR. Are we happy with throwing warnings through a `notification`? There isn't much space there if there's more than one error in the config. Would it be better to write it to stdout?

Issue #216, opened 2016-07-17
Make mouse wheel events scroll through `less`, `git log` and `get diff` style output


hharnisc:
Really miss this feature from Terminal.app

When you're viewing scrollable output with Terminal.app you're able to use the mouse wheel or trackpad to scroll through the content.

**Terminal.app**

![scrollterminal](https://cloud.githubusercontent.com/assets/1388079/16903730/bfce6442-4c4b-11e6-8be8-8f41fc0ddb56.gif)

**HyperTerm**

![scrollhyperterm](https://cloud.githubusercontent.com/assets/1388079/16903733/cab85566-4c4b-11e6-868c-59ff88f2b23b.gif)

It seems like this might be possible to make a plugin for this, and possibly be a candidate for a default plugin.

lkzhao:
Checkout this guys. [hyperterm-alternatescroll](https://github.com/lkzhao/hyperterm-alternatescroll)

matmunn:
This plugin no longer appears to work

orcinus:
@MAPESO It doesn't.

chabou:
Just tested with our v2 and it works great on macOS.

orcinus:
Have you tested on Windows?

chabou:
@orcinus no, the original poster was talking about Terminal.app.
Please open an issue if Windows native terminal handles it and not Hyper.

orcinus:
Sending scroll and mouse messages to local apps and the remote host is not a Terminal.app specific feature. Right now, things like tmux mouse events, tmux buffer scroll etc. don't work at all in Hyper on Windows. As in there is *no* mouse messaging support *of any sort*. It's not a matter of "Windows native terminal handling it" or not handling it. It's a feature missing in Hyper.

Finally, the issue / request as reported was:
> Make mouse wheel events scroll through `less`, `git log` and `get diff` style output

not:
> Make Hyper behave as Terminal.app on Mac OS

But fine, i'll create a new issue.

chabou:
> Make mouse wheel events scroll through less, git log and get diff style output
It works perfectly on macOS (platform of the original poster). So this issue is fixed.

There is some existing issue with tmux, do not create a new one please.

PR #563, opened 2016-08-04
Fix mapXDispatch && allow plugin to access onWheel

lkzhao:
I am building a plugin to support #216. 
But I noticed that `mapXDispatch` functions in plugin are not called at all.
I also need the ability to tap in the `onWheel` function on the `Term`.

This simple PR fix these two issues.

### DIFF
--- a/lib/utils/plugins.js
+++ b/lib/utils/plugins.js
@@ -266,7 +266,28 @@ export function connect (stateFn, dispatchFn, c, d = {}) {
         });
         return ret;
       },
-      dispatchFn,
+      function (dispatch) {
+        let ret = dispatchFn(dispatch);
+        connectors[name].dispatch.forEach((fn) => {
+          let ret_;
+
+          try {
+            ret_ = fn(dispatch, ret);
+          } catch (err) {
+            console.error(err.stack);
+            notify('Plugin error', `${fn._pluginName}: Error occurred in \`map${name}Dispatch\`. Check Developer Tools for details.`);
+            return;
+          }
+
+          if (!ret_ || 'object' !== typeof ret_) {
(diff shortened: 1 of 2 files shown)


lkzhao:
The plugin is here:
### [hyperterm-alternatescroll](https://github.com/lkzhao/hyperterm-alternatescroll)

MrRio:
This looks really good 👍

MrRio:
I think your plugin should be part of the main branch, but we'll start with this.

MrRio:
Can you update the rebase the branch?

MrRio:
What happens to dispatchFn that gets passed in? Does that need to be considered?

MrRio:
Hey there,

First off, this looks really good 👍 I think this functionality should be part of the main branch. 

I can't actually get this work after applying the patch, and then adding the plugin. I'm on OSX, latest HyperTerm master. If you could look into that and rebase the commit that'd be great :)

Cheers,
James

lkzhao:
@MrRio 
Just rebased my fork. 

no idea why it isn't working for you though. I just did a setup on my friend's computer and got it working. Can you try that again?

lib/utils/plugins.js
```diff
@@ -266,7 +266,28 @@ export function connect (stateFn, dispatchFn, c, d = {}) {
         });
         return ret;
       },
-      dispatchFn,
+      function (dispatch) {
+        let ret = dispatchFn(dispatch);
```
lkzhao: `dispatchFn` is triggered here. Similar to how the `stateFn` works.

MrRio:
This is merged in https://github.com/zeit/hyperterm/pull/578 👍

PR #572, opened 2016-08-05
Added shellArgs to the config.

efegurkan:
In case someone wants to use non-login shells there is a field added to
the configuration file.
If shellArgs not set default shell will be used as before.

i.e. for using interactive shell (and use .bashrc instead of
.bash_profile on linux) use shellArgs: ['-i'] and shell: '/bin/bash'
this is useful especially if you are on Linux.

### DIFF
--- a/app/config-default.js
+++ b/app/config-default.js
@@ -56,6 +56,10 @@ module.exports = {
     // if left empty, your system's login shell will be used by default
     shell: '',
 
+    // for setting shell arguments (i.e. for using interactive shellArgs: ['-i'])
+    // by default ['--login'] will be used
+    shellArgs: ['--login'],
+
     // for environment variables
     env: {},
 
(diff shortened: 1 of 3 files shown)


rauchg:
+1

app/config-default.js
```diff
@@ -56,6 +56,9 @@ module.exports = {
     // if left empty, your system's login shell will be used by default
     shell: '',
 
+    // for setting shell arguments (i.e. for using interactive shell ['-i'])
+    // by default ['--login'] will be used
+
```
leo: It looks like `shellArgs` is missing.

leo:
Also please rebase! 🚀

app/config-default.js
efegurkan: oops! sent new one

leo:
Thanks! ❤️

PR #575, opened 2016-08-06
Use single quotes for better compatibility

albinekb:
When attempting to open preferences using `⌘+,` when using fish, i get the following error

```
❯  bash -c "exec env ${EDITOR:=nano} ~/.hyperterm.js"
${ is not a valid variable in fish.
fish:  bash -c "exec env ${EDITOR:=nano} ~/.hyperterm.js"
```

That's because how fish handles single vs double quotes

This PR makes opening preferences work as expected with `fish`, without changing the behaviour when using `bash`, i haven't tested it using `zsh`


MrRio:
Works with zsh!

screen shot 2016-08-06 at 09 44 00

PR #576, opened 2016-08-06
Open new window offset from last focused window (#573)

### LINKED ISSUE #573: New windows from the Shell menu open slightly across, whereas new windows from the dock don't
screen shot 2016-08-05 at 21 08 01
screen shot 2016-08-05 at 21 08 09
screen shot 2016-08-05 at 21 08 17

This is because the logic that checks for the current open window doesn't work from the dock.

albinekb:
When clicking "new window" in the dock, no window is focused, thus opening it at the same position as the old window.

This PR will make it use our awesome `getLastFocusedWindow` so that it will work as expected even if no window is focused.

Fixes #573

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -60,9 +60,9 @@ app.on('ready', () => {
     // Open the new window roughly the height of the header away from the
     // previous window. This also ensures in multi monitor setups that the
     // new terminal is on the correct screen.
-    if (BrowserWindow.getFocusedWindow() !== null) {
-      const currentWindow = BrowserWindow.getFocusedWindow();
-      const points = currentWindow.getPosition();
+    const focusedWindow = BrowserWindow.getFocusedWindow() || app.getLastFocusedWindow();
+    if (focusedWindow) {
+      const points = focusedWindow.getPosition();
       const currentScreen = screen.getDisplayNearestPoint({ x: points[0], y: points[1] });
 
       const biggestX = ((points[0] + 100 + width) - currentScreen.bounds.x);


MrRio:
Looking good, one small tweak if possible, when you do "New Window" multiple times in a row (strange use case I realise) it keeps opening in the same spot. The focused window needs to be set to the newly opened one.

lordgiotto:
We could simply trigger `win.focus()` after window creation (regardless this PR): it's an intuitive behaviour in my opinion, and could fix this issue.

MrRio:
@lordgiotto yeah agreed

lordgiotto:
@MrRio We should make another PR to keep commit history clean? Or we could append this modification to this PR?
Eventually I can do the PR if you wish.

lordgiotto:
Ok, I just realized that's not the problem: on window creation `win.show()` is triggered, so it gives focus to the window automatically. Therefore the cause of the issue pointed out by @MrRio is somewhere else: maybe simply a timing issue?

MrRio:
I've fixed it in another PR :)

MrRio:
The issue is that the onFocus event doesn't fire, which doesn't update the time used by getLastFocusedWindow

lordgiotto:
Yeah, it's called on init rpc event, that triggers after session is created and everything in borwserWindow is initialized. I'll comment on your new PR thought, i've a suggestion about encapsulation :)

MrRio:
Fixed in https://github.com/zeit/hyperterm/pull/581

PR #581, opened 2016-08-06
Open new window offset from last focused window (Credit: albinekb)

MrRio:
Full credit to albinekb -- this PR also fixes the issue with opening multi windows in a row.


MrRio:
See https://github.com/zeit/hyperterm/pull/576

albinekb:
Looks good 👍

app/index.js
```diff
@@ -103,6 +103,9 @@ app.on('ready', () => {
 
     win.loadURL(url);
 
+    // Make sure this newly opened window is seen to now have focus
+    win.focusTime = process.uptime();
```
lordgiotto: The idea is perfect :) 
However it's a risky design choice in my opinion. Maybe it would be better to just trigger win.focus() (if it works) or encapsulate both this `win.focusTime = process.uptime();` and the one at line 287 in a function called `registerFocusTime` (or something similar).
So, if in a future someone will patch how the focus time is stored, it will only to change that function and `getLastFocusedWindow()`.

My 2 cents ;)

MrRio: Agreed

MrRio:
I reckon we just emit the event when the window opens

app/index.js
```diff
@@ -283,6 +283,7 @@ app.on('ready', () => {
     win.on('focus', () => {
       win.focusTime = process.uptime();
     });
+    win.emit('focus');
```
albinekb: Maybe add a comment on why this is here 🙈

MrRio: Hah, yeah you're right

app/index.js
```diff
@@ -283,6 +283,9 @@ app.on('ready', () => {
     win.on('focus', () => {
       win.focusTime = process.uptime();
     });
+    // The focus event doesn't fire on win.show, but we need
+    // this to track the most recent window in focusedWindow
+    win.emit('focus');
```
lordgiotto: Please don't hate me!! 🙈 Only a small and maybe irrelevant thing about the comment.
`win.show` triggers focus correctly, according to the electron docs: but we trigger `win.show` too late for our purposes.

MrRio:
The event does fire from Shell -> New Window, but not from Dock -> New Window, I've added a bug ticket for that.  https://github.com/zeit/hyperterm/issues/583

leo:
@MrRio Please also close https://github.com/zeit/hyperterm/issues/573 and https://github.com/zeit/hyperterm/pull/576 if they have been fixed with this PR.

PR #588, opened 2016-08-07
Add support for Alt modified keys

dparnell:
Hi,

  I've been trying to use Emacs under Hyperterm and without meta key support this is very painful. I have made a little patch that allows Emacs to work as expected under Hyperterm. I'm not too happy with the `global` reference, but other than that it seems to work very nicely.

Thanks for making a fun terminal app that is easy to hack on :smile:

Daniel

### DIFF
--- a/app/config-default.js
+++ b/app/config-default.js
@@ -67,10 +67,13 @@ module.exports = {
     bell: 'SOUND',
 
     // if true, selected text will automatically be copied to the clipboard
-    copyOnSelect: false
+    copyOnSelect: false,
 
     // URL to custom bell
     // bellSoundURL: 'http://example.com/bell.mp3',
+
+    // send Alt-x keys to the underlying terminal escaped
+    pass_meta_keys: false
 
     // for advanced config flags please refer to https://hyperterm.org/#cfg
   },
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -102,6 +102,11 @@ hterm.Keyboard.prototype.onKeyDown_ = function (e) {
   }
 
   if (e.metaKey || e.altKey) {
+    if (e.key !== 'Alt' && e.altKey && global.config.getConfig().pass_meta_keys) {
+      // Send this key escaped to the underlying terminal so Emacs will work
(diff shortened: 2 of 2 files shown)


dparnell:
It looks like there are still a few things that don't quite work in Emacs, but this is an improvement

leo:
Please rebase! 😊

lordgiotto:
Exists an hterm setting for this: term_.prefs_.set('alt-is-meta', true)

Maybe it could be a cleaner solution that listening for the keydown?

leo:
We need another rebase... 😊

nicksrandall:
Any word on this?? Can it be merged?

app/config-default.js
```diff
@@ -67,11 +67,14 @@ module.exports = {
...
 
     // if true, selected text will automatically be copied to the clipboard
-    copyOnSelect: false
+    copyOnSelect: false,
 
     // URL to custom bell
     // bellSoundURL: 'http://example.com/bell.mp3',
 
+    // send Alt-x keys to the underlying terminal escaped
+    pass_meta_keys: false
```
ekmartin: This should be camel cased, i.e. `passMetaKeys`

lib/hterm.js
```diff
@@ -102,13 +102,19 @@ hterm.Keyboard.prototype.onKeyDown_ = function (e) {
   }
 
   if (e.metaKey || e.altKey) {
+    if (e.key !== 'Alt' && e.altKey && global.config.getConfig().pass_meta_keys) {
```
ekmartin: Using `config.getConfig()` skips the decoration process (so you won't get the config processed by plugins). The `getConfig` wrapper in `lib/utils/config` fixes this. Normally config values are loaded into the UI reducer though, instead of accessed directly like here. I'm not sure if that's necessary here though, since you'll do the check on every alt button press (and thus won't need the config to reload or similar).

ekmartin:
`alt-is-meta` like @lordgiotto suggests is probably better though. Then you need to add the config option to the UI reducer (which is kind of misleading I guess, since it's not really UI...), so that it'll get reloaded when the config changes.

ekmartin:
Closing this as it was superseded by #686.

PR #605, opened 2016-08-10
Use os-locale to define $LANG in the terminal

MrRio:


MrRio:
This isn't the whole picture unfortunately... will update PR with more details soon

CodeTheory:
I think using `os-locale` might, for now, fix problems for a few users even if this isn't the whole issue. Should we merge this in in that case?

rauchg:
LGTM!

lordgiotto:
I think `os-locale` is not the right solution for our issue: I tested this PR talking with @MrRio on Slack last month, and I tested it again now. 
I guess the problem is due the fact that it uses `locale` check in first instance. So this is what happens in 3 different cases:

- npm start form iTerm2 (where the right locale is set by default): in hyperterm LANG inherits the right one `LANG="it_IT.UTF-8"`
- npm start from hyperterm 0.7.1 (where my locale is wrong set to `it.UTF-8`): LANG is set to `LANG="it.UTF-8"`
- packaged app: LANG is completely wrong formatted `LANG=""C".UTF-8"`

leo:
Shouldn't the dependency be defined in `app/`?

Stanzilla:
I made a rebased version at https://github.com/zeit/hyper/pull/2913

Stanzilla:
Thanks for your contribution, now committed as https://github.com/zeit/hyper/pull/2913

PR #608, opened 2016-08-11
Provide clear selection of text in terminal view

### LINKED ISSUE #591: Pressing ESC should clear selection
Pressing ESC should clear selection as in native `terminal` and `Iterm2`

**Steps to reproduce:**
- Open HyperTerm
- Press `CMD+A` to select everything
- Press `ESC`

**Expected:**
- Selection to cleared

**Actual:**
- Selection persisted.

ppot:
- Click on Terminal view now deselect text.
- Valid keyboard event now deselect text.
- Add configuration option for copy-on-select parameter

Fixes #591

### DIFF
--- a/app/config-default.js
+++ b/app/config-default.js
@@ -64,7 +64,10 @@ module.exports = {
     env: {},
 
     // set to false for no bell
-    bell: 'SOUND'
+    bell: 'SOUND',
+
+    // set to false for no copy-on-select
+    copyOnSelect: false
 
     // URL to custom bell
     // bellSoundURL: 'http://example.com/bell.mp3',
--- a/app/config-default.js
+++ b/app/config-default.js
@@ -66,7 +66,7 @@ module.exports = {
     // set to false for no bell
     bell: 'SOUND',
 
-    // set to false for no copy-on-select
+    // if true, selected text will automatically be copied to the clipboard
     copyOnSelect: false
 
     // URL to custom bell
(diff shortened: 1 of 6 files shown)


leo:
@ppot Great fix! Please rebase though.

lib/components/term.js
```diff
@@ -38,6 +38,7 @@ export default class Term extends Component {
     this.term.prefs_.set('receive-encoding', 'raw');
     this.term.prefs_.set('send-encoding', 'raw');
     this.term.prefs_.set('alt-sends-what', 'browser-key');
+    this.term.prefs_.set('copy-on-select', false);
```
rauchg: For another PR, this could be an option. A lot of Linux users are used to copy on select from xterm

ppot: Got it. I will work on an option setting for this. Any preferences ?

ppot:
@leo Rebased and ready for a merge!

ppot:
@rauchg  Configuration add for copy-on-select to this PR!

app/config-default.js
```diff
@@ -60,7 +60,10 @@ module.exports = {
     env: {},
 
     // set to false for no bell
-    bell: 'SOUND'
+    bell: 'SOUND',
+
+    // set to false for no copy-on-select
```
rauchg: the comment should be more descriptive? Like "If `true`, selected text will automatically be copied to the clipboard.

lib/components/term.js
```diff
@@ -45,6 +45,12 @@ export default class Term extends Component {
       this.term.prefs_.set('audible-bell-sound', '');
     }
 
+    if (props.copyOnSelect) {
+      this.term.prefs_.set('copy-on-select', true);
+    } else {
+      this.term.prefs_.set('copy-on-select', false);
+    }
+
```
rauchg: This should also be in `componentWillReceiveProps`

ppot: Updated and rebased in last commit.

leo:
Thanks! ❤️

PR #615, opened 2016-08-12
Add a Dockerfile for easy builds

rjsteinert:
My Mac doesn't meet the dependencies for building this. I'm getting some complaint about my version of Python. I don't really want to mod my Mac for every project's own unique needs these days so I use Dockerfiles as build machines so the build can happen in a container. This is is nice for documenting the requirements for building, documenting the the process for building, and lastly for consistency of builds not to mention avoiding the "Oh shit we lost Bob's laptop, now we can't compile, we have to shut the project down" problem. 

Now the only dependency is Docker so we can do a build like...
```
$ git clone https://github.com/zeit/hyperterm.git
$ cd hyperterm
$ docker build -t hyperterm .
$ docker run -v $(pwd)/dist:/hyperterm/dist hyperterm
```

The result is the same. You'll find a `dist` folder in your hyperterm directory on your host machine with the builds that happened inside the container.

... of course, the result of this still complains about missing Python so the build fails. Can someone solve this? I don't know what version of Python this project requires. Bob does.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,12 @@
 FROM ubuntu:14.04 
-RUN curl -sL https://deb.nodesource.com/setup_4.x | bash -
-RUN apt-get -y install nodejs
+RUN apt-get update
+RUN apt-get install -y curl software-properties-common build-essential git
+RUN sudo add-apt-repository ppa:fkrull/deadsnakes && \
+    sudo apt-get update && \
+    sudo apt-get install -y python2.7 && \
+    ln -s /usr/bin/python2.7 /usr/bin/python
+RUN curl -sL https://deb.nodesource.com/setup_4.x | bash - && \
+    apt-get -y install nodejs
 ADD . /hyperterm
 VOLUME /hyperterm/dist
 WORKDIR /hyperterm
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM ubuntu:14.04 
+RUN curl -sL https://deb.nodesource.com/setup_4.x | bash -
+RUN apt-get -y install nodejs
+ADD . /hyperterm
+VOLUME /hyperterm/dist
+WORKDIR /hyperterm
+RUN npm install
+ENTRYPOINT npm run pack
--- a/README.md
+++ b/README.md
@@ -30,6 +30,14 @@ If you want to build the binaries for all specified platforms, run the command:
 $ npm run pack
 ```
 
+Or use Docker if your system doesn't meet a dependency.
+```bash
+$ git clone https://github.com/zeit/hyperterm.git
+$ cd hyperterm
+$ docker build -t hyperterm .
+$ docker run -v $(pwd)/dist:/hyperterm/dist hyperterm
+```
+
 After that, you'll see the binaries in the `./dist` folder!
 
 ## Related Repositories


TooTallNate:
I have Python 2.7.10 on my OS X machine, FYI

dj-hedgehog:
Why would you want to build it as a OS X user using Docker? 
All you get is the Unix-binary you can only use in a VM or in a container using xhyve (Docker for Mac)

rauchg:
I don't understand the python problem. Where are we using python? Is it one of our deps?

rjsteinert:
@dj-hedgehog I think I see what you are saying. Is it "If you build hyperterm in a Linux environment, it will only result in a build of hyperterm that works for Linux ". Is that right? That's not the case when creating executables for platforms using tools like Cordova. Is it always the case that with Electron you can only build apps for the platform you are building from?

rauchg:
Our build process works on Linux, Windows and OSX.

rjsteinert:
@TooTallNate Looks like I'm on Python 3.5.1. 
```
 ⚡ python --version
Python 3.5.1 :: Anaconda 2.4.1 (x86_64)
```

@rauchg Here's a log containing a bunch of errors around python when I try `npm install`.

[log.txt](https://github.com/zeit/hyperterm/files/416808/log.txt)

> Our build process works on Linux, Windows and OSX.

The question is not "Can we perform a build on X platform?" but more like "Can we do a build on X platform that produces an executable on Y platform?" In this case it's "Can we do a build in Ubuntu that produces and executable for Mac?" If so, then we can reliably build Hyperterm for Mac from a Mac that has Docker for Mac installed.

ppot:
@rjsteinert  Seem pretty much an issue with support of node-gyp.
Coming from node-gyp: 
`On Mac OS X:
python (v2.7 recommended, v3.x.x is not supported)`
What is your node version and npm version ?
Take a look at https://github.com/nodejs/node-gyp/issues/746

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ubuntu:14.04
```
fearphage: I would recommend (alpine-node](https://hub.docker.com/r/mhart/alpine-node/). Images are way smaller.

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ubuntu:14.04 
+RUN apt-get update
+RUN apt-get install -y curl
+RUN curl -sL https://deb.nodesource.com/setup_4.x | bash -
```
fearphage: Also you can add remote files to the host.

```dockerfile
ADD https://deb.nodesource.com/setup_4.x /tmp/install-node.sh
```

The size of that script is much smaller than all the cache (and time) that Ubuntu pulls down just to install `curl`.

Dockerfile
jbergstroem: FYI, Node.js compiled against glibc isn't officially (upstream) supported on a musl-based system.

jbergstroem:
@rjsteinert Python 3 isn't supported by Node.js (through gyp). I wouldn't necessarily call this a hyperterm-related issue seeing how you will run into the same problem on any system using Python 3 and Node.js. There's "somewhat" progress [here](https://github.com/nodejs/node-gyp/issues/193) but I don't think one should expect py3 support in gyp anytime soon.

Dockerfile
fearphage: Understood.

rjsteinert:
Woohoo! This works now as described in my change to the README. Only problem is, as @dj-hedgehog alluded to, all I have generated is an executable for Linux when I'm ultimately trying to generate an executable for Mac. So this is a reliable way for generating the Linux version of Hyperterm on either Mac or Linux without having to modify your python version but not yet a way to reliably generate the Mac version. 

As I mentioned before, tools like Cordova allow you to compile for more than one "platform" as they call it. I wonder if we can do the same so that your Linux container can produce the Mac`.app`.

Dockerfile
rjsteinert: @fearphage I think the last time I tried to base an image on alpine I gave up because of a lack of package manager. Also, cool tip on the using `ADD` instead of `curl`! Unfortunately I rebase my last commit to this in and it looks like it blew away your comment on what you suggested. I'm surprised Github even removed it from the flow of this comments on this PR :(.

flybayer:
@rjsteinert, does this provide any insight for you?

https://github.com/electron-userland/electron-builder/wiki/Multi-Platform-Build

gurinderhans:
Possible to try this? https://github.com/electron-userland/electron-packager . I am not fully successfull with it however, the build works perfectly (I can build from linux docker container for macOS) however when I try to run it, I get this.
screen shot 2016-09-18 at 2 21 54 pm

I'm guessing this has to do with `hyperterm` being dependent on the system `npm` instead of using its own internal `npm`.

If someone else gets this to work, however it may be, using Docker, please comment here :)

rauchg:
I think this is a lot of work for limited effect. Like it was mentioned, this only gets you a Linux build. The only way to get all the builds right is to rely on the travis matrix. I appreciate the work, but ultimately I'd like one less thing to maintain.

Issue #311, opened 2016-07-20
Save window groups


jrdnbwmn:
Idea for a new feature: ability to save window groups [like you can in Terminal](http://osxdaily.com/2013/12/09/use-manage-terminal-window-groups-mac/). 

This is something I and a lot of other devs I know use all the time. I looked through current functionality and plugins and couldn't find anything that does this.

May be best as a plugin.

felipemrodrigues:
+1

rdewolff:
+1

chabou:
Saving windows/tabs/panes and their working directory is enough?
Or restoring remote SSH sessions is a big part of this ?

cbrwizard:
https://github.com/timolins/hyperlayout seems to solve this

jaredramirez:
+1

felipemrodrigues:
@cbrwizard nice tip

cguodesign:
+1

gurmukhp:
Any update on this? I'd love to be able to remember window layouts on reload.

alongenosar:
+1

ppot:
I could make this as a plugin ;) before session restoration if it's only the windowa

alongenosar:
you mean panes?

ppot:
@alongenosar I could hack on this this weekend. Well see what I cam up with. If you only want one save at the end I can invoke something. But if you want memory session restoration, i'm not there yet. Backend dosen't support this as a plugin.

alongenosar:
great

ppot:
What i'm working on is a way to: Set an environment.
What I meant is
1. Open window and pane etc.
2. Save as profile.
3. Define base profile on startup.
4. Open different profile.

alongenosar:
Sounds great! 
let me know when I can test it.

ppot:
Target 1: update Hyper window object: #3485

NeonWizard:
Hyperlayout seems to be discontinued and is no longer accepting pull requests. I would very much prefer if this basic functionality was provided in vanilla Hyper, it's been asked many times over the last several years.

linareyne:
Yeah, still nothing it looks like? This is such a basic feature, it's criminal for hyper to not have it. Makes it not very suited to use for a work environment, unfortunately.

PR #617, opened 2016-08-13
Memorize window position and size

### LINKED ISSUE #27: Hyperterm should remember the window size upon restart

ppot:
Since Terminal.app dosen't remember window size and only window position, this provide memorizing of the window position on close. On restart, it will use the last saved windows position.

- Use `electron-config` to record window X,Y position
- Use `electron-config` to record window size

Position interpretation order:
- If an `options.position` is set, the parameter is first used
- If the last window position is cached, it will be used 
- then if none of the above, the default `[50, 50]` is used

Size interpretation order:
- if `options.size` is set, the parameter is first used
- if  `windowSize` is set in `hyperterm.js`, the parameter is used as default size
- If the last window size is cached, it will then be used as default
- then if none of the above, the default `[540, 380]` is used

--

Here is a plugin code as en example and the result
```
'use strict';
const { globalShortcut } = require('electron');
exports.onApp = (app) => {
      app.createWindow(undefined,{
            position: [50,50], 
            size: [540,200]
      });
      app.createWindow(undefined, {
          position: [600,50], 
          size: [540,200]
      });
      app.createWindow(undefined, {
          position: [50,260], 
          size: [540,200]
      });
      app.createWindow(undefined, {
          position: [600,260], 
          size: [540,500]
      });
    });
};
```
This result in this:
screen shot 2016-08-16 at 9 00 27 pm

Fix #27
Provide information for windows group layout #311

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -9,7 +9,12 @@ const notify = require('./notify');
 const Config = require('electron-config');
 
 // local storage
-const winCfg = new Config();
+const winCfg = new Config({
+  defaults: {
+    position: [50, 50],
+    size: [540, 380]
+  }
+});
 
 const path = resolve(homedir(), '.hyperterm.js');
 const watchers = [];
@@ -101,11 +106,11 @@ exports.window = {
     let position = winCfg.get('windowPosition');
     let size = winCfg.get('windowSize');
     return {
-      position: (position !== undefined) ? position : [50, 50],
-      size: (size !== undefined) ? size : [540, 340]
+      position: (position !== undefined) ? position : winCfg.store.position,
+      size: (size !== undefined) ? size : winCfg.store.size
     };
   },
-  set (position, size) {
+  set (position, size, cwd) {
     this.position = position;
     this.size = size;
   },
--- a/app/config.js
+++ b/app/config.js
@@ -5,8 +5,8 @@ const vm = require('vm');
 
 const {dialog} = require('electron');
 const gaze = require('gaze');
-const notify = require('./notify');
 const Config = require('electron-config');
+const notify = require('./notify');
 
 // local storage
 const winCfg = new Config({
@@ -100,12 +100,12 @@ exports.getPlugins = function () {
 };
 
 exports.window = {
-  get () {
-    let position = winCfg.get('windowPosition');
-    let size = winCfg.get('windowSize');
+  get() {
+    const position = winCfg.get('windowPosition');
+    const size = winCfg.get('windowSize');
     return {position, size};
   },
-  recordState (win) {
+  recordState(win) {
     winCfg.set('windowPosition', win.getPosition());
     winCfg.set('windowSize', win.getSize());
   }
--- a/app/config.js
+++ b/app/config.js
@@ -94,11 +94,27 @@ exports.getPlugins = function () {
   };
 };
 
-exports.recordWindowState = function (position, size) {
-  cache.set('windowPosition', position);
-  cache.set('windowSize', size);
-};
-
-exports.getWindowState = function () {
-  return {'position': cache.get('windowPosition'), 'size': cache.get('windowSize')};
+exports.window = {
+  position: undefined,
+  size: undefined,
+  get: function () {
+    let position = cache.get('windowPosition');
+    let size = cache.get('windowSize');
+    return {
+      'position': (position !== undefined) ? position : [50, 50],
+      'size': (size !== undefined) ? size : [540, 380]
+    };
+  },
+  set: function (position, size) {
+    this.position = position;
+    this.size = size;
+  },
+  revoke: function () {
+    this.position = undefined;
+    this.size = undefined;
+  },
+  recordState: function (win) {
+    cache.set('windowPosition', win.getPosition());
+    cache.set('windowSize', win.getSize());
+  }
 };
--- a/app/config.js
+++ b/app/config.js
@@ -9,7 +9,7 @@ const notify = require('./notify');
 const Config = require('electron-config');
 
 // local storage
-const cache = new Config();
+const winCfg = new Config();
 
 const path = resolve(homedir(), '.hyperterm.js');
 const watchers = [];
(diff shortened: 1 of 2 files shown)


rauchg:
This looks awesome!

ppot:
Would it be a good idea to allow in plugins.js
`exports.decorateWindow = function () {
};`

app/index.js
```diff
@@ -51,17 +51,27 @@ app.on('ready', () => {
   function createWindow (fn) {
     let cfg = plugins.getDecoratedConfig();
 
-    const [width, height] = cfg.windowSize || [540, 380];
-    const { screen } = require('electron');
+    const winSet = app.config.window.get();
+
+    let startX = winSet.position[0];
+    let startY = winSet.position[1];
```
sindresorhus:
```js
let [startX, startY] = winSet.position;
```

app/config.js
```diff
@@ -84,3 +88,28 @@ exports.getPlugins = function () {
...
 };
+
+exports.window = {
+  position: undefined,
+  size: undefined,
+  get: function () {
+    let position = cache.get('windowPosition');
+    let size = cache.get('windowSize');
+    return {
+      'position': (position !== undefined) ? position : [50, 50],
```
sindresorhus: You can set defaults in the `electron-config` constructor.

app/config.js
```diff
@@ -84,3 +88,28 @@ exports.getPlugins = function () {
...
+      'size': (size !== undefined) ? size : [540, 380]
+    };
+  },
+  set: function (position, size) {
```
sindresorhus: I should use shorthand method syntax:

```js
set(position, size) {
```

app/index.js
```diff
@@ -51,17 +51,27 @@ app.on('ready', () => {
...
-    let startY = 50;
+    const [width, height] = app.config.window.size !== undefined ? app.config.window.size : (cfg.windowSize || winSet.size);
+    const { screen } = require('electron');
 
     // Open the new window roughly the height of the header away from the
     // previous window. This also ensures in multi monitor setups that the
     // new terminal is on the correct screen.
     const focusedWindow = BrowserWindow.getFocusedWindow() || app.getLastFocusedWindow();
-    if (focusedWindow) {
+    // Ignore focusedWindow if createWindow started from ascript or a plugins
```
sindresorhus: ascript => a script

And this sentence doesn't read very well.

app/index.js
```diff
@@ -51,17 +51,27 @@ app.on('ready', () => {
...
+    // setting the position and the size.
+    // Only walid when size and position are strictly set.
```
sindresorhus: walid => valid

app/index.js
```diff
@@ -51,17 +51,27 @@ app.on('ready', () => {
...
+    if (app.config.window.position !== undefined) {
+      startX = app.config.window.position[0];
+      startY = app.config.window.position[1];
```
sindresorhus: Use destructuring

app/config.js
```diff
@@ -8,7 +8,9 @@ const notify = require('./notify');
 const Config = require('electron-config');
 
 // local storage
-const cache = new Config();
+const winCfg = new Config({name: 'winConf'});
```
sindresorhus: I don't think you want to set a name. That would create a separate config file for just this. Better to have it in the main config.

app/config.js
```diff
@@ -8,7 +8,9 @@ const notify = require('./notify');
...
+winCfg.defaultPosition = [50, 50];
+winCfg.defaultSize = [540, 380];
```
sindresorhus: That's not what I meant by defaults. See: https://github.com/sindresorhus/electron-config#defaults

app/config.js
```diff
@@ -84,3 +90,28 @@ exports.getPlugins = function () {
...
+
+exports.window = {
+  position: undefined,
+  size: undefined,
+  get () {
+    let position = winCfg.get('windowPosition');
+    let size = winCfg.get('windowSize');
+    return {
+      'position': (position !== undefined) ? position : winCfg.defaultPosition,
+      'size': (size !== undefined) ? size : winCfg.defaultSize
```
sindresorhus: No reason to quote these keys

app/config.js
ppot: Ya, trying to set it like 
`const winCfg = new Config({'defaultPosition':[50, 50], 'defaultSize': [540, 380]});`

But I don't think its the right way.

ppot: @sindresorhus Do you have a more suited example ?

sindresorhus:
```js
const winCfg = new Config({
	defaults: {
		position: [50, 50],
		size: [540, 380]
	}
});
```

app/index.js
```diff
@@ -51,17 +51,25 @@ app.on('ready', () => {
...
 
     // Open the new window roughly the height of the header away from the
     // previous window. This also ensures in multi monitor setups that the
     // new terminal is on the correct screen.
     const focusedWindow = BrowserWindow.getFocusedWindow() || app.getLastFocusedWindow();
-    if (focusedWindow) {
+    // In case of strictly set position and size, we should ignore the focusedWindow.
+    if (winPos !== undefined) {
+      [startX, startY] = winPos;
+      // Revoke config to prevent undesired behavior
```
sindresorhus: The comment should answer what kind of undesired behavior it would cause.

app/config.js
```diff
@@ -84,3 +88,31 @@ exports.getPlugins = function () {
...
+    let size = winCfg.get('windowSize');
+    return {
+      position: (position !== undefined) ? position : [50, 50],
+      size: (size !== undefined) ? size : [540, 340]
+    };
+  },
+  set (position, size, cwd) {
+    this.position = position;
+    this.size = size;
+    this.cwd = cwd;
```
sindresorhus: Why save the `cwd`? From what I can see, you're actually not setting it anywhere.

app/config.js
```diff
@@ -84,3 +93,18 @@ exports.getPlugins = function () {
...
   };
 };
+
+exports.window = {
+  get () {
+    let position = winCfg.get('windowPosition');
+    let size = winCfg.get('windowSize');
+    return {
+      position: (position !== undefined) ? position : winCfg.store.position,
+      size: (size !== undefined) ? size : winCfg.store.size
```
sindresorhus: You've set `defaults` in the Config constructor, so `position` and `size` will never be `undefined`.

app/index.js
```diff
@@ -48,20 +48,25 @@ const url = 'file://' + resolve(
 console.log('electron will open', url);
 
 app.on('ready', () => {
-  function createWindow (fn) {
+  function createWindow (fn, options = {defaults: {position: undefined, size: undefined}}) {
```
sindresorhus: It's weird to nest them in a `defaults` object. If you're gonna have defaults, just have an option object.

```js
function createWindow (fn, options = {position: undefined, size: undefined}) {
```

Though, I fail to see the point of this default. It's not even used anywhere.

ppot: I agree, Just testing if options.position or options.size if undefined is better than using default.

rauchg:
This needs a rebase and I'll review!

ppot:
@rauchg Currently rebased!

Issue #619, opened 2016-08-14
react components in plugins


mathisonian:
Is there any way that we could make `require('react')` work in plugins? Passing `React` down makes sense, but it becomes a problem if you want to e.g. require a custom react component from npm that expects react to be available

mathisonian:
Here is a more concrete example of the usecase / issue:

```js
exports.decorateTerm = (Term, { React, notify }) => {

  // this library calls require('react') internally
  const Spinner = require('react-spinner');

  return class extends React.Component {
    render () {
      const children = [];
      children.push(React.createElement(Term, Object.assign({}, this.props, { key: 'key1' })));
      children.push(React.createElement(Spinner, {key: 'key2'}));
      return React.createElement('div', {style: {width: '100%', height: '100%', position: 'relative'}}, children);
    }
  }
};
```

rauchg:
Definitely interested in making this work.

mathisonian:
This is coming up when trying to make a plugin depending on https://github.com/FormidableLabs/victory

![hyperchart](https://cloud.githubusercontent.com/assets/1074773/17652474/0acef1cc-624b-11e6-8c8f-a03a765377fa.gif)

Adding `react` as an explicit dependency on the plugin works for simple things but anything complex gives this warning and causes rendering bugs https://facebook.github.io/react/warnings/refs-must-have-owner.html

I'm not sure what the best approach would be to give the plugin access. Somehow expose the node_modules folder from hyperterm itself?

TooTallNate:
Isn't this what `NODE_PATH` env var is for?

albinekb:
This is fixed in #622 👍 
found that 'react-dom' was missing, added here: https://github.com/zeit/hyper/pull/1878

PR #622, opened 2016-08-15
allow plugins to require('react')

mathisonian:
Related to #619, this PR allows plugin authors to `require('react')` in their libraries without having to have react as a dependency, and helps avoid ["multiple copies of react" errors](https://facebook.github.io/react/warnings/refs-must-have-owner.html).

It is a fairly hacky solution to patch the `require` function, I'm curious to hear if others have suggestions for an alternative implementation.

### DIFF
--- a/lib/utils/plugins.js
+++ b/lib/utils/plugins.js
@@ -6,6 +6,15 @@ import React from 'react';
 import Notification from '../components/notification';
 import notify from './notify';
 
+var Module = require('module');
+var originalLoad = Module._load;
+Module._load = function (path) {
+  if (path === 'react') {
+    return React;
+  }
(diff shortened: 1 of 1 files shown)


matheuss:
I would mock `Module._load` instead of the entire `require` 💭

mathisonian:
I've updated to override `_load` per @matheuss's suggestion, let me know if other changes are necessary

matheuss:
Hey @mathisonian I'm sorry for the delay. 

`Module.prototype._load` does not exists 😰 Can you change it to `Module._load` and rebase it? Then I think we can merge 😊

mathisonian:
Updated 👍 thanks @matheuss

matheuss:
Thank you! :)

PR #641, opened 2016-08-21
Add dblClick Prevent text expend condition 

ppot:
- When dblClick focus was on `X-SCREEN` or when focus was on`X-ROW` with no text, the selection was crashing since no element where selected


ppot:
@leo Minor fix to prevent dblClick selection expend on empty element!

PR #646, opened 2016-08-22
Use “Keep A Changelog” format

Zearin:
As specified by <http://keepachangelog.com> (github repo: olivierlacan/keep-a-changelog).

This is about 90% of the way there; a few small edits remain, but first I want to make sure that even this level of change is okay with the repo owners.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,181 +2,240 @@
 0.7.1 / 2016-07-26
 ==================
 
-  * notifications: fix `(notes)` link [@rauchg]
+### Added
+  * add a slight timeout to update the fontSmoothing pref (#407) [@mike-engel]
+
+### Changed
   * session: suffix `UTF-8` to locale [@rauchg]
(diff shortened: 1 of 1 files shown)


leo:
First of all: Thank you for the effort you've added to this. 😊

Secondly, I honestly don't think these kinds of changelogs will have a long life here at ZEIT. I will do my best to push the org into the direction of simply using [GitHub releases](https://github.com/zeit/hyperterm/releases). They're much cleaner and just made for proper release notes.

But until then: Sure, let's merge this!

Zearin:
> Secondly, I honestly don't think these kinds of changelogs will have a long life here at ZEIT. I will do my best to push the org into the direction of simply using GitHub releases. They're much cleaner and just made for proper release notes.

Cool.  I like GitHub Releases!  

These changes are not mutually exclusive, though.  Grouping the “Added”, “Changed”, “Removed”, etc. lists can be useful in the markdown descriptions of each GitHub Release, too. :)

PR #675, opened 2016-08-29
[FIX] return a Redux thunk for the INIT action

TooTallNate:
Before, Redux was throwing an error saying that a plain object was not being returned from the `init()` function which was breaking hyperterm.

### DIFF
--- a/lib/actions/index.js
+++ b/lib/actions/index.js
@@ -1,9 +1,13 @@
-import {INIT} from '../constants';
 import rpc from '../rpc';
+import { INIT } from '../constants';
 
-export function init() {
-  rpc.emit('init');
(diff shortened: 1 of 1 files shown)


chabou:
I was creating the same PR with the same solution.

Another solution is to put directly in the index file : 
```javascript
rpc.emit('init');
```
but I think it's a good thing to dispatch a redux 'INIT' action (for plugins especially)

ekmartin:
Sorry about the delay - do you mind rebasing @TooTallNate?

rauchg:
I like this solution better than what I just merged

TooTallNate:
Ok, rebased

PR #686, opened 2016-09-02
Add Setting for modifier Keys cmdIsMeta and altIsMeta

aurelienbottazini:
Adds a setting to allow the user to use the Cmd Key (OSX) or alt Key as meta.

Tried to pick sensible setting names for that.

I also tried another solution, using the hterm setting `term_.prefs_.set('alt-is-meta', true)`. But for some reason even when I set up this value to true, it does not work. There is something I am not understanding there.

### DIFF
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -102,6 +102,24 @@ hterm.Keyboard.prototype.onKeyDown_ = function (e) {
     console.warn('Uncaught dead key on international keyboard', e);
   }
 
+  const modifierKeysConf = global.config.getConfig().modifierKeys;
+  if (e.altKey &&
+      e.code !== 'alt' &&
+      e.code !== 'altLeft' &&
+      e.code !== 'altRight' &&
+      modifierKeysConf.altIsMeta) {
+    this.terminal.onVTKeystroke('\x1b' + String.fromCharCode(e.keyCode));
+    e.preventDefault();
+  }
+
+  if (e.metaKey &&
+      e.code !== 'MetaLeft' &&
+      e.code !== 'MetaRight' &&
+      modifierKeysConf.cmdIsMeta) {
+    this.terminal.onVTKeystroke('\x1b' + String.fromCharCode(e.keyCode));
+    e.preventDefault();
+  }
+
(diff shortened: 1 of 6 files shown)


lib/hterm.js
```diff
@@ -101,6 +101,23 @@ hterm.Keyboard.prototype.onKeyDown_ = function (e) {
     console.warn('Uncaught dead key on international keyboard', e);
   }
 
+  const modifierKeysConf = global.config.getConfig().modifierKeys;
+  if (e.altKey &&
+      e.code !== 'altLeft' &&
+      e.code !== 'altRight' &&
```
aurelienbottazini: Checking `e.code` because I don't want to send the alt KeyCode to the terminal. It would result to strange characters output in hyperterm.

Same on line 115 for the MetaKey

rauchg:
This is fantastic.
- Would you mind documenting it on the website repo as well?
- We should leave it out of the default config because it's kind of advanced
- Instead of reading the config directly, please populate the default state in the reducer, and pass down the prop

nicksrandall:
This seems to be trying to solve the same thing as #588. Based on the feedback of that PR, this PR seems more in line with the "right" way to do this. What do we need in order to get this PR merged?

aurelienbottazini:
I will double check on Sunday and finish my changes.

I am now using the default reducer state as suggested and passing down the prop.
Removed the setting from the default conf

What's missing now is just the documentation on the website repo. Will do that on sunday as well

aurelienbottazini:
@rauchg 

- [ X] Would you mind documenting it on the website repo as well?
   Not at all, here is the PR: https://github.com/zeit/hyperterm-website/pull/45
- [X] We should leave it out of the default config because it's kind of advanced
- [X] Instead of reading the config directly, please populate the default state in the reducer, and pass down the prop

nicksrandall:
I can't thank you enough @aurelienbottazini! I'm looking forward to this getting merged.

nicksrandall:
@rauchg Can we get this merged?

nicksrandall:
@leo Can you merge this?

leo:
Sure! 😛

Issue #689, opened 2016-09-05
Plugin updates are not recognized (hyperterm says "No changes!")


trevordmiller:
I've created a HyperTerm plugin (color scheme) called `nova-hyperterm`.
After publishing on npm and placing the plugin in my `.hyperterm.js` `plugins` array, everything worked as expected.
I've fixed some issues and published patch versions to npm as you can see in the attached screenshot.
When I restart HyperTerm and/or go to `Plugins > Update All Now`, HyperTerm doesn't recognize that my plugin has been updated - it says "Plugins Updated" "No changes!".

The only way I was able to update my plugin was by deleting and re-installing the npm modules in `~/.hyperterm_modules`.

![screen shot 2016-09-04 at 7 50 31 pm](https://cloud.githubusercontent.com/assets/5497885/18235562/bcd792ea-72d9-11e6-96ef-5b05b740568e.png)

![screen shot 2016-09-04 at 7 41 13 pm](https://cloud.githubusercontent.com/assets/5497885/18235464/7057d34a-72d8-11e6-811c-152f6afd985c.png)

![screen shot 2016-09-04 at 7 40 45 pm](https://cloud.githubusercontent.com/assets/5497885/18235465/705a8360-72d8-11e6-8784-816834df5fc9.png)

![screen shot 2016-09-04 at 7 40 31 pm](https://cloud.githubusercontent.com/assets/5497885/18235466/7067b85a-72d8-11e6-80fc-672b92620d50.png)

p.s. Thank you for your amazing work on HyperTerm! I already love it and after these few bugs have been ironed out I think it will be the best option for a modern terminal 💪 💥 Here is what my full setup looks like with HyperTerm + Vim + Tmux (all using my custom "Nova" color scheme):

![nova-hyperterm-vim-tmux](https://cloud.githubusercontent.com/assets/5497885/18235601/546754d8-72da-11e6-92e2-ffbe28cbfe42.png)

trevordmiller:
I also tried pushing a `minor` version (`1.1.0`) and that didn't work either. It seems that the `latest` tag in the `~/.hyperterm_plugins/package.json` isn't working correctly. If someone could take a look at this and fix it I would greatly appreciate it :) My hunch is that nobody is actually getting any updates right now. Unless there is something wrong with my package specifically, this would appear to be a global problem for all users not getting updates without a manual wipe/re-install of `./hyperterm_modules/node_modules`

[...]

PR #690, opened 2016-09-05
Fix semver range

trevordmiller:
`latest` doesn't seem to be a valid semver range. Per https://github.com/zeit/hyperterm/issues/689.


matheuss:
It is not a valid `semver` range but it's a valid npm tag 😄 See https://docs.npmjs.com/files/package.json#dependencies

How would this fix #689?

trevordmiller:
> "How would this fix #689?"

That was the only thing I could think of for why HyperTerm isn't getting updates; but good to know that's not it.

PR #556, opened 2016-08-03
Improve developer pleasure

krzkaczor:
This PR adds:
 - one command to start development `npm run dev`
 - **react and redux dev tools**
 - redux logger middleware

These things work only in development mode. It's greatly inspired by https://github.com/chentsulin/electron-react-boilerplate

PR #691, opened 2016-09-05
Add support for Chrome DevTools extensions

Rowno:
Allows people to install arbitrary Chrome DevTools extensions via their config file. Also adds integration for the Redux DevTools extension.


lib/index.js
```diff
@@ -22,11 +22,15 @@ webFrame.setZoomLevelLimits(1, 1);
...
-    effects
+  compose(
+    applyMiddleware(
+      thunk,
+      plugins.middleware,
+      thunk,
+      effects
+    ),
+    // Redux DevTools extension integration
+    window.devToolsExtension ? window.devToolsExtension() : f => f
```
timneutkens: Just curious what `window.devToolsExtension ? window.devToolsExtension() : f => f` does?
Could it be simplified to: `window.devToolsExtension ? window.devToolsExtension() : false`? Since I don't see why there should be an arrow function there?

Rowno: Not sure, I just used the recommended setup: https://github.com/zalmoxisus/redux-devtools-extension#12-advanced-store-setup

I'm guessing `compose()` expects a function, so a noop function that just returns what it's given (`f => f`) is used when the Redux Devtools extension hasn't been installed yet.

timneutkens: Just implemented redux-devtools yesterday in another project. since it's a middleware what I was saying makes no sense at all haha.

rauchg:
This is incredible. I wonder if bundling and requiring `electron-devtools-installer` adds significant overhead to bootup time? Since it's a dev only feature, we have to be careful there

Rowno:
I'm not sure if this pull request is still relevant since #556 was merged?

Rowno:
Using a plugin for this is a better approach: https://www.npmjs.com/package/hyperterm-install-devtools

PR #693, opened 2016-09-06
Split Panes

ekmartin:
This adds an initial implementation of split panes, as shown in the screenshot below.
Split Panes

## Features
* Split panes can be created by hitting `Cmd+D` and `Cmd+Shift+D` (which are also accessible from the `Window` menu).
* Split panes can be arbitrarily nested, and support tabs like you'd expect.
* The active pane can be changed with mouse clicks, or by cycling through them using `Show Next Pane`/`Show Previous Pane`.
* Panes can be closed with `CmdOrCtrl+W`.

## How
There's definitely lots of ways to implement something like this, however I've opted for a pretty Redux heavy direction. The whole split pane structure is stored in a new reducer, `term-groups.js`, which reacts to create/remove/focus change actions and adapts the structure as needed.

### Reducer Structure
The `term-groups` reducer maintains a tree like structure of the different tabs.
The split structure in the screenshot would resemble something like this:

```
   +-----+Root+----+
   |               |
   v               v
++Left++       ++Right++
|      |       |   +   |
v      v       v   v   v
S      S       S   S   S
```

Each term group would have the following properties:
* `uid` - group identifier
* `parentUid` - the parent's `uid`
* `sessionUid` - if the group is a leaf node, conforms to the actual session's `uid`
* `direction` - what direction the split is created in (only for internal nodes)
* `children` - an array of uids of the term group's children.

Additionally, the reducer holds a mapping between each root term group and their respective active `sessionUid`, so that you can change tabs without losing the focus of the session you'd selected. 

### Caveats
To be able to create new term groups the `SESSION_ADD` action is hijacked, and the split direction (if it's a split that caused the new session to be created) is added [here](https://github.com/ekmartin/hyperterm/blob/splits/lib/actions/term-groups.js#L20-L23). This feels a bit dirty, and I'd love suggestions on how this could be improved (and suggestions on how to improve the data structure in general!).

### Rendering
The actual split panes are rendered using [react-split-pane](https://github.com/tomkp/react-split-pane) at the moment, but this probably needs to be changed before this is ready for primetime. A few issues:
* It doesn't support [more than two split panes](https://github.com/tomkp/react-split-pane/issues/81) out of the box without nesting split panes ([workaround](https://github.com/ekmartin/hyperterm/blob/splits/lib/components/term-group.js#L28-L47)).
* The styling is a slightly awkward. This could be fixed with an upstream pull request to allow custom class names for the pane divider (resizer), so we could use it with Aphrodite.
* Mouse resizing also seems a bit glitchy, but this might be an issue on our end as well - haven't spent too much time on it yet.

We could either look into fixing these issues upstream, or just implement our own flexbox heavy approach.

### ✅  Preserving already created hterm instances
At the moment we're creating hterm instances at the bottom of the React hierarchy, in the `Term` component. This doesn't work very well with split panes, since `Term`s will be removed and recreated when the split pane structure is changed. That means the same happens to their respective hterm instances, which means that sessions lose their output when splitting (i.e. needs to be fixed before splits are usable).

One of the ways of solving this would be to create hterm instances at a higher level, and then pass them down to the `Term` component. This could either be in a React component such as `Terms`, or even in a Redux reducer. The latter may sound a bit crazy, but I think it might make sense together with the current reducer structure. I'm interested in more ideas around this though.

### ✅  Hotkeys
This definitely needs some bikeshedding. Even if we need a way of configuring hotkeys in the future, sane defaults will always be important. As mentioned splits are currently created with `CtrlOrCmd+D` and `CtrlOrCmd+Shift+D`. `Cmd+D` is definitely okay, but `Ctrl+D` obviously isn't 😙. Maybe we can change it based on the user's OS. What would be a decent hotkey set for Linux?

There also isn't a hotkey for rotating back and forth between your open split panes yet. iTerm defaults this to `Cmd+[`/`Cmd+]`, however this doesn't work on a lot of European keyboards. Personally I'd prefer something like `Ctrl+Tab`/`Ctrl+Shift+Tab`, which we've currently set to changing tabs (afaik it doesn't work though, the combination is eaten by hterm). 

**EDIT:** Cycling between split panes is currently bound to `Ctrl+Alt+Tab` and `Ctrl+Shift+Alt+Tab`.

In the future we should definitely also support changing panes based on direction (`Cmd+Alt+Arrow` in iTerm).

### ✅  Resizing
Drag and drop resizing of split panes sort of works at the moment, but it's really glitchy.

### ✅  Styling
Needs to support custom styling in the same way as the rest of HyperTerm. We could probably add some sort of indicator to show what pane is in focus (i.e. an overlay over the inactive panes, like iTerm) as well.

### Bugs
- [x] Clicking to change focus doesn't work when applications such as `vtop/htop` are open.
- [x] Splitting panes will make `git status` output look funky (and probably other resizing issues) - [screenshot](https://cloud.githubusercontent.com/assets/3536982/18336213/60507c2c-7553-11e6-9e44-53a98961f342.png).
- [x] A tab will sometimes crash after a pane is closed, needs more investigation.
- [x] Closing the last tab won't exit HyperTerm (regression).
- [x] Creating multiple horizontal splits is buggy.
- [x] Webviews don't respect focus changes as they should.

...and probably a lot more!

## Binary
I really recommend pulling down the branch, however I've made a binary that can be downloaded here: https://github.com/ekmartin/hyperterm/releases/download/0.8.0-splits.5/mac.zip (updated 2016-10-02)


matheuss:
THAT'S FREAKING AWESOME

chabou:
Great work !

I found a little bug when you reduce the height of a horizontally splitted tab, the bottom one keeps its original size.
But it seems to be a react-split-pane bug : if I add ```display: flex```to the second Pane div, it fixes its size. 
Maybe related to https://github.com/tomkp/react-split-pane/issues/106

How can we help you with this PR ?

ekmartin:
Thanks! 

@chabou I think the main steps in moving forward will be deciding on what the correct direction to go regarding actions/reducers is, and what to do with the hterm instance issue I mentioned above. 

I'm going to be looking into what to do with `react-split-pane`, probably by seeing if reimplementing some of the functionality directly in HyperTerm might be better than using it.

If you're interested in helping out with some of the bugs that need to be squashed before it's ready that would be helpful. Quick summary:

* Clicking to change focus doesn't work when applications such as vtop/htop are open.
Normally this is done by attaching a `focus` listener [here](https://github.com/ekmartin/hyperterm/blob/splits/lib/components/term.js#L75-L78), however this is never fired when you for example click on a term that has `htop` open.

* A tab will sometimes crash after a pane is closed.
Try opening a few splits in different tabs, then closing them. Usually it works fine, but I've seen it crash a few times.

* Hotkeys for cycling back and forth between split panes.

piotr-yuxuan:
This feature (plus some minor bug corrections) is what stops me (aka the average coding Joe ^^) from switching from iTerm to hyperTerm
So I just drop this message as a support to the team! :)

iamstarkov:
seems like all todos are implemented and all bugs are resolved

iamstarkov:
@ekmartin @rauchg thank you for all the work done in this issue.
Can you share some news/plans on this pr?

ekmartin:
@iamstarkov: We found a few bugs with multiple horizontal (Y-directional) split panes, we're looking into that now :)

austinyearlykim:
It's crazy seeing a vital feature like this in progress in real time right now.  I'M SO STOKED

flybayer:
I just realized I'm testing `zet:add/splits` instead of `ekmartin:splits`. Which branch should we be testing?

maxdeviant:
@flybayer Doesn't `Cmd/Ctrl + -` conflict with zoom out?

We should just make the hotkeys configurable :smile:

flybayer:
@maxdeviant oops yes :laughing: I guess I'm actually using `Meta + -` and `Meta + \` and forgot the difference. Is `Meta + -` and `Meta + \` a viable default for this?

ekmartin:
@flybayer: the exit bug should be fixed with b2ca957 :)

ekmartin:
@flybayer: I added the split hotkeys `Ctrl+Shift+O` (horizontal) and `Ctrl+Shift+E` (vertical) - stolen from Gnome Terminator. If there's better options we can open an issue and do some bikeshedding after this has been merged :)

iamstarkov:
🎉 yay 🎉

iamstarkov:
any plans to release?

naim:
This would be really nice to get into mainline.

chabou:
I think it is included in 0.8.1 release https://github.com/zeit/hyper/releases/tag/0.8.1

iamstarkov:
@naim @chabou thats true

iamstarkov:
@naim though you need to download hyper. before it was hyperterm and has been renamed

chabou:
`In Progress` and `Review Needed` labels seem confusing 😄

PR #695, opened 2016-09-07
Refactor plugin retrieval

ekmartin:
The decorator methods for plugins and reducers were very similar, merged them into two functions  instead. It got a bit awkward for props because of the different function arity of the decorator methods, so open for suggestions there.

Also seems like we were missing a few methods in our "expose at least one of these" check, so plugins that i.e. only exposed `getTermProps` didn't work. The list should be complete now.

### DIFF
--- a/lib/utils/plugins.js
+++ b/lib/utils/plugins.js
@@ -35,6 +35,7 @@ let tabPropsDecorators;
 let tabsPropsDecorators;
 let termPropsDecorators;
 let termGroupPropsDecorators;
+let propsDecorators;
 
 // the fs locations where usr plugins are stored
 const {path, localPath} = plugins.getBasePaths();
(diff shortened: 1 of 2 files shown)


lib/utils/plugins.js
```diff
@@ -166,85 +179,44 @@ export function reload () {
...
       return;
     }
 
     if (!ret_ || 'object' !== typeof ret_) {
-      notify('Plugin error', `${fn._pluginName}: Invalid return value of \`getTermProps\` (object expected).`);
+      notify('Plugin error', `${fn._pluginName}: Invalid return value of \`${name}\` (object expected).`);
       return;
     }
 
-    props = ret_;
```
ekmartin: `getTermProps` was setting `props = ret_`, unlike the others which did `props_ = ret_`. Think that's a bug?

rauchg:
I've been meaning to do this. Great thinking @ekmartin

rauchg:
I added you as a repo collaborator 🎉

ekmartin:
Thanks @rauchg!

PR #707, opened 2016-09-13
autohide menubar by default

capaj:
this seems like a non issue on a Mac, but I think majority of Linux/windows users also prefer not to waste their screen space on the menu bar.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -81,6 +81,7 @@ app.on('ready', () => {
     }
 
     const browserDefaults = {
+      autoHideMenuBar: true,
       width,
       height,
       minHeight: 190,


MrRio:
You got screenshots/videos?

capaj:
@MrRio 
image

flybayer:
In Ubuntu, by default the application menu bar is in the top bar of the screen similar to OSX. So effectively the menu is already hidden for me. Will your change affect this behavior?

capaj:
@flybayer I don't think so. This only affects menu bar in ubuntu Gnome and Windows as far as I know.

capaj:
@leo rebased

leo:
I think what we forgot is that this also hides the menu bar on macOS - but we need it.

matheuss:
#946 😄

PR #723, opened 2016-09-20
[WIP] Use XO instead of Standard

matheuss:
TODO:
- [x] Fix React errors 

There was a discussion in [zeit.chat](https://zeit.chat) about changing the code style of HyperTerm's code base – and later on @zeit's code base (via @leo). We compared [XO](https://github.com/sindresorhus/xo), [Standard](https://github.com/feross/standard) and [Airbnb's JavaScript Style Guide](https://github.com/airbnb/javascript).

Since I use XO in my projects and I'm very happy with the rules it enforces, I'm sending this PR so we can discuss changing HyperTerm's code style to XO's 😄 

There was a lot of changes – many of them had to be done manually (`xo --fix` helped in the simpler ones), so I created a summary:

# Summary of changes

`const { something } = require('module');` => `const {something} = require('module');`

`fn({ key: 'value' });` => `fn({key: 'value'});`

`function name () {...}` => `function name() {...}`

`(arg) => {...}` => `arg => {...}`

`<Tag prop='value'>` => `<Tag prop="value">`

`return <Tag>;` => `return (<Tag>);`

`something("'");` => `something('\'');`

`let a, b, c;` => one `let` per line

```js
if (value !== 0) {
  // value is not zero
} else {
  // value is zero
}
``` 
was changed to 
```js
if (value === 0) {
  // value is zero
} else {
  // value is not zero
}
```

```js
something
? path1
: path2;
```
was changed to 
```js
something ?
path1 :
path2;
```

`if ('object' !== typeof value)` => `if (typeof value !== 'object')`

```js
if (something) {
  return true;
} else {
  return false;
}
```
was changed to 
```js
if (something) {
  return true;
}
return false;
```

`if (something) return true;`
was changed to 
```js
if (something) {
  return true;
}
```

`obj. hasOwnProperty(prop);` => `{}. hasOwnProperty.call(obj, prop);`

`==`, `!==` => `===`, `!==`

# Another changes

The `store` variable in[ Redux middlewares](http://redux.js.org/docs/advanced/Middleware.html) was removed  when it's not used – https://github.com/zeit/hyperterm/commit/8a8fdacffe370ca0e267a7cffbff6734bef43036#diff-6c6a3219ca05ccbe3ce89f7544a8f40aR8

Some rules relating to switch-case were disabled because the code was kinda complex and I didn't want to change it – https://github.com/zeit/hyperterm/commit/8a8fdacffe370ca0e267a7cffbff6734bef43036#diff-7da6d687f5a3beed5725bb8324e4b1adR82

The `babel/new-cap` rule was disabled due to things like this: https://github.com/zeit/hyperterm/commit/8a8fdacffe370ca0e267a7cffbff6734bef43036#diff-8303185c505c28257453bad1aadc768bR22

For some reason (probably related to `this` binding) the (only) test we have couldn't be changed from standard functions to arrow functions – https://github.com/zeit/hyperterm/commit/8a8fdacffe370ca0e267a7cffbff6734bef43036#diff-f7cc7aca73879ae481a889ab216ba926R1

# React rules

There's currently **151 errors** related to React code styling. Since I have 0 expertise with React, I didn't change anything related to it. Any help is more than welcome 😄 

# Conclusion 

Since these changes affect every contributor, we need to discuss them carefully. I'm 100% open to any help, input and feedback about them.

A lot of the changes **can** affect how the code works, so any help with testing it is very welcome.


leo:
The tab thing is actually broken on master. Not the result of our changes!

leo:
cc @sindresorhus

lib/actions/ui.js
```diff
@@ -121,36 +123,36 @@ export function moveRight () {
...
       index: i,
-      effect () {
-        const { sessions } = getState();
+      effect() {
+        const {sessions} = getState();
         const uid = sessions.activeUid;
         const sessionUids = keys(sessions.sessions);
         if (uid === sessionUids[i]) {
           console.log('ignoring same uid');
-        } else if (null != sessionUids[i]) {
```
ekmartin: `null != sessionUids[i]` checks for `undefined` as well @matheuss, now you're explicitly checking for the value `null`

lib/hterm.js
```diff
@@ -28,8 +29,8 @@ hterm.Terminal.prototype.overlaySize = function () {};
 // a non-collapsed selection whose text is '', and results
 // in an infinite copy loop
 hterm.Terminal.prototype.copySelectionToClipboard = function () {
-  var text = this.getSelectionText();
-  if (text != null && text !== '') {
+  const text = this.getSelectionText();
+  if (text !== null && text !== '') {
```
ekmartin: same here, text as `undefined` would get through `text !== null`  but not `text != null`. Maybe `if (text)` would suffice

lib/reducers/ui.js
```diff
@@ -91,71 +91,69 @@ const reducer = (state = initial, action) => {
             ret.fontSizeOverride = null;
           }
 
-          if (null != config.fontSize) {
```
ekmartin: same here

lib/actions/ui.js
matheuss: That's true @ekmartin 😰 What do you think we should do?

ekmartin: Set `eqeqeq` to `["error", "always", {"null": "ignore"}]` and revert the `== null` and `!= null` changes maybe?

http://eslint.org/docs/rules/eqeqeq#allow-null

sindresorhus: Checking against both `null` and `undefined` is an anti-pattern these days. Native things like "default parameters" only work on `undefined`, not `null`, so better to get used to `null` meaning an empty value, and `undefined` meaning no value.

ekmartin: Generally I agree, however I'm a bit unsure when it comes to config values. If it never makes sense for a specific config value to be null it might be better to prevent plugins and similar from setting it as such.

sindresorhus: Then I would rather throw a `TypeError` and be explicit about it.

But if you really want both `undefined` and `null`, I would strongly recommend just being verbose and explicit about it: `sessionUid[i] !== undefined && sessionUid[i] !== null`. Not everyone knows that `sessionUid[i] != null` coerces to the same. XO is about enforcing explicit and readable code.

sindresorhus:
Awesome! Let me know if you have any concerns or questions. I'm happy to discuss any of the rule choices. There's a good reasoning behind every single rule.

lib/actions/ui.js
ekmartin: Sounds reasonable :)

package.json
```diff
@@ -36,51 +36,47 @@
...
+      "babel/new-cap": 0,
+      "quote-props": 0,
+      "import/no-extraneous-dependencies": 0,
+      "no-warning-comments": 0,
+      "complexity": 0,
+      "react/no-danger": 0,
+      "react/no-string-refs": 0,
+      "react/jsx-key": 0,
+      "no-nested-ternary": 0,
+      "react/jsx-no-bind": 0
```
sindresorhus: Why?

> A bind call or arrow function in a JSX prop will create a brand new function on every single render. **This is bad for performance, as it will result in the garbage collector being invoked way more than is necessary.**

package.json
```diff
@@ -36,51 +36,47 @@
...
-      }
+      "react/jsx-filename-extension": 0,
+      "react/prop-types": 0,
+      "babel/new-cap": 0,
+      "quote-props": 0,
+      "import/no-extraneous-dependencies": 0,
+      "no-warning-comments": 0,
+      "complexity": 0,
+      "react/no-danger": 0,
+      "react/no-string-refs": 0,
```
sindresorhus: Why?

> Currently, two ways are supported by React to refer to components. **The first one, providing a string identifier is considered legacy in the official documentation.** Referring to components by setting a property on the this object in the reference callback is preferred.

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-string-refs.md

package.json
```diff
@@ -36,51 +36,47 @@
...
-      ],
-      "react/jsx-uses-react": "warn",
-      "react/jsx-uses-vars": "warn"
-    },
-    "parserOptions": {
-      "ecmaFeatures": {
-        "jsx": true
-      }
+      "react/jsx-filename-extension": 0,
+      "react/prop-types": 0,
```
sindresorhus: Why?

Seems like a good rule in general. Would be better to disable it inline for where it doesn't make sense or use the `ignore` option.

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md

package.json
```diff
@@ -36,51 +36,47 @@
...
+      "babel/new-cap": 0,
+      "quote-props": 0,
```
sindresorhus: This is meant for consistency. If you quote on key, you should probably quote the rest as they're strings. Happy to discuss this though. Currently it's set to `"consistent-as-needed"`.

http://eslint.org/docs/rules/quote-props

package.json
```diff
@@ -36,51 +36,47 @@
...
+      "import/no-extraneous-dependencies": 0,
+      "no-warning-comments": 0,
+      "complexity": 0,
+      "react/no-danger": 0,
```
sindresorhus: Why?

> Dangerous properties in React are those whose behavior is known to be a **common source of application vulnerabilities**. The properties names clearly indicate they are dangerous and should be avoided unless great care is taken.

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-danger.md

Would be better to fix the places those are used instead.

package.json
```diff
@@ -36,51 +36,47 @@
...
+      "react/no-string-refs": 0,
+      "react/jsx-key": 0,
```
sindresorhus: Why? Where did this cause problems?

Seems like a good rule: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-key.md

test/app/plugins.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable prefer-arrow-callback */
```
sindresorhus: Why not just use arrow functions?

webpack.config.js
```diff
@@ -1,6 +1,7 @@
+const path = require('path');
+
```
sindresorhus: No need for an empty line here.

lib/utils/file.js
```diff
@@ -6,12 +6,14 @@
...
+    return true;
+  }
 
   return Boolean(
-    (fileStat['mode'] & parseInt('0001', 8)) ||
-    (fileStat['mode'] & parseInt('0010', 8)) ||
-    (fileStat['mode'] & parseInt('0100', 8))
+    (fileStat.mode & parseInt('0001', 8)) ||
+    (fileStat.mode & parseInt('0010', 8)) ||
+    (fileStat.mode & parseInt('0100', 8))
```
sindresorhus: No rule for this yet, but these should use octal literals instead of `parseInt`.

http://www.2ality.com/2015/04/numbers-math-es6.html

lib/hterm.js
```diff
@@ -188,7 +189,7 @@ lib.colors.hexToRGB = function (arg) {
   }
 
   if (arg instanceof Array) {
-    for (var i = 0; i < arg.length; i++) {
+    for (let i = 0; i < arg.length; i++) {
```
sindresorhus: Here it would be better to use a `for..of` loop.

No rule for it [yet](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/49), but consider me a linter :p

lib/hterm.js
```diff
@@ -188,7 +189,7 @@ lib.colors.hexToRGB = function (arg) {
   }
 
   if (arg instanceof Array) {
```
sindresorhus: Use `Array.isArray()`. instanceof won't work across realms (frames/iframes/vm).

leo:
@sindresorhus Just for clarity: Yes, disabling these rules is definitely a bad practise and I completely agree with you. Which is why we're going to [break them down to zero](https://github.com/zeit/hyperterm/issues/730). The only reason why I haven't did that before merging the PR was because it's a lot of work... 😊 

As you can see in the issue description, I already put some effort into making the code comply with the rules, but these ones are left.

> I would like to help you get this down closer to zero...

We're thankful for every helpful hand that we can get (especially for yours)!

package.json
```diff
@@ -36,51 +36,47 @@
...
+      "no-nested-ternary": 0,
+      "react/jsx-no-bind": 0
     },
```
leo: https://github.com/zeit/hyperterm/pull/723#issuecomment-248816264

package.json
leo: https://github.com/zeit/hyperterm/pull/723#issuecomment-248816264

package.json
leo: https://github.com/zeit/hyperterm/pull/723#issuecomment-248816264

package.json
leo: https://github.com/zeit/hyperterm/pull/723#issuecomment-248816264

package.json
leo: https://github.com/zeit/hyperterm/pull/723#issuecomment-248816264

package.json
```diff
@@ -36,51 +36,47 @@
...
-      "ecmaFeatures": {
-        "jsx": true
-      }
+      "react/jsx-filename-extension": 0,
+      "react/prop-types": 0,
+      "babel/new-cap": 0,
+      "quote-props": 0,
+      "import/no-extraneous-dependencies": 0,
+      "no-warning-comments": 0,
+      "complexity": 0,
```
leo: https://github.com/zeit/hyperterm/pull/723#issuecomment-248816264

package.json
```diff
@@ -36,51 +36,47 @@
...
-    "parserOptions": {
-      "ecmaFeatures": {
-        "jsx": true
-      }
+      "react/jsx-filename-extension": 0,
+      "react/prop-types": 0,
+      "babel/new-cap": 0,
+      "quote-props": 0,
+      "import/no-extraneous-dependencies": 0,
+      "no-warning-comments": 0,
```
leo: https://github.com/zeit/hyperterm/pull/723#issuecomment-248816264

package.json
```diff
@@ -36,51 +36,47 @@
...
-    },
-    "parserOptions": {
-      "ecmaFeatures": {
-        "jsx": true
-      }
+      "react/jsx-filename-extension": 0,
+      "react/prop-types": 0,
+      "babel/new-cap": 0,
+      "quote-props": 0,
+      "import/no-extraneous-dependencies": 0,
```
leo: https://github.com/zeit/hyperterm/pull/723#issuecomment-248816264

package.json
leo: https://github.com/zeit/hyperterm/pull/723#issuecomment-248816264

package.json
```diff
@@ -36,51 +36,47 @@
...
-      "react/jsx-uses-react": "warn",
-      "react/jsx-uses-vars": "warn"
-    },
-    "parserOptions": {
-      "ecmaFeatures": {
-        "jsx": true
-      }
+      "react/jsx-filename-extension": 0,
+      "react/prop-types": 0,
+      "babel/new-cap": 0,
```
leo: https://github.com/zeit/hyperterm/pull/723#issuecomment-248816264

package.json
leo: https://github.com/zeit/hyperterm/pull/723#issuecomment-248816264

package.json
```diff
@@ -36,51 +36,47 @@
...
-        }
-      ],
-      "react/jsx-uses-react": "warn",
-      "react/jsx-uses-vars": "warn"
-    },
-    "parserOptions": {
-      "ecmaFeatures": {
-        "jsx": true
-      }
+      "react/jsx-filename-extension": 0,
```
leo: 👌

package.json
```diff
@@ -36,51 +36,47 @@
...
+      "react/prop-types": 0,
+      "babel/new-cap": 0,
+      "quote-props": 0,
+      "import/no-extraneous-dependencies": 0,
+      "no-warning-comments": 0,
+      "complexity": 0,
+      "react/no-danger": 0,
+      "react/no-string-refs": 0,
+      "react/jsx-key": 0,
+      "no-nested-ternary": 0,
```
leo: 👌

sindresorhus:
Ah, makes sense. I should have looked at the issue tracker before commenting.

package.json
ekmartin: The places those are used they're used because it's the only option though. I think this is a pretty good take on the issue: https://github.com/yannickcr/eslint-plugin-react/pull/636

Issue #731, opened 2016-09-21
Tabs not responding to click


davegomez:
I just clone the repository, ran the project, and created the binary for Mac. And notice the tabs are not responding to the click events, so is not possible to navigate between tabs using the mouse.

This is happening on the Master branch.

matheuss:
Fixed via #732 🔥

PR #732, opened 2016-09-21
Fix tab navigation using mouse clicks

davegomez:
This PR fixes the issue #731 where the user is unable to navigate through the opened tabs.


matheuss:
Thank you @davegomez! 😄 I knew I was introducing at least one bug with #723 😅 cc @leo

PR #736, opened 2016-09-22
On Linux/Windows, hide OSX specific menu items and change 'Shell' to 'File'

### LINKED ISSUE #122: Application > About does nothing
Ubuntu 16.04

flybayer:
This PR:

1. Refactors `app/menu.js` so each menu is a separate object. 
    - This makes it easier to understand each separate menu and rearrange them if needed.
    - It's also easier to conditionally add entire menus based on platform type
2. Hides all OSX specific items on Linux and Windows. Based on [this Electron document](https://github.com/electron/electron/blob/master/docs/api/menu.md#user-content-notes-on-os-x-application-menu).
    - Mainly, this hides the first 'Application' menu.
3. Renames the `Shell` menu as `File` on Linux and Windows for a more native feel. 
4. Adds the `Preferences` menu item to `Edit` on Linux and Windows for a more native feel and because the original `Application` menu is no longer shown on these platforms.
5. Renames the `Close Terminal Window` menu item to `Quit` on Linux and Windows because that's [the current behavior](https://github.com/zeit/hyperterm/blob/master/app/index.js#L327-L332).

Fixes #122

### DIFF
--- a/app/menu.js
+++ b/app/menu.js
@@ -8,260 +8,284 @@ const appName = app.getName();
 // https://github.com/sindresorhus/anatine/blob/master/menu.js
 
 module.exports = function createMenu({createWindow, updatePlugins}) {
-  return [
-    {
-      label: 'Application',
-      submenu: [
-        {
-          role: 'about'
-        },
-        {
-          type: 'separator'
-        },
-        {
-          label: 'Preferences...',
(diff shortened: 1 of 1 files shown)


rauchg:
Great stuff!

Issue #720, opened 2016-09-19
Option to disable tab close button


karlhorky:
Let me start by thanking everyone involved for this project. Great to have hackable web tech drive our terminal!

The tab close button introduced in #58 (original issue #40) is a nice feature, thanks for that. However, sometimes on a small terminal window with many tabs, this can result in mis-clicks, closing important tabs unintentionally.

What do you think about introducing a configuration option to disable the close button if a user wants to?

cc @johanbrook

bildepunkt:
Hey all. I'd love to tackle this one if that's alright.

matheuss:
Is more than alright @bildepunkt 😄 Go ahead 🙌

johanbrook:
We could hide the buttons on the tabs if they're small enough too :) One part of me thinks a configuration option is overkill, but I guess some users want hardcore configurability.

karlhorky:
> I guess some users want hardcore configurability

Yeah I don't want to click on a tab in the wrong place and lose some important process. Since this can potentially be quite damaging, I would suggest that a configuration option is a quite small price to pay.

bildepunkt:
> We could hide the buttons on the tabs if they're small enough

@johanbrook I was thinking the same thing. The user can still close it with the keyboard once it's in focus. Maybe that can be a separate feature request.

knownasilya:
Why not have a confirmation dialog ask if you're sure you want to close? I do this all the time... even with large tabs. The icon should always be visible so you know you're clicking it.

Tallestthomas:
I think @knownasilya has the right idea.  Maybe make it possible for someone to disable the prompt if they want?

karlhorky:
> Why not have a confirmation dialog ask if you're sure you want to close?

This would be annoying to me to have to close a dialog every time I make a mistake. It would still be penalizing the user for making the mistake of clicking in the wrong place.

johanbrook:
> It would still be penalizing the user for making the mistake of clicking in the wrong place.

Agree.

On tors 29 sep. 2016 at 09:10 Karl Horky

<
mailto:Karl Horky <notifications@github.com>
> wrote:

a, pre, code, a:link, body { word-wrap: break-word !important; }

Why not have a confirmation dialog ask if you're sure you want to close?

This would be annoying to me to have to close a dialog every time I make a mistake. It would still be penalizing the user for making the mistake of clicking in the wrong place.

—

You are receiving this because you were mentioned.

Reply to this email directly,
[...]

PR #737, opened 2016-09-23
feat: add close tab button config (#720)

bildepunkt:


knownasilya:
Not sure this is the best solution, see my comment https://github.com/zeit/hyperterm/issues/720#issuecomment-250276027

rauchg:
Probably easier to just add a line of CSS?

bildepunkt:
good call @rauchg! leave it to me to over-engineer...

bildepunkt:
@rauchg 
now that I've looked at it, I'm not sure how to do this with css. I thought this would work:
``` 
// lib/components/tab.js
styles() {
  ...
    icon: {
      ...
      display: this.props.showCloseTabBtn
        ? 'inline-block'
        : 'none',
```
but styles are not parsed by webpack before they are bundled, and seeing how there are no css files, I can't simply write a css rule.

it seems like a small case for beefing up webpack's css handling, do you have any suggestions on where to start?
thanks

knownasilya:
I think he means in your settings styles, not as a change to the project.

bildepunkt:
Thank you @knownasilya, that makes sense.

leo:
Do it via CSS; we won't introduce a config option for this. :blush:

karlhorky:
Thanks for giving this a shot @bildepunkt.

bildepunkt:
np :) it was fun

karlhorky:
@leo I'm assuming until there's a solution for #14 I'll have to create a plugin for this?

karlhorky:
Created a plugin for now: https://github.com/karlhorky/hyper-no-close

Issue #730, opened 2016-09-21
Rewrite code to make xo happy without any rules


leo:
The long-term goal is to comply with all of xo's standards, so that we can remove [these rules](https://github.com/zeit/hyperterm/blob/e30c49994f5d70ee8a91ad0c2777afbe28c5b7f9/package.json#L62). I already made the code comply to many rules, but these are left.
- [ ] react/prop-types
- [ ] react/jsx-no-bind
- [ ] new-cap
- [ ] complexity
- [x] no-warning-comments (https://github.com/zeit/hyper/commit/630f0f4504532ed20c0a706ded08d7d225be5ec4)
- [x] react/jsx-filename-extension
- [x] react/no-danger (https://github.com/zeit/hyperterm/pull/756)
- [x] react/no-string-refs (https://github.com/zeit/hyperterm/pull/766)
- [x] react/jsx-key (https://github.com/zeit/hyperterm/pull/767)
- [x] quote-props
- [x] import/no-extraneous-dependencies
- [x] no-nested-ternary

davegomez:
I can work on this!

davegomez:
The changes for this issue are massive in some cases. I'm planning to send several PRs for each rule fulfilled.

leo:
@davegomez Cool, thanks! 😘

ekmartin:
`react/prop-types`: Since we don't have any prop types from before, wouldn't it be better to add [flow](https://flowtype.org/)? It can be added incrementally as well, and we don't have to use it for anything else than props. 

`react/no-danger`: https://github.com/zeit/hyperterm/pull/723#discussion_r80324406 (the only place we use `dangerouslySetInnerHTML` it's the only way we can do it, and as the link mentions I think the name covers it more than enough by itself)

sindresorhus:
Context of why those rules should be enabled: https://github.com/zeit/hyperterm/pull/723#discussion_r79976176

sindresorhus:
> react/no-danger: #723 (comment) (the only place we use dangerouslySetInnerHTML it's the only way we can do it, and as the link mentions I think the name covers it more than enough by itself)

Then I would recommend just disabling it inline where it's used with a ESLint directive:

```js
// eslint-disable-line react/no-danger
```

davegomez:
So are we gonna add the Proptypes or don't?

And I agree with @sindresorhus that we should use the inline comment to disable the `react/no-danger` rule.

maxdeviant:
[...]

PR #739, opened 2016-09-23
Comply XO's "jsx-filename-extension" rule

davegomez:


davegomez:
Any comment @ekmartin?

ekmartin:
Yes, sorry - here it is: I'm not sure if I agree with this. Dan Abramov explains it much better than I do though: https://github.com/airbnb/javascript/pull/985#issuecomment-239145468

davegomez:
I agree with that too but we are trying to comply with XO's rules #730, is not about what we think or what we like.

ekmartin:
Of course it is - for such a large project as this there's really not that much virtue in trying to comply with rules that we don't agree with, at least in my opinion. If the rule makes sense, definitely, but if not I don't see the huge win in bending the project in a way that feels unnatural.

I don't think a lot of people will go into developing HyperTerm with an expectation of how everything should look based on them knowing about `xo`. They'll run the linter, and fix what's being complained about - regardless of whether they agree with it or not. Thus I think some bikeshedding on issues like this is warranted, and not something to be afraid of.

ekmartin:
We shouldn't disable the rule nonetheless though, if we don't want `.jsx` we should set it to `react/jsx-filename-extension: ["error", { "extensions": [".js"] }]`

davegomez:
I don't fight with linters and I don't see any problem in renaming the files, but I agree that use the rule `react/jsx-filename-extension: ["error", { "extensions": [".js"] }]` could be a solution, but I don't see any gain either way.

matheuss:
This rule was [removed](https://github.com/sindresorhus/eslint-config-xo-react/commit/00853d45598a892e487e61d561b1adb65ec55508) from `eslint-config-xo-react` 😅

davegomez:
@matheuss awesome, we should update ruleset then

matheuss:
cc @sindresorhus

ekmartin:
Closing in favor of #741.

sindresorhus:
I already commented about it here: https://github.com/zeit/hyperterm/pull/723#discussion_r79976176

PR #741, opened 2016-09-23
Update ESLint and XO rules dependencies

davegomez:
With this update we can leave the JS files containing JSX code keep the extension `.js` and remove the XO's `react/jsx-filename-extension` rule from the `package.json`.

Also the #739 PR can be ignored.


ekmartin:
Great, thanks!

PR #748, opened 2016-09-25
redux-logger TYPE emitting

ppot:
- use constant `INIT` in `init()` dispatch function
- error was due to non-dispatching action


ekmartin:
#675 😄

ppot:
@ekmartin  Oh! Il cancel it!

PR #752, opened 2016-09-27
Fix ^ for the BÉPO keyboard layout

rbelouin:
Tons of dead keys are broken when you use BÉPO keyboard layout, but I particularly miss '^' :)


ekmartin:
Hi! Would you mind trying out https://github.com/zeit/hyperterm/pull/214 and see if that resolves your issue?

rbelouin:
Hi! I didn't notice this PR, thanks! It is still a bit buggy, but as it tries to solve my problem in a more generic way, I'm gonna close mine.

PR #760, opened 2016-09-30
Replace explicit null check with isNil

ekmartin:
This replaces the explicit `!== null` checks we have in the UI reducer with basically `!== null && !== undefined`. This was breaking the changes introduced in #686, as `modifierKeys` would be overriden by `undefined`. 

Basically this reintroduces the behavior we had pre #723 (with `!= null`).

### DIFF
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -31,7 +31,7 @@ hterm.Terminal.prototype.overlaySize = function () {};
 // in an infinite copy loop
 hterm.Terminal.prototype.copySelectionToClipboard = function () {
   const text = this.getSelectionText();
-  if (text !== null && text !== '') {
+  if (text) {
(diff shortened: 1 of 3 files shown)


maxdeviant:
@ekmartin Is it still considered good practice to use `typeof value === 'undefined'` vs `value === undefined`? I always prefer the former, but I'm not sure if it's a practice that I have continued perhaps unnecessarily 😅

ekmartin:
@maxdeviant: `typeof value === 'undefined'` is good if you don't know if the variable is declared at all, i.e. if you're in an environment where you don't know if `window` exists or not (because then `window !== undefined` would throw). However, a function argument is always declared (and is `undefined` if missing) - so there wouldn't be a case where `typeof` would make a difference here.

PR #761, opened 2016-09-30
fix(meta-key): fix edge cases for meta key usage.

nicksrandall:
This PR solves some edge cases that #686 doesn't account for. It turns out that `String.fromCharCode(e.keyCode)` isn't very reliable to determining which key was pressed. On top of that, it will always incorporate the fact that the alt key was pressed so we need a manual way to determine which key was pressed when the option to use alt as meta is turned on. That way we can add the meta (esc) sequence before the key that was actually pressed. I've tested this on my Mac (I use the meta key heavily) and this is working for all of my use cases. 

Let me know if you'd like me to attach an issue to this PR as well.

### DIFF
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -1,4 +1,5 @@
 import {hterm, lib} from 'hterm-umdjs';
+import fromCharCode from './utils/key-code';
 
 const selection = require('./utils/selection');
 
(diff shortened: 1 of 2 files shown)


nicksrandall:
@rauchg @leo Will you review and give me feedback? Thanks for all your work on this! Hyperterm is awesome! 

CC: @aharris88

nicksrandall:
@aurelienbottazini, you may have some insight on this as well. P.S. Thanks for your contributions!

rauchg:
I'll wait on comments by @aurelienbottazini, otherwise LGTM!

aurelienbottazini:
@rauchg Look good to me.
Thanks for the improvement @rauchg !

leo:
Please rebase, @nicksrandall!

nicksrandall:
@leo @rauchg I rebased so you should be good to merge. Thanks!!

leo:
:blush: thx!

PR #765, opened 2016-10-02
No nested ternary

maxdeviant:
This PR removes the one instance of a nested ternary, and re-enables the xo rule to check for them, as per #730.

### DIFF
--- a/lib/components/tabs.js
+++ b/lib/components/tabs.js
@@ -12,48 +12,62 @@ export default class Tabs extends Component {
 
   template(css) {
     const {
-      tabs = [],
-      borderColor,
-      onChange,
-      onClose
+      tabs = []
     } = this.props;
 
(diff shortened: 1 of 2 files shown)


rauchg:
I really don't mind nested ternaries. Thoughts?

maxdeviant:
@rauchg It really depends on the ternary, but I thought this one wasn't difficult to understand. I was just changing it to make xo happy :smile: 

If leaving the xo rule disabled is not a big deal then we can leave it be.

sindresorhus:
I already removed this rule and will be gone in the next version soon to be released: https://github.com/sindresorhus/eslint-config-xo/commit/d3944f6d7993b9557ca1814b1e5d4c9b827d290f

maxdeviant:
@rauchg @sindresorhus I'll close this out then, if it no longer pertains.

PR #766, opened 2016-10-02
No string refs

maxdeviant:
This PR removes the one instance of a string ref, and re-enables the xo rule to check for them, as per #730.


rauchg:
Is this the only string ref we have?

maxdeviant:
@rauchg Surprisingly, yes! It's the only one `xo` caught, and I also just ran a search for `this.refs` and it was the only one.

leo:
@maxdeviant Please resolve the conflicts!

maxdeviant:
@leo Resolved :tada:

leo:
Thanks! 😘

PR #767, opened 2016-10-02
xo: react/jsx-key

maxdeviant:
This PR adds missing JSX keys and re-enables the xo rule to check for them, as per #730.


lib/components/tabs.js
```diff
@@ -26,6 +26,7 @@ export default class Tabs extends Component {
             <div className={css('title')}>{tabs[0].title}</div> :
             [
               <ul
+                key="tabs-container"
```
maxdeviant: Wasn't really sure what to name this. Should we use something more unique?

ppot: `tabs-component` at least? Since it's a component

maxdeviant:
Also, there's a second element in the array that is only added on OS X that wasn't causing an error, probably because it is only conditionally added to the array.

[tabs.js](https://github.com/maxdeviant/hyperterm/blob/eea6331ae1377b84a5b70dba86d55c710ba2b431/lib/components/tabs.js#L49-L52):
```jsx
// I imagine we would also want a key here
isMac && <div
  style={{borderColor}}
  className={css('borderShim')}
  />
```

lib/components/tabs.js
maxdeviant: It's not really a component though, just a list of `Tab` components. I think `tabs-list` might be better?

maxdeviant: `tabs-list` also goes along with the `list` class name that is being added to it :smile:

ppot: `nav-list` ? since the upper class is nav and the ul  class is list?

maxdeviant: That works. Went ahead and added that and one for `nav-border`

ppot: ok good

ekmartin:
@maxdeviant each child should have a unique key, React uses it to be able to uniquely identify children when it's rerendering. See https://facebook.github.io/react/docs/multiple-components.html#dynamic-children

leo:
@maxdeviant Please resolve the conflicts here as well 😊

maxdeviant:
@leo All done :tada:

leo:
😍

PR #768, opened 2016-10-02
Ensure modifierKeys config is set before reading

maxdeviant:
This PR ensures the `modifierKeys` config is set before trying to reading the options from it, preventing a TypeError from being thrown.


ekmartin:
`modifierKeys` is set by default, the problem is that it's being overridden by a `!== null` check - which is fixed through something like #760.

maxdeviant:
The extra checks probably don't hurt :wink:

But I'll close in favor of #760.

Issue #701, opened 2016-09-09
Certain characters broken on npm-generated trees


revelt:
Hi
I'm using Hyperterm, ZSH and `Roboto for Powerline`. While the same font works OK on iTerm2, it does not on Hyperterm — some branch corner characters look broken on any trees such this `npm install`:

![zzz](https://cloud.githubusercontent.com/assets/8344688/18381554/ba85a768-7674-11e6-9a14-d68ad41ccd4f.png)

I tried different Powerline fonts, such as `Source Code Pro`, `Meslo` and others, the same problem. It looks like it's not font's problem, but how Hyperterm renders `npm install` output trees.

I'm using Mac, ZSH shell and everything is updated to the latest versions (npm etc).

maxdeviant:
It looks like the characters are already broken by the time we get them back from RPC: [index.js:45-47](https://github.com/zeit/hyperterm/blob/master/lib/index.js#L45-L47)

![screen shot 2016-10-02 at 1 32 32 am](https://cloud.githubusercontent.com/assets/1486634/19018903/27f2af9e-8840-11e6-88ee-d46810d5b296.png)

One interesting thing I noticed is that the broken characters always seemed to appear when there were actions dispatched between two segments of session data:

<details>
![screen shot 2016-10-02 at 1 33 38 am](https://cloud.githubusercontent.com/assets/1486634/19018909/5fff9d0c-8840-11e6-99b1-da48ee284823.png)
![screen shot 2016-10-02 at 1 33 49 am](https://cloud.githubusercontent.com/assets/1486634/19018911/64f35b32-8840-11e6-9b68-7bf822386400.png)
![screen shot 2016-10-02 at 1 34 00 am](https://cloud.githubusercontent.com/assets/1486634/19018912/67d5b714-8840-11e6-913d-c9320098275a.png)
</details>

Although I can't tell if that's part of the problem or just a coincidence, because it doesn't happen at every break, as seen here:

![image](https://cloud.githubusercontent.com/assets/1486634/19018916/a8f6c77e-8840-11e6-8e8a-a7df2023b452.png)

Either way, it's happening further up the chain than during the render.

maxdeviant:
The issue appears to be the [`.toString('utf8')`](https://github.com/zeit/hyperterm/blob/master/app/session.js#L51) that is called when emitting the data.

If you write the data buffer out to a file directly, you won't see the broken characters:

```js
this.pty.stdout.on('data', data => {
  if (this.ended) {
[...]

PR #769, opened 2016-10-02
Preserve multi-byte encoded utf-8 characters

maxdeviant:
This PR tweaks the session data emitter to properly decode the data buffer using a [`StringDecoder`](https://nodejs.org/api/string_decoder.html).

>The string_decoder module provides an API for decoding Buffer objects into strings in a manner that preserves encoded multi-byte UTF-8 and UTF-16 characters. 

This fixes the issue seen in #701.


rauchg:
amazing

PR #771, opened 2016-10-04
split panes: set hotkeys based on OS

ekmartin:
Forgot to include this in the last PR. This sets the split hotkeys based on what OS you're on, so that the correct hotkeys show in the menu. The hotkeys are still the same:

macOS:
* Vertical split: `Cmd + D`
* Horizontal split: `Cmd + Shift + D`

Other:
* Vertical split: `Ctrl+Shift+E`
* Horizontal split: `Ctrl+Shift+O`

### DIFF
--- a/app/menu.js
+++ b/app/menu.js
@@ -2,6 +2,7 @@ const os = require('os');
 const path = require('path');
 const {app, shell, dialog} = require('electron');
 
+const isMac = process.platform === 'darwin';
 const appName = app.getName();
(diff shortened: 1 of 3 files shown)


fearphage:
Why the weird key combos on non-mac OSes? Any rhyme or reason?

ekmartin:
It's what Gnome Terminator uses. Definitely open to better suggestions though!

PR #779, opened 2016-10-05
Run the one tests that exists in Travis before a release

fearphage:
So this did a few things:

* changed project from `c` to `node_js` project
* defaulted to node v6
* no longer overriding `install` nor `script` (test) steps in Travis. We're now using the default inferred values -- `npm install` and `npm test` respectively.
* We only run `npm run dist` if the tests pass

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ install:
 - npm install
 - npm prune
 
-script:
+after_success:
 - npm run release
 
 branches:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,13 @@ sudo: required
 dist: trusty
 
 language: node_js
-node_js: 6
 
 matrix:
   include:
     - os: osx
+      node_js: 6
     - os: linux
+      node_js: 6
       env: CC=clang CXX=clang++ npm_config_clang=1
       compiler: clang
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ sudo: required
 dist: trusty
 
 language: node_js
-node: 6
+node_js: 6
 
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


fearphage:
I just noticed that `npm run dist` is currently a part of `npm test`. That seems unnecessary and unexpected since the goal is to call `npm run dist` when tests pass.

```shell
➜  hyperterm git:(run-tests-at-all) ✗ npm run
Lifecycle scripts included in hyperterm:
  start
    concurrently --kill-others --raw "npm run dev" "electron app"
  test
    npm run dist && xo && ava
  postinstall
    install-app-deps

available via `npm run-script`:
  dev
    webpack -w
  build
    NODE_ENV=production webpack
  prepush
    npm test
  pack
    npm run build && build --dir && babel --no-comments --compact --minified --out-file app/dist/bundle.js app/dist/bundle.js
  dist
    npm run build && build
```

ekmartin:
@fearphage: The testing setup uses [spectron](https://github.com/electron/spectron), which actually runs the application binary when testing. See https://github.com/zeit/hyper/blob/master/test/index.js. Thus you need to create the binary before the tests can be run.

fearphage:
@ekmartin That makes sense. At a minimum, there's no need to call `npm run dist` after tests pass since it was already called before the tests ran.

rauchg:
Also, if you read some of my commits before the 0.8.0 / 0.8.1 releases, I struggled with some tests failing in travis :|

rauchg:
(I think it was linux)

fearphage:
> I struggled with some tests failing in travis

What test**s** do you speak of? :eyes: This one? There was a problem with this test failing in Linux, but it was resolved a few commits back. `npm run build` and many other things were failing in Linux for a while. This should at least notify someone when that is the case next time.

fearphage:
@rauchg @ekmartin It's reviewed and tests are green. What are the next steps?

rauchg:
This is wonderful @fearphage

fearphage:
@rauchg Thanks. I'm happy to help!

Issue #666, opened 2016-08-27
Faster full screen.


jordwalke:
I enjoy the very fast (no animation) full screen behavior in iTerm/MacVim etc. I think hyperterm should have an option to allow full-screening very quickly.

beane:
Agreed! I think the default iterm command is `Cmd`+`Enter`.

PR #803, opened 2016-10-07
Quick full screen

maxdeviant:
This PR is for investigating/adding the quick full screen functionality that @jordwalke requested in #666 (I guess I'm doing the devil's work here 😱 ).

Currently I have it working quickly by just expanding the window to take up all available space on the screen. It feels a little hacky to me, but it is fast :smile:

#### TODO:
- [x] Enter quick full screen mode
- [x] Leave quick full screen mode
  - [x] Restore the window to the previous size/location it was in before entering full screen
- [x] Toggle the menu item appropriately
- [x] Add accelerators for Linux
- [x] ~~Figure out if there is a more elegant way to do this~~
  - I tried the simple stuff first (i.e., using [`BrowserWindow.setFullScreen`](http://electron.atom.io/docs/api/browser-window/#winsetfullscreenflag), but that still uses the OS X animations. 
  - **EDIT:** After making some tweaks and sleeping on it, I think that my current implementation is probably the best way to go about it, given the constraints of Electron.


jordwalke:
I hope that other electron apps can learn from your approach.

app/menu.js
```diff
@@ -312,6 +312,15 @@ module.exports = function createMenu({createWindow, updatePlugins}) {
       },
       {
         role: 'togglefullscreen'
+      },
+      {
+        label: 'Enter Quick Full Screen',
+        accelerator: 'Cmd+Enter',
```
jordwalke: Nice choice! (Is it configurable?)

app/index.js
```diff
@@ -245,6 +246,16 @@ app.on('ready', () => installDevExtensions(isDev).then(() => {
       win.maximize();
     });
 
+    rpc.on('enter quick full screen', () => {
+      const {width, height} = electron.screen.getPrimaryDisplay().workAreaSize;
+      win.setSize(width, height);
```
jordwalke: Does this render on top of the menu bar?

maxdeviant: Not currently. I was going to ask what you had in mind for the behavior.

maxdeviant:
@jordwalke If it works well I can see about abstracting it into an Electron plugin so other apps can add the same functionality easily 👍

maxdeviant:
Here it is in action: https://gfycat.com/BitterScentedBarnowl

app/menu.js
```diff
@@ -312,6 +313,28 @@ module.exports = function createMenu({createWindow, updatePlugins}) {
...
+        click(item, focusedWindow) {
+          if (focusedWindow) {
+            focusedWindow.rpc.emit('enter quick full screen req');
+          }
+        }
+      },
+      {
+        id: 'LEAVE_QUICK_FULL_SCREEN',
+        label: 'Leave Quick Full Screen',
+        accelerator: 'Cmd+Enter',
```
flybayer: Please add an accelerator for Linux users :grinning: We don't have `Cmd`

maxdeviant: Got it!

maxdeviant: @flybayer Any thoughts on what the accelerator should be? :thinking: `Alt+Enter`?

flybayer: That may be ok, although `Ctrl+Shift+Enter` might be better. It's similar to `Cmd+Shift+F` for OSX.

rauchg:
How about ^⌘F ?

rauchg:
(not sure it's better, but seems more standard?)

maxdeviant:
@rauchg ^⌘F is already reserved for normal full screen on OSX (which triggers the animation).

shime:
Greta work, can't wait for it to be merged. Worth noting that the hotkey for entering full screen in iTerm is also `⌘↵`.

shime:
Detected a couple of issues after using it for longer time:

* it remembers the window size after exiting, so when it's restarted it will show in full screen - maybe not a problem, but it's unable to exit full screen as a result
* the event is not firing every time the hotkey is pressed, it fires after 2-3 attempts occasionally

maxdeviant:
@shime Yea, I was aware of the first issue, but I'm not totally sure what the best way to handle it is.

Because Hyper saves your terminal position and size when it exits (as part of another feature).

I was debating between:

- Simply not saving the position/size when in quick full screen
  - This would restore the position/size you had before entering full screen on the next load
- Saving the "is full screen" state to the Electron config (where the position and size are recorded)
  - This would bring you back to full screen on the next load, but you would be able to toggle it off using the hotkey

The former is easier since we just have to track the state in-memory, as opposed to the latter where it must be written out when the app exits.

cc @rauchg @jordwalke Any input on how this should work?

maxdeviant:
@shime Also, that second issue is interesting. I've only done cursory testing, so the next time I have a look at it I will see if I can reproduce and fix the issue.

jordwalke:
Me? I'm not picky about which of those two solutions you go for as long as it's not buggy. I'm just happy to have full screen!

lib/index.js
```diff
@@ -80,6 +80,14 @@ rpc.on('split request vertical', () => {
...
 });
 
+rpc.on('enter quick full screen req', () => {
+  store_.dispatch(uiActions.enterQuickFullScreen());
+});
+
+rpc.on('leave quick full screen req', () => {
+  store_.dispatch(uiActions.leaveQuickFullScreen());
+});
+
```
rauchg: I like this "roundtrip" approach. Gives us lots of flexibility

jordwalke:
Did I miss why this was reverted?

maxdeviant:
@jordwalke There was an issue with it where sometimes you would get locked in quick full screen and not be able to get out. Usually it happened if you switched back and forth too fast.

I've been meaning to have a look at it, but haven't had any time to devote to it lately :slightly_frowning_face:

hojberg:
damn i was excited about this

ChadTaljaardt:
Bump for this, it is a must have feature for productivity in my opinion.

asantos00:
Any plans about the `Cmd+enter` action to fullscreen on mac?

albinekb:
you can do that with the custom keymaps @asantos00

asantos00:
So it will not be added as default behaviour? I think it is default behaviour for all OSX apps :(

PR #804, opened 2016-10-07
Move notification fetching to the main process

ekmartin:
Right now the new notification fetching is done in each window, this makes sure it's done in the main process instead.

I've also added `npm run start-app`, so that it's possible to start the electron app without needing `./node_modules/.bin` in your path.

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,7 +103,8 @@
     }
   },
   "scripts": {
-    "start": "concurrently --kill-others --raw \"npm run dev\" \"electron app\"",
+    "start": "concurrently --kill-others --raw \"npm run dev\" \"npm run app\"",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -102,7 +102,8 @@
     }
   },
   "scripts": {
-    "start": "concurrently --kill-others --raw \"npm run dev\" \"electron app\"",
+    "start": "concurrently --kill-others --raw \"npm run dev\" \"npm run start-app\"",
+    "start-app": "electron app",
```
flybayer: What if this was just `app` :point_right: `npm run app` ?

rauchg:
This is absolutely awesome

PR #809, opened 2016-10-07
Change cask install to hyper

DaleJefferson:
It looks like the cask package name has changed in the renaming

```
brew cask install hyperterm
Error: No available Cask for hyperterm
Error: nothing to install
```


leo:
😊 thanks!

PR #812, opened 2016-10-07
Cleanup (rename HyperTerm to Hyper)

danielbayerlein:
* Rename HyperTerm connector and extension
* Rename container
* Add HYPER_DEBUG


rauchg:
We should still have backwards compat with older plugins that use `decorateHyperTerm` imho

danielbayerlein:
@rauchg Should I revert 3568b57144edcd634fdf81f48fbb3eebc566d9f2 or should we offer `decorateHyperTerm` and `decorateHyper`?

rauchg:
@danielbayerlein we should call everything hyper like you did, but if a plugin defined `decorateHyperTerm`, we treat it as `decorateHyper` and `console.error` with a deprecation message.

Thanks for this work btw, super welcome

danielbayerlein:
@rauchg Thank you for your feedback! Is this your favorite solution? d4560000df128631748b514dbda95403fe4c8407

rauchg:
🎉

PR #818, opened 2016-10-07
Fix copy/paste in vim/tmux

### LINKED ISSUE #773: Copy/paste broken in Vim/Tmux + Text selection broken in tmux
Clicking and dragging on text in tmux, such as a directory listing, does nothing.

Copy & paste also does nothing in tmux. I can paste all day long in a non-tmux pane but never in a tmux pane.

This is a show-stopper for me :sob: 

_I'm on Ubuntu 14.04_

flybayer:
This partially reverts commit a4315042fa3266d42b67e74f9ac3403ab7e3ca16.

Fixes #773, #789

/cc @ekmartin

### DIFF
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -207,25 +207,6 @@ hterm.Terminal.prototype.onMouse_ = function (e) {
   return oldOnMouse.call(this, e);
 };
 
-// sine above we're no longer relying on `preventDefault`
-// to avoid selections, we use css instead, so that
-// focus is not lost, but selections are still not possible
-// when the appropiate VT mode is set
-hterm.VT.prototype.__defineSetter__('mouseReport', function (val) {
-  this.mouseReport_ = val;
-  const rowNodes = this.terminal.scrollPort_.rowNodes_;
-  if (rowNodes) {
-    if (val === this.MOUSE_REPORT_DISABLED) {
-      rowNodes.style.webkitUserSelect = 'text';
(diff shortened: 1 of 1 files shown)


rauchg:
My question is: `-webkit-user-select` is producing this side effect, or was my logic for deciding when to make text unselectable wrong?

flybayer:
@rauchg, I think it's the WebKit prefix. Moreover, it breaks pasting, not just text selection, which I don't understand 🤔 I'm limited on the debugging I can do on my daily Linux machine because my dev tools are broken.

rauchg:
The problem is that I took out the `preventDefault` that made us lose focus when performing terminal clicks, but in the process I lost the functionality of making text unselectable when doing so. This was an attempt to restore that. I was unaware of these side effects :|

rauchg:
I do agree with you that if we don't figure out something really soon, I'm down for sacrificing the text selection quirk when clicking and restoring copying and pasting :D

pascalw:
I can confirm that this indeed fixes both being able to paste in vim as well as mouse support in tmux.

dpo:
Any chance that this will be merged soon and included in a release?

leo:
Please resolve the conflicts! Will merge it 😊

leo:
Done in https://github.com/zeit/hyper/commit/f96d02480df9f8fd6116805b68ba1afbead459e9

PR #821, opened 2016-10-07
Fix theme colors

### LINKED ISSUE #819: Text pasted has white background on white
When I paste any text in to Hyper.app the pasted text shows white background on white:

screen shot 2016-10-07 at 1 38 37 pm

I have no plugins.

dotcypress:
Fix #819


rauchg:
We should fix the paste instead. 

The problem with this is that if we set a background to the terminal, it'll create a "shade" if we're using transparency in the background (basically it'll be "transparent bg on top of transparent bg".

The goal is that we only set one global background color: the window's one.

rauchg:
Thanks for the patch! Let me know if you have any luck with fixing the hterm paste instead.

dotcypress:
This patch also fix autocomplete issue.

@rauchg 

image

Issue #87, opened 2016-07-15
Lower Window Opacity? 


khayyamsaleem:
I usually use a transparent terminal. Is something like that possible for hyperterm? I know that CSS allows for lowering opacity, but I'm not sure where I would put that property.

michaelmior:
This discussion might be helpful [electron/electron#381](https://github.com/electron/electron/issues/381).

marcbachmann:
Looks like transparency is working on mac using hex colors with alpha transparency:
`#[alpha transparency code][hex color code]`
https://gist.github.com/marcbachmann/18df2187b3d1693a43ee60ccbfeab096#file-hyperterm-js-L16

<img width="578" alt="screen shot 2016-07-16 at 9 13 50 pm" src="https://cloud.githubusercontent.com/assets/431376/16896676/45837d5a-4b9a-11e6-9b4a-d57926cf33ee.png">

hhsnopek:
@marcbachmann your config doesn't produce the same results locally on my machine

bjankord:
I've set `backgroundColor: '#DD0a1a1f',` in my config. When I launch hyperterm on a fresh start, the transparency doesn't work, but if I do a full refresh after opening it, the transparency works.

jaydson:
Anyone manage to get transparent background on Linux?

leo:
Works with latest release!

c0b41:
@leo Linux not working   `--enable-transparent-visuals --disable-gpu`   Ref: [Electron Docs](https://github.com/electron/electron/blob/master/docs/api/frameless-window.md#limitations)

jacksonSingleton:
M
Few faffvV x c

fJackson Singleton
D.C. Did

alexmarmon:
rgba colors are working with 0.8.1 on MacOS Sierra:

<img width="884" alt="screen shot 2016-10-10 at 3 17 15 pm" src="https://cloud.githubusercontent.com/assets/17775646/19252961/b844c20e-8efc-11e6-877d-5829b0669fb5.png">

khayyamsaleem:
The transparency works, but there's occasionally ghosting from the last text that was in the terminal. Resizing the window fixes the problem, but I don't know what kind of bug this is.

NekR:
`rgba` works fine on El Capitan too for me, but only when I don't use any themes. When I use themes they override the background.

LeoAJ:
Is there any way to make background transparent along with the theme ? Thanks.

[...]

PR #842, opened 2016-10-10
Fix linux transparency

c0b41:
https://github.com/zeit/hyper/issues/87#issuecomment-252330644

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -17,6 +17,8 @@ const createRPC = require('./rpc');
 const notify = require('./notify');
 
 app.commandLine.appendSwitch('js-flags', '--harmony');
+app.commandLine.appendSwitch('enable-transparent-visuals');
+app.commandLine.appendSwitch('disable-gpu');
 
 // set up config
 const config = require('./config');


app/index.js
```diff
@@ -17,6 +17,8 @@ const createRPC = require('./rpc');
 const notify = require('./notify');
 
 app.commandLine.appendSwitch('js-flags', '--harmony');
+app.commandLine.appendSwitch('enable-transparent-visuals');
+app.commandLine.appendSwitch('disable-gpu');
```
ekmartin: Doesn't this disable GPU hardware acceleration?

ekmartin: As in, "isn't GPU hardware acceleration a good thing to have for performance?".

c0b41: @rauchg i don't know seems to be need electron for transparent background.. ref: [Limitations](https://github.com/electron/electron/blob/master/docs/api/frameless-window.md#limitations)

rauchg: @c0b41 can we make the flag linux only for now ? `process.platform`

c0b41:
@rauchg now?

StrictlySkyler:
+1 Please get this in.  Love to use the app; need the transparency.

rauchg:
Have we established that this doesn't downgrade performance?

StrictlySkyler:
Might objectively be slower, but not in any way that matters, I think.

If you want to future proof, "opting in" might be a good way to do it by enabling a flag in config; or perhaps allowing an array of flags to pass on startup, in which case linux users could opt-in as they desire.

However it's done doesn't so much matter to me; I just want transparency in Linux.  Until then I'm stuck with AltYo; this one's a dealbreaker.

rauchg:
If someone can run a few tests to confirm that the performance degradation is not significant, I'll merge

StrictlySkyler:
I've tested performance, and saw the following:

- Performance degradation from transparency is not noticeable
- Transparency doesn't work properly

I've attached a gif for an example:
anim

Seems that, at least by implementing the flags in this PR, transparency is complete and not adjustable, which is just as unusable.

StrictlySkyler:
@rauchg I think a userland fix for this, such as through a plugin, makes more sense, actually.  As Electron continues to evolve and grow, it may very well expose other flags.

By exposing `app` for plugins to consume, it would let the users "opt-in" to each flag they might desire to use, and (hopefully) reduces the amount of updates the Hyper team would need to reactively implement.

Regarding transparency for linux specifically, it seems that it's related to a bug in Chromium; see [this](https://github.com/electron/electron/issues/2170) and [this](https://github.com/electron/electron/issues/381).

samkelly:
this is no longer fixed -- users must now use 1.14.x as the newest version no longer works

riatzukiza:
@sam0x17 1.14.x? npm only has 1.4.x for electron, amd I mis understanding?

sam0x17:
Sorry for the confusion -- meant to say 1.4.14

Issue #799, opened 2016-10-06
Simple install for Linux? (Download link for Linux not working)


runvnc:
Hello, I clicked on the Linux download and it says 'version not found 'latest'.

Why isn't there something like `brew install` to install this on linux?  Would be some nice if there was an `apt-get install hyper` for Ubuntu or an `npm i -g hyper.app` or maybe this appimage thing could be useful (I could not find an appimage).

I mean I don't really mind cloning the repo and installing everything manually but it should be easy to install.  Unless you don't like people who use Linux in which case you should continue making it hard :P Just kidding.

BTW this such a cool project.

pveyes:
You can always use this -> https://github.com/zeit/hyper/releases, but I agree, `apt-get` would be awesome

saule1508:
The doc is not clear, it took me a while to realize it is possible to install it on Fedora. There are some other - closed - issues that mention that but it should really be on the main download page.

On Fedora there was a missing depencency, 

/tmp/.mount_aytrIO/usr/bin/hyper: error while loading shared libraries: libXss.so.1: cannot open shared object file: No such file or directory

which got resolved by installing libXScrnSaver.

so 
- download the AppImage file from the link above
- dnf install libXscrnSaver
- chmod +x the AppImage file and run it

I am glad I can run it on my Fedora, it is really great

probonopd:
In this case it looks like the AppImage is missing libraries that should be bundled inside @saule1508. That is a bug of the Hyper AppImage.

An AppImage should contain all libraries except the [blacklisted ones](https://github.com/probonopd/AppImages/blob/master/excludelist) which are expected to be part of every target system.

programmador:
I've installed 0.8.0 deb package from releases page and found some packaging issues:
1. It writes "HyperTerm is now Hyper.app. Please re-download" though it's the latest deb package.
2. It's being installed to /opt and does not include any symlinks in /usr/bin or similar dir, so it needs to add a new PATH to .bashrc.
3. It does not include system menu icon, so it's able to run only either from other terminal or from custom launch icon.

doug-wade:
I found a reasonable looking [askubuntu](http://askubuntu.com/questions/71510/how-do-i-create-a-ppa) on how to create a ppa -- would it be okay if I took a crack at setting one up?

fearphage:
FYI, I've seen people host PPA's using github pages. 

[...]

PR #851, opened 2016-10-11
Build 32 and 64 bit Linux builds

fearphage:
I can't test this, because I can't get this to run locally due to a host of failures and errors during the build process. 

Maybe you can?


fearphage:
The goal was to fix #799. I'm not sure what the next steps are here.

rauchg:
This is a very good call

fearphage:
@rauchg Does it do whatever it's supposed to?

rauchg:
@fearphage let's see!

Issue #49, opened 2016-07-05
Improve terminal process name / title extraction


rauchg:
Our current method is kinda flawed :\

cseelus:
Yeah … when using NeoVim (or Vim) (both installed via Homebrew) for example, it constantly jumps between `bash`, `Python`, `Ruby` and `nvim`.

trashhalo:
@rauchg how are you extracting the process name? I have been trying to figure out how how to override it.

lkzhao:
@rauchg 
This plugin done by @parro-it seems pretty good at getting the title. 
### [hyperterm-title](https://github.com/parro-it/hyperterm-title/blob/master/index.js)
It grabs the title used by hterm and put it as the tab title. 
Can you take a look? Maybe this should be the default behaviour.

parro-it:
@rauchg: if you'll decide to use my plugin way to set the title as the default, I will be happy to help with a PR

rauchg:
@parro-it my understanding was that only a few programs use the escape code to set the title. Is that not the case?

rauchg:
@parro-it if that technique is sufficient (not sure why I had concluded it wasn't), I would totally love a patch that includes your approach and removes `ps` polling

parro-it:
> @parro-it my understanding was that only a few programs use the escape code to set the title. Is that not the case?

Not at all... E.g. I use zsh with [pure prompt](https://github.com/sindresorhus/pure), it set the title to the `cwd`. VI set it to the name of editing file, etc..

> @parro-it if that technique is sufficient (not sure why I had concluded it wasn't), I would totally love a patch that includes your approach and removes ps polling

I will happily do one soon... :smiley:

rauchg:
@parro-it you'll be improving the performance and UX of Hyper significantly 🙏 thank you, looking forward to it

rauchg:
Also @lkzhao thanks for bringing this to my attention

parro-it:
> @parro-it you'll be improving the performance and UX of Hyper significantly :pray: thank you, looking forward to it

You're welcome, I hope to do a good job :smile:. Process polling is so expensive in term of performance?

ppot:
@matheuss  #892 closed this i think!

matheuss:
Thank you @ppot 😄

PR #854, opened 2016-10-11
fix ps filtering for detecting current process

shime:
Instead of relying on

```
$ ps aux | grep [s]001
```

to filter out grep from ps output, let's use

```
$ ps aux | grep -v 'grep\|ps' | grep s001
```

_Examples_

```
$ ps uxac | grep [s]000
hrvoje          82052   0.5  0.0  2463080   2420 s000  Ss   10:54AM   0:01.13 bash
root            65951   0.1  0.0  2438504   1028 s000  R+   12:27PM   0:00.01 ps
hrvoje          65952   0.0  0.0  2424596    400 s000  R+   12:27PM   0:00.00 grep
```

```
$ ps uxac | grep s000 | grep -v 'grep\|ps'
hrvoje          82052   0.0  0.0  2463080   2420 s000  Ss   10:54AM   0:01.14 bash
```

### DIFF
--- a/app/session.js
+++ b/app/session.js
@@ -88,13 +88,9 @@ module.exports = class Session extends EventEmitter {
     let tty = this.pty.stdout.ttyname;
     tty = tty.replace(/^\/dev\/tty/, '');
 
-    // try to exclude grep from the results
-    // by grepping for `[s]001` instead of `s001`
-    tty = `[${tty[0]}]${tty.substr(1)}`;
-
     // TODO: limit the concurrency of how many processes we run?
     // TODO: only tested on mac
-    exec(`ps uxac | grep ${tty} | head -n 1`, (err, out) => {
+    exec(`ps uxac | grep ${tty} | grep -v 'grep\\|ps' | head -n 1`, (err, out) => {
       this.fetching = false;
       if (this.ended) {
         return;


iamstarkov:
can we not do that for windows platforms? like basic if statement around [`os.platform()`](https://nodejs.org/api/os.html#os_os_platform)

rauchg:
Tricky with the advent of bash for windows @iamstarkov

rauchg:
Question for the wisdom of the Hyper community: is this really the best way to determine what the top process for the term is? I find it preetty ugly :|

iamstarkov:
@rauchg, Sindre Sorhus or James Halliday might have relevant cross platform package. in 80% of my cases they have one

rauchg:
Yep I was mostly curious about approach (i.e: polling `ps`)

shime:
Yup, definitely dirty and has issues. Noticed that the title occasionally switches between `vim` and `bash` when I'm in vim, probably due to polling and not sorting output by start time. 

Curious if there's a better way of detecting last process by tty.

chabou:
hterm's title is not relevant ? discussed here #49

app/session.js
```diff
@@ -88,13 +88,9 @@ module.exports = class Session extends EventEmitter {
...
     tty = tty.replace(/^\/dev\/tty/, '');
 
-    // try to exclude grep from the results
-    // by grepping for `[s]001` instead of `s001`
-    tty = `[${tty[0]}]${tty.substr(1)}`;
-
     // TODO: limit the concurrency of how many processes we run?
     // TODO: only tested on mac
-    exec(`ps uxac | grep ${tty} | head -n 1`, (err, out) => {
+    exec(`ps uxac | grep ${tty} | grep -v 'grep\\|ps' | head -n 1`, (err, out) => {
```
rauchg: 🤔

ppot:
@matheuss , @rauchg  This can be closed since #892 landed.

PR #859, opened 2016-10-11
Fix xo 0.17 errors and use root xo config for app

ekmartin:
I've ignored `import/prefer-default-export` - as we have a lot of plural filenames that currently only export one function (basically most of `./lib/utils`). We could change these to default exports, but I don't think it makes it any clearer (rather the opposite).


app/package.json
```diff
@@ -9,6 +9,7 @@
   },
   "repository": "zeit/hyper",
   "description": "HTML/JS/CSS Terminal",
+  "xo": false,
```
ekmartin: This makes `xo` look upwards in the directory hierarchy for the next `package.json`, see https://github.com/sindresorhus/xo/pull/151.

lib/actions/sessions.js
```diff
@@ -88,7 +88,7 @@ function createExitAction(type) {
         }
 
         const sessions = keys(getState().sessions.sessions);
-        if (!sessions.length) {
+        if (sessions.length === 0) {
```
ekmartin: :(

lib/components/term.js
```diff
@@ -299,11 +299,11 @@ export default class Term extends Component {
             height: '100%'
           }}
           /> :
-        <div
-          className={css('scrollbarShim')}
-          onMouseEnter={this.handleScrollEnter}
-          onMouseLeave={this.handleScrollLeave}
-          />
+            <div
```
ekmartin: I had to indent the second part of the ternary here, not sure if this is intended behaviour or a bug. @sindresorhus?

app/package.json
sindresorhus: I would recommend placing non-standard package.json properties at the bottom.

app/plugins.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/no-dynamic-require */
```
sindresorhus: This would be better placed inline with the code it applies to. `eslint-disable-line` can be used or `eslint-disable-next-line`.

lib/actions/sessions.js
sindresorhus: Readable and consistent code ;)

```js
sessions.length === 0
sessions.length > 3
sessions.length === 5
!sessions.length 👺
```

lib/components/term.js
sindresorhus: What was the error?

package.json
```diff
@@ -63,13 +63,14 @@
...
       "no-eq-null": 0,
-      "react/prop-types": 0,
-      "babel/new-cap": 0,
+      "new-cap": 0,
       "quote-props": 0,
-      "import/no-extraneous-dependencies": 0,
       "no-warning-comments": 0,
       "complexity": 0,
       "no-nested-ternary": 0,
+      "import/no-extraneous-dependencies": 0,
```
sindresorhus: You can remove this. Not needed anymore as XO handles `electron`.

package.json
```diff
@@ -63,13 +63,14 @@
...
       "eqeqeq": ["error", "allow-null"],
       "no-eq-null": 0,
-      "react/prop-types": 0,
-      "babel/new-cap": 0,
+      "new-cap": 0,
       "quote-props": 0,
-      "import/no-extraneous-dependencies": 0,
       "no-warning-comments": 0,
       "complexity": 0,
       "no-nested-ternary": 0,
```
sindresorhus: Was removed from XO. So can be removed here too.

package.json
```diff
@@ -63,13 +63,14 @@
     "rules": {
       "eqeqeq": ["error", "allow-null"],
       "no-eq-null": 0,
-      "react/prop-types": 0,
-      "babel/new-cap": 0,
+      "new-cap": 0,
       "quote-props": 0,
```
sindresorhus: Was changed to what you want, so can be removed and enabled again.

test/index.js
```diff
@@ -27,11 +27,11 @@ test.before(async () => {
     path: pathToBinary
   });
 
-  await app.start();
+  return app.start();
```
sindresorhus: Would be better to use a ESLint ignore comment until it's fixed: https://github.com/avajs/eslint-plugin-ava/issues/147 I would also add a `TODO` comment about removing the ignore when fixed in eslint-plugin-ava.

sindresorhus:
> We could change these to default exports, but I don't think it makes it any clearer (rather the opposite).

But does it makes it any worse? I don't think so. I feel it's generally better to use default exports when possible. One thing. One concern.

package.json
ekmartin: Thanks, found out that we missed another dependency in app as well... 👍

app/plugins.js
ekmartin: Changed it. Didn't know about `eslint-disable-next-line` - thanks.

lib/components/term.js
ekmartin: image

test/index.js
sindresorhus: I don't see how using async/await here is over the top. It works pretty much the same internally (async functions return a promise), but is more readable and easier to work with. What if you suddenly want to do something after `app.start()`?

sindresorhus:
For reference, the XO release notes: https://github.com/sindresorhus/xo/releases/tag/v0.17.0

lib/actions/sessions.js
ekmartin: Still, I would usually write `if (!something)` instead of `if (something === false)` - even if I only wanted to compare it to `false` (not `null` etc.). If it was up to me an empty array would be falsey as well, and I'd have just written `if (!sessions)`.

lib/components/split-pane.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable quote-props */
```
sindresorhus: Just curious, why?

lib/components/term.js
sindresorhus: Hmm, maybe related to https://github.com/yannickcr/eslint-plugin-react/issues/901

That being said, the code there is not very readable. I would personally have put them in two separate variables and used the variables in the ternary instead.

lib/components/split-pane.js
ekmartin: [divider_horizontal](https://github.com/zeit/hyper/blob/master/lib/components/split-pane.js#L159) and family aren't camel case identifiers. Preferably I would've changed it to be `divider-horizontal` instead, but it would be a bit rash as someone might be using the underscore separated identifiers in a plugin or similar.

The reason it's not `dividerHorizontal` is this: https://github.com/zeit/hyper/blob/master/lib/components/split-pane.js#L86 (could of course capitalize though)

ekmartin:
Thank you for the feedback. 

In regards to the default exports issue I completely agree if the filename matches the intent of the export, however in these cases it really doesn't. Take [this](https://github.com/zeit/hyper/blob/master/lib/utils/array.js#L1) as an example. If this should've been a default export I think the file should've been called `last` as well (like [this one](https://github.com/zeit/hyper/blob/master/lib/utils/url-command.js#L5)), to at least make it a little less confusing. I don't think that makes it any clearer, or easier to use though.

lib/components/term.js
ekmartin: Ye, I would've probably used a couple of extra stateless components or a render function as well - should take a look at it some time. Thanks for the link.

app/index.js
```diff
@@ -410,7 +410,7 @@ function installDevExtensions(isDev) {
   if (!isDev) {
     return Promise.resolve();
   }
-  const installer = require('electron-devtools-installer'); // eslint-disable-line global-require
+  const installer = require('electron-devtools-installer'); // eslint-disable-line import/no-extraneous-dependencies
```
sindresorhus: Why is this one being ignored?

lib/actions/sessions.js
sindresorhus: `!something` and `!something.length` is not the same thing. The first one is a lot clearer than the latter.

sindresorhus:
I would have called the file `array-last.js`.

app/index.js
ekmartin: Hm. Normally when we have dependencies that are used in both the main process and the renderer we duplicate the dependencies to both `package.json` files. `electron-devtools-installer` is only used in dev however, and is only imported in the `app` itself. The problem is that `install-app-deps` from electron-builder doesn't install devDependencies, so we can't have it there. Not sure what the best solution would be.

package.json
jfmengels: Yay, it's useful! :D

ekmartin: It definitely is @jfmengels - great work :)

app/index.js
jfmengels: It could be a slightly better idea to, instead of turning it off, changing the config of the rule at that position, to turn on the `devDependencies` option. That way, if someone removes it from the project while it's still being used, you'd still get the error here.

It's a bit akin to nitpicking as it's a very edge case probably, but I think it's a worthwhile change.

test/index.js
jfmengels: eslint-plugin-ava 3.1.1 was just released and should have fixed the issue.

ekmartin: fixed, thanks @jfmengels  :)

Issue #797, opened 2016-10-06
Setting `background: rgb(0,0,0)` on <webview> makes some websites unreadable


colepeters:
First off: love the new Hyper.app. This project is going to some incredible places, and everyone working on it deserves a big pat on the back.

After upgrading today and trying out the new webview (opening a web address by entering the URL as a command), I noticed that the `<webview>` component sets a background colour of `rgb(0,0,0)`, which makes some websites unreadable. It looks like this is due to those websites not providing a background colour property (going on the presumption that pretty much every browser defaults to a white background).

I first noticed this on my website, and then tried a couple other sites to make sure it wasn’t just an issue on my site alone. (Shown below: [my website](https://colepeters.com) and [jxnblk.com](http://jxnblk.com).) After noticing the bug on my site, I pushed an update to explicitly set a white background, but it looks like Hyper has cached my page and I’m not able to dump the cache, as my site still renders without a background colour.

![screen shot 2016-10-06 at 21 57 23](https://cloud.githubusercontent.com/assets/1713932/19170636/2ce585e2-8c11-11e6-9dda-8dc31673b25e.png)

![screen shot 2016-10-06 at 21 57 34](https://cloud.githubusercontent.com/assets/1713932/19170637/2cf78602-8c11-11e6-8167-9312d736fe07.png)

I tried something quick by opening the dev tools within Hyper and changing the `<webview>` component to have a white background instead of a black one, and this seems to solve the problem:

![screen shot 2016-10-06 at 22 01 56](https://cloud.githubusercontent.com/assets/1713932/19170635/2ce3d06c-8c11-11e6-918a-639ef7fa8cec.png)

If this is only change required, and if it makes sense to the maintainers, I’d be happy to open a PR to make this change, but wanted to open this issue first to check if setting a black background was intentional, and if this issue was already known.

Thanks in advance!

colepeters:
I’ve opened #862 to address this issue.

PR #862, opened 2016-10-11
Use white background on webviews

colepeters:
## Description
Most web clients apply a default white background to web pages, so some websites omit explicitly setting their page backgrounds to white. The default black background on the webview component was making some sites unreadable. This commit just changes it from black to white.

## Notes
I had to do a `git push --no-verify` as there are some failing tests (linting) on `origin/master`. I ran the tests before pushing, though, and it doesn't look like my changes have added to the failures.

**Fixes my previously-raised issue #797.**


ekmartin:
Thanks 👍!

PR #870, opened 2016-10-12
Change default cursor opacity. Fix #869

### LINKED ISSUE #869: The initial cursor opacity in the default config should be `.8`
Not `.75`

chabou:


rauchg:
Yaaaaaaaaaay

PR #879, opened 2016-10-14
Add PropTypes validations to make XO happy

davegomez:
This PR includes PropTypes validation so we can remove the XO's `react/prop-types` from the `package.json` file as described in the issue #730. 

**Note:** Please do a careful review of this PR since I could made a mistake figuring out the prop types.


rauchg:
Awesome. Should we wrap them with if checks so they're not included in the production bundle?

davegomez:
@rauchg just to leave them out the bundle code? Because the PropTypes are only checked on development.

rauchg:
Yep. Less code in prod = better

rauchg:
Might require some webpack wizardry. I believe the constant is __DEV__ ?

pcalves:
Instead of doing it by hand, I'd recommend [oliviertassinari/babel-plugin-transform-react-remove-prop-types](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types) which does all that work for you.

davegomez:
That's exactly what I did @pcalves, thanks!

package.json
```diff
@@ -81,6 +84,9 @@
   "babel": {
     "presets": [
       "react"
+    ],
+    "plugins": [
```
pcalves: The docs for **babel-plugin-transform-react-remove-prop-types** show the following config for `.babelrc`:
```json
{
  "env": {
    "production": {
      "plugins": ["transform-react-remove-prop-types"]
    }
  }
}
```

Without adhering to that structure, won't that plugin be applied regardless of the `ENV`?

rauchg:
@pcalves very interesting! cc @bvaughn

package.json
bvaughn: @rauchg: @pcalves:

The way to use "transform-react-remove-prop-types" in this case is to also override the default mode:
```json
[
  "transform-react-remove-prop-types", {
    "mode": "wrap"
  }
]
```

This should change `propTypes` in source:
```js
class MyComponent extends Component {
  static propTypes = {
    foo: PropTypes.string
  };
}
function MyFunctionalComponent ({ bar }) {
}
MyFunctionalComponent.propTypes = {
  bar: PropTypes.string
}
```

To this:
```js
class MyComponent extends Component {
}
if (process.env.NODE_ENV !== "production") {
  MyComponent.propTypes = {
    foo: PropTypes.string
  };
}
function MyFunctionalComponent ({ bar }) {
}
if (process.env.NODE_ENV !== "production") {
  MyFunctionalComponent.propTypes = {
    bar: PropTypes.string
  };
}
```

The default mode ("remove") is good for application code. "wrap" is meant for library code. It leaves `propTypes` in place to be later removed.

davegomez:
I agree with @ekmartin that this is not the better way to handle the proptypes and is a pain to maintain. I don't like Flow either xD.

Good catch on the `isRequired` part.

sindresorhus:
// @thejameskyle

thejameskyle:
> and no further development has been planned for the latter in part because of this.

@ekmartin PropTypes is not going to go away, but we do use Flow extensively for this purpose at Facebook

Flow is far more useful for validation as it catches way more and does so statically, meaning that you don't need to go searching for propType invalidations

davegomez:
So what about the XO's rule?
Any plans to make it disappear from XO @sindresorhus?
Or should we get rid of this, make the change to Flow and live with the rule exception?

sindresorhus:
PropTypes are better than nothing for those that don't use Flow, right?

albinekb:
@davegomez do you want to rebase this PR? 😄

albinekb:
Closing this, feel free to re-open when it's up to date and we shall merge it right away 🎉

PR #888, opened 2016-10-16
Enforce null comparisons to make XO happy

davegomez:
This pull request intents to remove the `react/no-eq-null` XO rule according to issue #730.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,6 @@
     ],
     "rules": {
       "eqeqeq": ["error", "allow-null"],
-      "no-eq-null": 0,
       "new-cap": 0,
       "no-warning-comments": 0,
       "complexity": 0,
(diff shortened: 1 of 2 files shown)


rauchg:
I honestly love the loose comparison to both null and undefined :( Am I the only one?

leo:
I don't hate them either. But we want to make XO happy!

ekmartin:
`typeof config.prop !== 'undefined'` isn't necessary though (`config.prop !== undefined` would do the same). See my comment here: https://github.com/zeit/hyper/pull/760#issuecomment-250995605

package.json
```diff
@@ -62,7 +62,6 @@
     ],
     "rules": {
       "eqeqeq": ["error", "allow-null"],
```
ekmartin: this line can also be removed

leo: @ekmartin The whole line 64? Should I push to master?

ekmartin: Yep, xo's `eqeqeq` is good if we don't want allow null.

leo: Done!

PR #892, opened 2016-10-16
Improve tabs title  

parro-it:
As discussed in #49 
First commit add an hook on hterm instance to catch ANSI escape instructions to set terminal title.
Second commit remove process polling code to set window title. 
The commented code in this commit has to be replaced with something that listen to active tab changes, and set the window title the same as the current tab title.

 Do you know if there is an easy way to listen for active tab changes from `session` class?

Actually, the window title is always the one taken from html file.


lib/components/term.js
```diff
@@ -66,6 +66,10 @@ export default class Term extends Component {
       this.term.modifierKeys = props.modifierKeys;
       // this.term.CursorNode_ is available at this point.
       this.term.setCursorShape(props.cursorShape);
+
+      // emit onTitle event when hterm instance
+      // whants to set the title of the app
+      this.term.setWindowTitle = props.onTitle.bind(props);
```
rauchg: I think `null` is fine as a bind

app/session.js
```diff
@@ -76,46 +76,46 @@ module.exports = class Session extends EventEmitter {
...
-        if (title !== this.lastTitle) {
-          this.emit('title', title);
-          this.lastTitle = title;
-        }
-      }
-
-      if (this.subscribed) {
-        this.titlePoll = setTimeout(() => this.getTitle(), TITLE_POLL_INTERVAL);
-      }
-    });
```
rauchg: We can straight up remove

parro-it: Sure, I didn't mean to leave this part commented... but I think it could be useful to replace this code with some logic to set the window title in a way that it reflect the title of the current tab.e.g. iTerm works this way...

lib/components/term.js
parro-it: I agree, binding to props is nonsensical...

app/session.js
parro-it: I used this on macOS today (I wrote this PR working on linux). I noticed the window title is not used at all on macOS, so the comment above apply only for linux OS...

parro-it:
In latest commit, I go ahead and I removed all parts of code that seems now unused or redundant. Please double check this, I fear to had removed something useful :fearful: 

Also I removed SESSION_SET_PROCESS_TITLE action, since it appear was only emitted by the process polling code.

But, apart from my `hyperterm-title` plugin, it seems to be used also by https://github.com/patrikholcak/hypertitle and https://github.com/sdomino/hyperterm-tab-cwd. So this could be a breaking change... let me know if you prefer I revert these code cleanup commits

rauchg:
Honestly I think it's ok. If the polling is not happening anymore, nor should the action.

parro-it:
> Honestly I think it's ok. If the polling is not happening anymore, nor should the action.

Fine 🎉. Could you please restart the Travis build? Seems that the macOs build was stalled and never ran.

rauchg:
Just restarted this

Issue #225, opened 2016-07-18
JavaScript error in main process


isaacs:
Seems to be triggered by the hypterterm-1password plugin, so it might be a problem with that extension.  Figured I'd share it anyhow, since probably this shouldn't be possible anyway.

```
Uncaught Exception:
Error: watch /Users/isaacs/.node-spawn-wrap-24946-c86a459da0b6 ENOENT
    at exports._errnoException (util.js:949:11)
    at FSWatcher.start (fs.js:1433:19)
    at Object.fs.watch (fs.js:1460:11)
    at Gaze._watchDir (/Applications/HyperTerm.app/Contents/Resources/app/node_modules/gaze/lib/gaze.js:288:30)
    at /Applications/HyperTerm.app/Contents/Resources/app/node_modules/gaze/lib/gaze.js:357:10
    at iterate (/Applications/HyperTerm.app/Contents/Resources/app/node_modules/gaze/lib/helper.js:69:5)
    at /Applications/HyperTerm.app/Contents/Resources/app/node_modules/gaze/lib/helper.js:78:11
    at /Applications/HyperTerm.app/Contents/Resources/app/node_modules/gaze/lib/gaze.js:422:5
    at iterate (/Applications/HyperTerm.app/Contents/Resources/app/node_modules/gaze/lib/helper.js:69:5)
    at /Applications/HyperTerm.app/Contents/Resources/app/node_modules/gaze/lib/helper.js:78:11
```

SamVerschueren:
Getting the exact same error when I try to use `nyc`

```
$ nyc ava
```

johanbrook:
I got this after installing `hyperterm-colors` and `hyperterm-spacegray`:

<img width="532" alt="skarmavbild 2016-07-25 kl 23 21 26" src="https://cloud.githubusercontent.com/assets/437261/17117973/e894861c-52be-11e6-9fe1-6d864a26e29e.png">

This was for a first time install, couldn't repro for subsequent installs of the plugins.

xjamundx:
Yeah @SamVerschueren. I get this error right now and I'm runing running `ava` in another window (not-hyperterm).

matheuss:
Since this has been reported a lot of times, I'm going to mark it as `Priority: high`. 

Related/duplicates: #357, #401, #444, #542, #564, #680

matheuss:
Anyone have any insight on this? Can this exception be safely ignored?

isaacs:
[...]

PR #893, opened 2016-10-17
Ignore file watch errors

shama:
Fixes GH-225

This ignores all errors gaze hits when watching files. With `EventEmitter`, if you don't supply at least 1 listener for `error` events, they get thrown. See https://github.com/nodejs/node/blob/master/lib/events.js#L149

I was able to repro using these steps: https://github.com/zeit/hyper/issues/225#issuecomment-253762729 and can confirm it no longer hard errors with this patch.

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -40,6 +40,9 @@ function watch() {
         });
       }
     });
+    this.on('error', () => {
+      // Ignore file watching errors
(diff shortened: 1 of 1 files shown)


rauchg:
Can't believe we missed that. Thank you

PR #894, opened 2016-10-17
Fix glitch on term resize with some prompts

chabou:
With some prompts which use end of line, quick repeated resizes cause a glitch.
This glitch occurs when resize is dispatched during hterm redraw.

We can't know when hterm finished to rerender after a resize. So we can't make these resizes sequentially.

The only solution I found is to debounce Term onResize calls.
These calls must be debounced per uid.

Before
before

After
after


iam4x:
@chabou Have you tried with `_.throttle` to keep the UI responsive to movements? Make it a little more fluid.

chabou:
Yeah I tried but it doesn't work as well.
Because, it doesn't prevent to have resize event during hterm redraw. It reduce it, but the more we want a fluid resize, the more the race condition (and the glitch) occurs.

Technically, debounce doesn't prevent it totally but reduce time window at a minimum.

I'm not really satisfied with this solution (less fluid and ugly debounce), but I don't see another one. Keep this glitch is acceptable too 😄 

Thank you for your feedback.

ppot:
@chabou relevant for xterm?

chabou:
Too early to know that but I hope it will not be relevant... Debouncing is ugly :)

chabou:
Not relevant anymore

PR #896, opened 2016-10-18
Hide header when no tabs

dkkoval:
Before:
before
After:
after

### DIFF
--- a/lib/components/tabs.js
+++ b/lib/components/tabs.js
@@ -23,7 +23,12 @@ export default class Tabs extends Component {
       {
         tabs.length ?
           tabs.length === 1 ?
-            <div className={css('title')}>{tabs[0].title}</div> :
+            <div
+              className={css('title')}
+              style={{height: (this.props.tabs.length > 1 ? '34px' : '0px')}}
+              >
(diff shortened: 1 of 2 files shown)


CodeTheory:
I'm not sure this will work because of Mac OS. Me or someone else will test it soon.

danprince:
Already been through this once on https://github.com/zeit/hyper/pull/232. The end result was the decision to push this out to a plugin.

If this landed in the hyper codebase, then it'd need some conditionals to handle OSX, which puts the tabs directly into the title bar.

screen shot 2016-10-18 at 18 31 30

dkkoval:
OK, thanks man

danprince:
I put together [danprince/hypertile](https://github.com/danprince/hypertile) to make hyperterm more usable for tiling window managers, by removing tabs, titles and window decorations altogether. Looks like you're using a tiling window manager, so you might find it useful.

rauchg:
@danprice that's a great solution

rauchg:
That said, optimizations for the look n feel of Windows / Linux are coming. Stay tuned!

PR #906, opened 2016-10-19
Prevent Hterm to share same preferences between instances

chabou:
Hterm use a mechanism of profile for its preferences.
All Hterm instances with the same profileId share the same PreferenceManager.
If not provided, `profileId` is set to `default`.
We can set it at instantiation or with method `setProfile()`.

We need to abstract this PreferenceManager mechanism in order to style Terms separately by ourselves in a React/Redux way. Every Hterm instance must have his own `profileId`.

In addition, this PR cleans prefs access and replace direct access to `this.term.prefs_`with 'this.term.getPrefs()' public method.

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -1,6 +1,7 @@
 /* global Blob,URL,requestAnimationFrame */
 import React from 'react';
 import Color from 'color';
+import uuid from 'uuid';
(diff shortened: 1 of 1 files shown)


rauchg:
@chabou <3

PR #913, opened 2016-10-20
increase notification timeout up to 30m

iamstarkov:
fetchNotifications error can happen in two cases:
* https://hyper-news.now.sh is down
* you are offline

In both cases it doesn't hurt to wait half an hour.

PS. right now it hurts, when you are offline and developing
anything for Hyper, `console.error` logs the same error
every 10s and ruins your investigation.


iamstarkov:
yeah 🎉

PR #924, opened 2016-10-23
add onRendererWindow to availableExtensions

iamstarkov:
_PR is ready to be reviewed and be merged._

because it actually available. check [./lib/utils/plugins.js:193](https://github.com/zeit/hyper/blob/master/lib/utils/plugins.js#L193-L195)


iamstarkov:
Also i wanted to add to that list `decorateTerms`,  because it used to be there, but now its not true, which is good. At the same time its bad, because Hyper api actually says that `decorateTerms` exists and supported

rauchg:
@iamstarkov it would be really nice to update the website accordingly. Thank you again!

Issue #872, opened 2016-10-13
Flexible keymap (hotkeys / shortcut)


iamstarkov:
the important key to achieve it in flexible and reliable way is to have two things:
- list of possible events to `rpc.emit` <sup>1</sup>
- one place to bind keys to those events

<sup>1</sup> Note: `rpc` bus should be the only way to broadcast key events).

Initial array of events can be collected from [`lib/index.js#L38-L129`](https://github.com/zeit/hyper/blob/aa285a8bc662f7fae1f9f265f326c87c2de88420/lib/index.js#L38-L129): like `split request vertical` or  `split request horizontal`, etc.

Then in config we can do smth like this: 

``` js
keymap: {
 "CmdOrCtrl+Alt+Left": "prev pane req",
}
```

where keys are electron [accelerators](https://github.com/electron/electron/blob/master/docs/api/accelerator.md) and values are those events for rpc to emit/receive.

After that, we need to modify `attachKeyListeners()` [lib/containers/hyper.js#L38-L76](https://github.com/zeit/hyper/blob/master/lib/containers/hyper.js#L38-L76) to not bind hardcoded keys, but instead to pick up keys and events from keymap:

``` js
// mousetrapify needs to be more advanced, for sure, but this is enough for now
const mousetrapify = key => key.replace('CmdOrCtrl', 'mod').toLowerCase();
Object.keys(this.config.keymap).map(item => {
  const accelerator = item;
  const event = this.config.keys[key];
  keys.bind(
    mousetrapify(accelerator),
    () => {
      if (focusedWindow) {
          focusedWindow.rpc.emit(event);
      }
    }
 );
})
```

Also, Hyper needs to verify that this.config is merged object similar to`Object.assign({}, defaultConfig, ...[pluginsAndThemesConfigs], userConfig)`

_Thats it._

---

One interesting benefit of proposed approach is that _you can update menu accelerators very easy and smooth_. just needs to write a function which will lookup desired accelerator by key event relevant to that menu item. <sup>2</sup>, <sup>3</sup>
-   <sup>2</sup> because right now [keeping hotkeys in a menu relevant is a fragile mess](https://github.com/iamstarkov/hyper-panes-iterm2-hotkeys/blob/master/index.js#L27-L43)
-  <sup>3</sup> findAccelerator is pretty simple function:
  
  ``` js
  const findAccelerator = event => Object.keys(this.config)
    .map(key => { key, value: this.config[key] })
    .find({ value } => value === event)
    .key
[...]

Issue #783, opened 2016-10-05
Rich Support for Tweaking Modifier Keys


rye:
#330 references remapping the Option key and using that as Meta on OS X.
#686, which introduced support for this, resolved #330, but I think that it wasn't enough to satisfy most users, especially those like myself who use Emacs in terminal and require having full access to all the modifiers.

I'd prefer to see something more like:

``` js
"modifierKeys": {
  "cmd": "super",
  "option": "meta"
}
```

In my configuration, personally.

[`lib/hterm.js`](https://github.com/zeit/hyperterm/blob/master/lib/hterm.js) will almost certainly need to be modified, specifically the `Keyboard.onKeyDown` callback.

Some questions to raise would be:
- What style would users like to see for a `"modifierKeys"` configuration?
- Is there even anything we can do to prevent operating systems like macOS from stealing these keybindings and entering 'special' characters?

I can and would love to work on this, but I'm not very familiar with the inner workings of things and I'd need to educate myself on VT.

ferakpeter:
I've played around with the configurations for the meta keys, I would prefer your config style. However, the way cmdAsMeta is working right now, it's not usable, it overrides the hotkeys for new tab and new windows, and so forth. So Hyper loses core features in that case. And on windows and mac platforms there is no cmd button, so the setting with cmdIsMeta would be misleading. You configuration style can solve this.

So from my perspective there's only one question that should be answered in the config: do you want use ALT as meta, or do you want to not have a meta key.

curioussavage:
I like the config too, and once https://github.com/zeit/hyper/pull/925 is resolved the cmd key issue will be resolved too.

ppot:
close in #1876

PR #925, opened 2016-10-23
keymap feature

iamstarkov:
_PR needs feedback_

I want flexible keymap for Hyper #872. First approach to use single RPC to communicate all the didnt really work out well, because of two separate RPCs, one for Electron and on for Renderer windows. So there is another approach to use Menu as kind of one entry-point to bind all hotkeys.
- 995bfa0 Remove all possible to implement via Menu hotkeys
- f629cc1 Change hard coded Menu accelerators to "convention"ed `command` field, which can be used to identify what for this menu item is responsible
- 69e4fd8 Add [hyper-keymap](https://github.com/iamstarkov/hyper-keymap) as standard plugin
### 🤔 Suggestion:
- keep keymap management as a default, but still separate plugin <sub>Atom is doing it with everything</sub>
- Move hyper-keymap into hyper org <sub>should we create one?</sub>
- Find a way to to notify end users about need to have hyper-keymap in their configs. Might be possible to use [`question messageBox`](http://electron.atom.io/docs/api/dialog/#dialogshowmessageboxbrowserwindow-options-callback) with "add it" or "dont bother me" buttons.
#### 😟 Cons (temporarily):
- at the time PR is being submitted `hyper-keymap` is not working fully, due to existing key bindings declared in `lib/containers/hyper.js`. Though, this pull-request almost fixes it.
- remaining key bindings, like `move-word-{left,right}` needs to be implemented the same way as everything else. I think this is possible, because as far as it is hotkey, there is no need in passing additional arguments, so it should possible to implement. That said, i seek for help to do that.
#### 😎 Pros:
- 🎉 Configured keymap, yay! 
- 👍 Even in one place 
- 🆓 [Documentation](https://github.com/iamstarkov/hyper-keymap#usage) about all [existing hotkeys](https://github.com/iamstarkov/hyper-keymap/blob/master/src/default-keymap.js) in one place, for free!

I tried my best, to bring keymap management to Hyper and now looking for feedback and discussion. What do you think about proposed approach?

### DIFF
--- a/app/config-default.js
+++ b/app/config-default.js
@@ -81,7 +81,12 @@ module.exports = {
   //   `hyperpower`
   //   `@company/project`
   //   `project#1.0.1`
-  plugins: [],
+  //
+  // Make sure, you dont remove 'hyper-keymap' plugin,
+  // unless you know what you are doing
+  plugins: [
+    'hyper-keymap'
+  ],
 
   // in development, you can create a directory under
   // `~/.hyper_plugins/local/` and include it here
--- a/app/menu.js
+++ b/app/menu.js
@@ -22,7 +22,7 @@ module.exports = ({createWindow, updatePlugins}) => {
       },
       {
         label: 'Preferences...',
-        accelerator: 'Cmd+,',
+        command: 'show-settings',
         click(item, focusedWindow) {
           if (focusedWindow) {
             focusedWindow.rpc.emit('preferences');
@@ -64,14 +64,14 @@ module.exports = ({createWindow, updatePlugins}) => {
     submenu: [
       {
         label: 'New Window',
-        accelerator: 'CmdOrCtrl+N',
+        command: 'new-window',
         click() {
           createWindow();
         }
       },
       {
         label: 'New Tab',
-        accelerator: 'CmdOrCtrl+T',
+        command: 'new-tab',
         click(item, focusedWindow) {
           if (focusedWindow) {
             focusedWindow.rpc.emit('termgroup add req');
@@ -85,7 +85,7 @@ module.exports = ({createWindow, updatePlugins}) => {
       },
       {
         label: 'Split Vertically',
-        accelerator: isMac ? 'Cmd+D' : 'Ctrl+Shift+E',
+        command: 'split-vertical',
         click(item, focusedWindow) {
           if (focusedWindow) {
             focusedWindow.rpc.emit('split request vertical');
@@ -94,7 +94,7 @@ module.exports = ({createWindow, updatePlugins}) => {
       },
       {
         label: 'Split Horizontally',
-        accelerator: isMac ? 'Cmd+Shift+D' : 'Ctrl+Shift+O',
+        command: 'split-horizontal',
         click(item, focusedWindow) {
           if (focusedWindow) {
             focusedWindow.rpc.emit('split request horizontal');
@@ -106,7 +106,7 @@ module.exports = ({createWindow, updatePlugins}) => {
       },
       {
         label: 'Close',
-        accelerator: 'CmdOrCtrl+W',
+        command: 'close',
         click(item, focusedWindow) {
           if (focusedWindow) {
             focusedWindow.rpc.emit('termgroup close req');
@@ -115,8 +115,8 @@ module.exports = ({createWindow, updatePlugins}) => {
       },
       {
         label: isMac ? 'Close Terminal Window' : 'Quit',
-        role: 'close',
-        accelerator: 'CmdOrCtrl+Shift+W'
+        command: 'close-window',
+        role: 'close'
       }
     ]
   };
@@ -150,13 +150,51 @@ module.exports = ({createWindow, updatePlugins}) => {
       },
       {
         label: 'Clear',
-        accelerator: 'CmdOrCtrl+K',
+        command: 'clear',
         click(item, focusedWindow) {
           if (focusedWindow) {
             focusedWindow.rpc.emit('session clear req');
           }
         }
       }
+      /**
+      ,{
+        type: 'separator'
+      },
+       * @TODO: implemented in rpc, ./lib/index.js
+      ...[{
+        label: 'Move word Left',
+        command: 'move-word-left',
+      }, {
+        label: 'Move word Right',
+        command: 'move-word-right',
+      }, {
+        label: 'Delete word Left',
+        command: 'delete-word-left',
+      }, {
+        label: 'Delete word Right',
+        command: 'delete-word-right',
+      }, {
+        label: 'Delete line',
+        command: 'delete-line',
+      }, {
+        label: 'Move to line start',
+        command: 'move-to-line-start',
+      }, {
+        label: 'Move to line end',
+        command: 'move-to-line-end',
+      }, {
+        label: 'Select All',
+        command: 'select-all',
+      }].map(submenuItem => {
+        item.click = function click(item, focusedWindow) {
+          if (focusedWindow) {
+            focusedWindow.rpc.emit(submenuItem.command);
+          }
+        };
+        return item;
+      }),
+      */
     ]
   };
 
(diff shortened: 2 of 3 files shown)


flybayer:
This looks promising to me! 👍

iamstarkov:
@CodeTheory @matheuss what do you think?

ekmartin:
Cool :) Is it necessary to couple keybindings so tightly to `Menu` though? I think being able to have keybindings that aren't listed in the menu is a good thing, especially if you consider wanting to have multiple keybindings per action (which I'm not sure is even possible with accelerator).

Couldn't an alternative be to just treat it as a regular part of the config? That way you could do something like `accelerator: keyBindings['new-window']`, and still have `keys.bind(keyBindings['move-left'], moveLeft)` in `hyperterm.js`. Could still have menu on the form you've suggested though, but just transform it to explicitly use the keybindings from the config - like you've done.

Also, regarding the keymap, I think it would really help in terms of user experience if all the default keybindings are listed explicitly where you're able to change them, so that you don't have to look up the names to know how to make modifications (seeing as we don't have an insane amount of different combinations). This might make the config too large though, which is why I previously argued for having it in a separate file (and bundle everything below a `.hyper` folder).

In regards to if it should be implemented in the core or as a default plugin, I think it would definitely make more sense to have this in Hyper itself, as it wouldn't be too different from the already existing stuff that makes use of values from the config.

ChennyBaBy:
Very excited to see this being implemented! I found this PR while looking to add ketbindings to hyper. i have a suggestions for a Hyper command, I would love to be able to have a global key binding that would bring up the hyper app to the foreground in focused where ever I was, not sure how you would name that command tho.

iamstarkov:
@ChennyBaBy i believe there are several plugins with that idea in mind. check out https://github.com/bnb/awesome-hyper

rye:
This looks an awful lot like what I was looking for in #783.  A couple of suggestions:

* I'd like to make sure that the user has the option to send the actual keycode through to the shell.  For me, personally, I'd much rather have access to Emacs keybinds in my terminal like `C-w`, `M-w` (which is CmdOrControl-w or the `close` command in Hyper in my setup).
  * What this really means is the option to *disable* things altogether.  Sure, it might be lame to disable creating new tabs for most users, but I'd like to have the option for my own setup, and I'm sure many would as well.  Perhaps allow users to bind keybinds to `null`, then have the internal system simply forward keycodes on if the keybind is `null`?

If the PR can fulfill these requirements, then it's a huge win and I'd be proud to switch to Hyper full-time.  I'm stuck on iTerm because of the lack of disable-ability, if that makes any sense.

rye:
I think it might also make sense to think about generalizing everything and letting users specify which keys correspond to which keybinds&mdash;I quite like the `C-`, `M-`, and `s-` prefixes that Emacs uses for commands, and I think they'd be good for using to write keybinds.

jritsema:
Great idea @iamstarkov.  I'm interested in [pc-style word navigation inside the terminal](https://github.com/zeit/hyper/issues/1047), something like...  

```
keymap: {
  'Ctrl+Left': 'editor:move-to-beginning-of-word',
}
```

Would this PR support that?

iamstarkov:
@ekmartin 

> Cool :) Is it necessary to couple keybindings so tightly to Menu though?

Correct, Its not necessary to couple keybindings to Menu. Though, its the only working RPC-ish way to emit actions to both `Renderer` and `Electron` windows.

iamstarkov:
@ekmartin 

> I think being able to have keybindings that aren't listed in the menu is a good thing, especially if you consider wanting to have multiple keybindings per action (which I'm not sure is even possible with accelerator).

Correct, Its not possible with accelerators.

iamstarkov:
@ekmartin 

> Couldn't an alternative be to just treat it as a regular part of the config? That way you could do something like accelerator: keyBindings['new-window'], and still have keys.bind(keyBindings['move-left'], moveLeft) in hyperterm.js. Could still have menu on the form you've suggested though, but just transform it to explicitly use the keybindings from the config - like you've done.

I finally see what you are saying, thats probably gonna work out

iamstarkov:
@ekmartin 

> Also, regarding the keymap, I think it would really help in terms of user experience if all the default keybindings are listed explicitly where you're able to change them, so that you don't have to look up the names to know how to make modifications (seeing as we don't have an insane amount of different combinations). This might make the config too large though, which is why I previously argued for having it in a separate file (and bundle everything below a .hyper folder).

Its debatable, user's keymap config rarely would grow much. And default keybindings will be configured in default config in Hyper itself.

iamstarkov:
@ekmartin 

> In regards to if it should be implemented in the core or as a default plugin, I think it would definitely make more sense to have this in Hyper itself, as it wouldn't be too different from the already existing stuff that makes use of values from the config.

will copy our discussion from slack:


> 
> 1:24 AM @ekmartin
> it makes more sense to have it in core imo
> 
> 1:25 AM @CodeTheory
> I agree
> 
> 1:26 AM @iamstarkov
> ehm
> ```
> ☯ apm ls | grep key
> ├── keybinding-resolver@0.35.0
> ```
> atom did it via plugin
> core plugin for sure
> 
> 1:27 AM @CodeTheory
> Most of everything that makes atom is via a plugin
> 
> 1:27 AM @ekmartin
> ^
> 1:27 AM @CodeTheory
> Hyper is different, it should be in the core
> 
> 1:27 AM @iamstarkov
> yep
> why its that different?
> 
> 1:28 AM @ekmartin
> keybindings really isn’t that different from anything else that gets read from the config though
> 
> 1:28 AM @iamstarkov
> im not arguing against it, just want to know details, before i can say im sold with this approach
> 
> 1:29 AM
> @ekmartin and? plugins and themes utilise config just fine
> 
> 1:29 AM @ekmartin
> I think if you’d want to split things into plugins it would make more sense to rewrite some of the existing core with that in mind as well
> 
> 1:29 AM @iamstarkov
> probably
> 
> 1:29 AM @CodeTheory
> Yeah, for now I think it's best in the core since that's how Hyper's architecture is
> 
> 1:29 AM @iamstarkov
> okay
> 
> 1:29 AM @CodeTheory
> We could split things out into plugins later if we decide to go that way
> although, in my opinion, it'd be cool to see Hyper created with core plugins
> 
> 1:30 AM @ekmartin
> @matheus is more up to date on hyper than I am now though :slightly_smiling_face:
> 
> 1:30 AM @CodeTheory
> since it'd allow the user to get acquainted with plugins
> But Hyper's config is a bit hard for _everyone_ to use
> as we can see from the issues
> 
> 1:30 AM @iamstarkov
> i cant rewrite core in one pr, even if i could it would be horrible to review such big pr
> 
> 1:30 AM @CodeTheory
> No, no. We shouldn't do that anyway :stuck_out_tongue:
> 
> 1:31 AM @iamstarkov
> so i thought i'd start with one core plugin
> 
> 1:31 AM @CodeTheory
> If @rauchg or @leo were around, I'd ask them
> They're the ultimate architects
> But I do agree that core plugins would be a cool thing
> for many reasons
> 
> 1:33 AM @iamstarkov
> @CodeTheory
> > We could split things out into plugins later if we decide to go that way
> 
> thats legit point as well

megalithic:
@iamstarkov mind resolving the merge conflicts? @PSNAppz are you able to merge once conflicts are accurately resolved? this would be great and open up even more doors for hyper via https://github.com/iamstarkov/hyper-keymap#-caveats

PSNAppz:
@megalithic Yes.

iamstarkov:
heads up to @megalithic and everybody in this thread: yay, i will be working on this on this holyday

macobo:
@iamstarkov To make your life easier, I took a stab at rebasing this.

https://github.com/zeit/hyper/compare/master...macobo:iamstarkov-feat/keymap-2?expand=1

macobo:
Bump on this as without this hyper is unusable on linux - should I open a separate PR with the rebase?

iamstarkov:
@macobo can i ask you to make a pull-request to my fork with your rebase?

iamstarkov:
sorry, i have to say this: i couldnt make it and have no energy to fix this issue on my own. That said, I will close this pull-request. Another reason to close PR is this branch and code is outdated. 

Please refer to original issue #872

ppot:
Plz add request inside issue: https://github.com/ppot/hyper/issues/11

PR #928, opened 2016-10-24
increase timeout for update checks

### LINKED ISSUE #880: “API rate limit exceeded for rauchg” on download link
```
‣ curl https://hyper-updates.now.sh/download/osx
{"message":"API rate limit exceeded for rauchg.","documentation_url":"https://developer.github.com/v3/#rate-limiting"}
```

iamstarkov:
_PR is ready for code-review_

It seems like, there are lots of them and GitHub is not happy.
So increasing timeout 6x should fix that. In the end there is no need
to check updates every 5 minutes.

Fix #880

### DIFF
--- a/app/auto-updater.js
+++ b/app/auto-updater.js
@@ -19,6 +19,10 @@ function init() {
 
   autoUpdater.setFeedURL(`${FEED_URL}/${version}`);
 
+  setTimeout(() => {
+    autoUpdater.checkForUpdates();
+  }, ms('10s'));
+
   setInterval(() => {
     autoUpdater.checkForUpdates();
   }, ms('30m'));
--- a/app/auto-updater.js
+++ b/app/auto-updater.js
@@ -19,13 +19,9 @@ function init() {
 
   autoUpdater.setFeedURL(`${FEED_URL}/${version}`);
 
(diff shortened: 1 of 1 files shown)


albinekb:
isn't checking for update every 30 min a bit too often? once a day would probably be enough? (since releases are days/weeks apart) and then have a manual way to check if you really want the new version asap? 

Just a thought :)

app/auto-updater.js
```diff
@@ -19,13 +19,9 @@ function init() {
 
   autoUpdater.setFeedURL(`${FEED_URL}/${version}`);
 
-  setTimeout(() => {
```
iamstarkov: you are right

iamstarkov:
> isn't checking for update every 30 min a bit too often?

idk, it was 5m before. On the other hand, once a day, is not that good, 'because you'll get update next day

iamstarkov:
lets start with 6x increase, we can find appropriate balance in the future

albinekb:
sound fair 👍

rauchg:
30 minutes is good. Think of Hyper more as a hot-reloadable website rather than a static app ;)

PR #932, opened 2016-10-25
Make the close menu items less ambiguous

### LINKED ISSUE #123: Shell > Close is ambiguous

mike-engel:
@Tyriar brought up a good point in #123 that the close menu items can be
ambiguous and misleading. This clarifies them so you know exactly what
you're closing.

Fixes #123.

### DIFF
--- a/app/menu.js
+++ b/app/menu.js
@@ -105,7 +105,7 @@ module.exports = ({createWindow, updatePlugins}) => {
         type: 'separator'
       },
       {
-        label: 'Close',
+        label: 'Close Session',
         accelerator: 'CmdOrCtrl+W',
         click(item, focusedWindow) {
           if (focusedWindow) {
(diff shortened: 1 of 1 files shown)


app/menu.js
```diff
@@ -105,7 +105,7 @@ module.exports = ({createWindow, updatePlugins}) => {
         type: 'separator'
       },
       {
-        label: 'Close',
+        label: 'Close Tab',
```
ekmartin: This also closes split panes now. Maybe something like `Close Session` or `Close Pane` could work? Not sure.

mike-engel: Good point! Updated to `Close Session`

ekmartin:
Cool, thanks :)

PR #943, opened 2016-10-28
Change URL regex - fixes #867

### LINKED ISSUE #867: URL Parsing error: Type git@github.com:zeit/hyper.git and Hyper crashes
If you just run 

```bash
$ git@github.com:zeit/hyper.git
```

Hyper crashes. I assume this is because it is trying to parse the "command" as a URL. 

Explored a bit:

Running `git@github.com` opens GitHub.com up just fine.

Running `git@github.com:` opens GitHub.com up, but it was delayed a little bit - this could have quite possibly been my connection, though.

Running `git@github.com:zeit`crashes the terminal - this is the origin of the issue, as I don't think anything is expected after the `:`.

oliverdunk:
Hi,

This fix addresses an issue in which URLs were passed too kindly resulting in the terminal regularly being redirected to a blank screen (as the location it was being redirected to did not make sense).

I have re-written the URL regex and matching system, and this PR is ready to be merged. Some obscure IPV6 addresses are not matched which will need to be fixed in the future, but for now it is better that we miss a few URLs than occasionally breaking the terminal.

### DIFF
--- a/lib/utils/url-command.js
+++ b/lib/utils/url-command.js
@@ -1,6 +1,6 @@
 import * as regex from './url-regex';
 
-export const domainRegex = /\b((?=[a-z0-9-]{1,63}\.)(xn--)?[a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,63}\b|^localhost$|^127(?:\.[0-9]+){0,2}\.[0-9]+$|^(?:0*:)*?:?0*1$/;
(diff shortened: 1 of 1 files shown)


rauchg:
Thank you @oliverdunk. Does `localhost:3000` work with this?

oliverdunk:
@rauchg Currently not, as this is the only exception where the domain doesn't have an extension. It should be trivial to add a specific case for localhost.

PR #947, opened 2016-11-01
Remove prevUid test from setActiveSession

ppot:
### Ready to be merged

#892 removed the blur and focus on the session.
Therefore we don't need to test the `prevUid`


ppot:
@matheuss @rauchg @leo

matheuss:
Thank you @ppot 😄

PR #968, opened 2016-11-07
Fix inverted text display. Fix #883

### LINKED ISSUE #883: Foreground color makes nano menu unreadable
Whatever foreground color I set, nano seems to use it for the menu text and their background colors, making it unreadable. For example with `foregroundColor: '#f2777a'` I get:

screenshot 2016-10-15 16 29 54

flyngate:
This is similar to #821 but preserves background transparency.

hterm swaps `foreground` and `background` text attributes when handling inverted text (e.g. `echo -e "Normal \e[7minverted"`). Therefore, we can't set `background-color` as transparent as we'll get invisible text in this case.

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -37,6 +37,10 @@ export default class Term extends Component {
     prefs.set('cursor-color', this.validateColor(props.cursorColor, 'rgba(255,255,255,0.5)'));
     prefs.set('enable-clipboard-notice', false);
     prefs.set('foreground-color', props.foregroundColor);
+
+    // hterm.ScrollPort.prototype.setBackgroundColor is overriden
+    // to make hterm's background transparent. we still need to set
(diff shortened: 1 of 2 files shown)


lib/components/term.js
```diff
@@ -37,7 +37,7 @@ export default class Term extends Component {
     prefs.set('cursor-color', this.validateColor(props.cursorColor, 'rgba(255,255,255,0.5)'));
     prefs.set('enable-clipboard-notice', false);
     prefs.set('foreground-color', props.foregroundColor);
-    prefs.set('background-color', 'transparent');
+    prefs.set('background-color', props.backgroundColor);
```
rauchg: maybe we should add a comment here about the expected override in `hterm.js` for this to work

flyngate:
@rauchg the comment has been added.

PR #970, opened 2016-11-07
Update old electron target engine for dependencies (ready to be merged)

dfrankland:
This fixes dependencies of Hyper that build based on the engine being used ([`nodegit`](https://github.com/nodegit/nodegit) for example). Uses [electron’s process version](http://electron.atom.io/docs/api/process/#processversionselectron) so we’re always up to date.

Without this, plugins like [`hyperterm-sync-settings`](https://github.com/dfrankland/hyperterm-sync-settings) (will be renamed soon) will break.

### DIFF
--- a/app/plugins.js
+++ b/app/plugins.js
@@ -233,7 +233,7 @@ function install(fn) {
     }
     /* eslint-disable camelcase  */
     env.npm_config_runtime = 'electron';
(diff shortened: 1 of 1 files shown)


matheuss:
Makes sense – thanks! 😄

PR #976, opened 2016-11-09
Always create a new session when creating a new window

CWSpear:
**DO NOT MERGE** until we get a change to discuss further.

I am not sure of any case where you would want to create a new window and not start a new session? This is changed behavior and makes some plugins that created a new window appear to be broken (a window opens, but you can't interact until you create a new tab).

I see this was added in #329, for something to do with opening files.

This behavior is kind of confusing for people who want to create a new "regular window" and have a reference to that window (`app.createWindow` currently does not return anything). So they _have_ to pass a callback, but then that causes their window to appear "broken" as mentioned.

See https://github.com/zeit/hyper/pull/329#r87125720

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -153,17 +153,16 @@ app.on('ready', () => installDevExtensions(isDev).then(() => {
     rpc.on('init', () => {
       win.show();
 
-      // If no callback is passed to createWindow,
-      // a new session will be created by default.
-      if (!fn) {
-        fn = win => win.rpc.emit('termgroup add req');
-      }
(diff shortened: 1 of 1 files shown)


chabou:
Sorry for the delay.

It is a good thing to have a decoupled way to create and init windows.
I want to keep this.

But you're right, it could be useful for plugin to return created window in `app.createWindow`.
I created a new PR for this: https://github.com/zeit/hyper/pull/2081

Thank you for your help and to have pointed this.

PR #977, opened 2016-11-09
Allow require of plugin apis

timneutkens:
Ready to be merged.
Related to #778
Extended upon work here #622


timneutkens:
:shipit:

PR #982, opened 2016-11-11
[Win] Improve design of the scrollbars

stefanivic:
Hyper on Windows uses default scrollbars that do not go well with the new design for Windows. 
I updated the design to change only on win32 so that MacOS and Linux remain the same.

Would love to get some feedback on this (code or design).

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -204,7 +204,7 @@ export default class Term extends Component {
   }
 
   getStylesheet(css) {
-    const blob = new Blob([`
+    const hyperCaret = `
       .hyper-caret {
         outline: none;
         display: inline-block;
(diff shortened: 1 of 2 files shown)


lib/components/term.js
```diff
@@ -184,13 +184,35 @@ export default class Term extends Component {
...
-      }
-      ${css}
-    `], {type: 'text/css'});
-    return URL.createObjectURL(blob);
+    if(process.platform === 'win32') {
+      const scroll = new Blob([`
+        .cursor-node[focus="false"] {
+          border-width: 1px !important;
+        }
+        ::-webkit-scrollbar {
```
fearphage: I have a suggested simplifcation.

```js
getStylesheet(css) {
  const osSpecifcCss = process.platform === 'win32' 
    ? `/* win32 css things */`
    : ''
  ;

  return URL.createObjectURL(new Blob([`
    .cursor-node[focus="false"] {
       border-width: 1px !important;
    }
    ${osSpecificCss}
    ${css}
  `], {type: 'text/css'}));
} 
```

It's less repetitive and it makes it easy to slot in the os-specific bits.

stefanivic: Yeah, this is great. Will update. 
Thanks.

matheuss:
Amazing @stefanivic ☺️ 
I think applying white to it is not a good idea – it makes more sense to apply something like the `borderColor`. Given that, I rebased it for you and changed the implementation to use the border color 😄 Let me know what you think 👌 

kapture 2016-11-19 at 14 30 03

stefanivic:
@matheuss yeah, looks even better. :D

matheuss:
Thank you @stefanivic 🙌

PR #985, opened 2016-11-11
Added pressed state to window buttons on windows / linux

levrik:
Not really visible on minimize, maximize and close buttons but on the menu button.

### DIFF
--- a/lib/components/header.js
+++ b/lib/components/header.js
@@ -232,10 +232,15 @@ export default class Header extends Component {
         height: '34px',
         padding: '12px 15px 12px 15px',
         WebkitAppRegion: 'no-drag',
-        color: '#808080',
+        color: '#FFFFFF',
+        opacity: 0.5,
         shapeRendering: 'crispEdges',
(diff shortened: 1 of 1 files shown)


ppot:
Need picture

levrik:
@ppot Added.

ppot:
Why the preset state on the `X` is red ?

rauchg:
Ping @levrik

levrik:
@rauchg @ppot Ah sorry. I did not change the `X` to being red. That was done before my PR.

matheuss:
Yeah, the close button is read as per the design. CC @CodeTheory please review 😄

CodeTheory:
The close button has a red hover to convey that the close button is something to take attention towards. As has most close buttons on most systems. Since I was trying to design the window controls to be an average of systems that use those window control positions and icons, I found that windows has just the close button holding a coloured hover. ¯\_(ツ)_/¯ consistency.

As for the PR, looks good to me!

Issue #897, opened 2016-10-18
Cannot paste unless window is focused via mouseclick


CodeTheory:
#### Info

Hyper.app version: **0.8.2**
OS: **OSX 10.11.5** _(El Capitan)_
#### Steps to reproduce
1. Blur focus to the Hyper.app window
2. Copy something to your clipboard
3. Tab into Hyper.app
4. Paste

_Additionally:_
Click into Hyper.app after these steps and then try pasting again and it works.

jacksonSingleton:
I'm on Hyper 0.8.1 on El Capitan 10.11.3 and this works perfectly fine for me

WillsB3:
I have this issue on Hyper 0.83 on OS X 10.11.6

xjamundx:
Yeah this used to work fine then started for me about a week or two ago.

DanielSiepmann:
Same issue here with Mac OS Sierra.

deanputney:
I see this issue as well. It will actually start again if I type anything in between clicking on the window and paste. For example: `click on window, paste` works, but `click on window, type 'x', paste` fails.

pablomolnar:
Same as @deanputney on Hyper 0.8.3.873 on OS X 10.12.1 (Sierra)

ppot:
Since PR #892 This should be working since we are not using blur and focus anymore

pablomolnar:
@ppot Still happening in master (6d91ab48d63e30d94c7d43434440772fdb976a7c / 1.4.5).

ambethia:
Oh, thank you for pointing out that it requires a click to focus, that was killing me, it felt so random if it would work or not, at least now I have a work around.

matheuss:
Yo y'all, could you test on latest master? I might have introduced a fix via #1006 👌

pablomolnar:
@matheuss Works great! Thanks!

ppot:
@CodeTheory Test it on resent master if it work. Close 👍

CodeTheory:
Yes, it works for me. Will close.

Issue #861, opened 2016-10-11
Selecting pane on blurred window requires two clicks


jacob-israel-turner:
I use three screens, and constantly jump between them.  When going back to Hyper, I often want to use a different pane than I last had selected.  On first click, the last active pane is selected.  Only on the second click is the new pane (on which I clicked) active.

If this seems like an easy-ish fix, I'd love to submit a PR.

kevinSuttle:
I wonder if this is related #897

Issue #674, opened 2016-08-29
Cursor change on bell


jbergstroem:
If I change default cursor and hit bell (tab or w/e) it switches to block:

![](http://bergstroem.nu/cursor_block.gif)

This is a23022d5e0a7b9d04552337f057a4ac9489b033b built/run on Sierra.

robindboer:
I'm also experiencing this bug.

maxdeviant:
I can have a look at this one.

maxdeviant:
@robindboer Are you on Sierra as well? I haven't upgraded yet, so I'm wondering if the issue is specified to Sierra.

robindboer:
@maxdeviant I'm on Sierra but I have also experienced this issue on El Capitan. Thanks for having a look at this issue. Looking forward to the solution.

maxdeviant:
I'm able to reproduce consistently in `zsh` on OSX El Capitan.

maxdeviant:
The problem appears to be somewhere within `hterm`, but I have yet to track it down.

If you unfocus and then refocus the pane, the cursor will change back to the config value, since we call `this.term.setCursorShape(nextProps.cursorShape);` upon receiving props.

Looking inside of the `node_modules/hterm-umdjs/dist/index.js`, there are some places where it sets the cursor shape by itself:

`hterm.VT.OSC['50']`:

<details>
```js
/**
 * Set the cursor shape.
 *
 * Parameter is expected to be in the form "CursorShape=number", where number is
 * one of:
 *
 *   0 - Block
 *   1 - I-Beam
 *   2 - Underline
 *
 * This is a bit of a de-facto standard supported by iTerm 2 and Konsole.  See
 * also: DECSCUSR.
 *
 * Invalid numbers will restore the cursor to the block shape.
 */
hterm.VT.OSC['50'] = function(parseState) {
  var args = parseState.args[0].match(/CursorShape=(.)/i);
  if (!args) {
    console.warn('Could not parse OSC 50 args: ' + parseState.args[0]);
    return;
  }

  switch (args[1]) {
    case '1':
      this.terminal.setCursorShape(hterm.Terminal.cursorShape.BEAM);
      break;

    case '2':
      this.terminal.setCursorShape(hterm.Terminal.cursorShape.UNDERLINE);
      break;

    default:
      this.terminal.setCursorShape(hterm.Terminal.cursorShape.BLOCK);
  }
};
```
</details>

`hterm.VT.CSI[' q']`:

<details>
```js
[...]

Issue #484, opened 2016-07-28
Issues with copy-paste from keyboard on OS X El Capitan


brettneese:
Having issues with paste using CMD-V on my OSX machine. Hyperterm flat-out refuses to acknowledge my keyboard command.

Other key commands, such as CMD-+ and CMD-- work just fine, however, as do the menu commands. I thought it was just not implemented yet, but I haven't seen any other issues with this reported on OSX, so it looks like a bug.

The only variable on my machine I can think of is that I am using http://pasteapp.me/helper, however, killing Paste altogether does not change anything.
## 

HyperTerm 0.7.1
Electron 1.2.8
darwin x64 15.4.0

matheuss:
If you select some text on HyperTerm you can't `cmd + c` it?

brettneese:
OK, cmd-c is working, but I don't think it was in 0.7. cmd-v is still not working.

jamesfzhang:
I've been able to reproduce this bug by trying to paste something twice. The first paste works, but the second paste yields nothing.

aeromusek:
Like @jamesfzhang, I see this when trying to paste something twice. It happens reliably when inside an SSH connection. Killing the connection and reconnecting solves the problem.

matheuss:
I still can't reproduce 😰 (Tried inside `ssh`, same result)
![cant-reproduce](https://cloud.githubusercontent.com/assets/4721750/17374598/c417f1ae-5983-11e6-9675-4d2be8310475.gif)

brettneese:
Yeah, this is super weird. Anything I can do on my end to help debug?

matheuss:
The console says anything? (`Cmd + Option + I`)

brettneese:
![image](https://cloud.githubusercontent.com/assets/504279/17377968/c5c64002-5981-11e6-82df-0acbcb03e8c7.png)

Only this. Interestingly, I can paste into the console.

jamesfzhang:
Nothing from console, but if I refresh (`Cmd + R`), I can paste multiple times until I press enter, after which I can only paste once.

![copypasta](https://cloud.githubusercontent.com/assets/2057670/17378237/fde7d712-5971-11e6-8180-79a671d69e11.gif)

chabou:
Paste only once ? Even after adding a space after the first past ?

chabou:
Maybe the same origin than #479

rauchg:
[...]

PR #1006, opened 2016-11-16
Add support for composed chars and fix issues with "foreign" keyboard layouts

### LINKED ISSUE #29: Special characters and foreign keyboard not displayed correctly

### LINKED ISSUE #66: can't write accented characters on a portuguese keyboard

### LINKED ISSUE #466: Can't type ~ character
Hello guys,

I can't type the ~ character in Hyperterm in 0.7.1 as in the previous versions. Could someone help me? Thank you so much!

### LINKED ISSUE #518: Can't type " or ' on US Int
When i am typing with the `U.S. International - PC` layout of my mac i cant type the caracter `'` or `"`  neither can i type `~`  nor ` ` ` though. :(  This is kind of a dealbreaker for me.

### LINKED ISSUE #579: Entering a "~" / "`" symbol does not work with German keyboard
I have a iMac with a German keyboard. Entering the tilde symbol using the normal key shortcut `alt+n` does not work. The same with the backtick symbol "`" itself which is placed after the`ß` symbol.

There is also some issue with the dollar sign (shift+3). This jumps a word to the beginning instead of entering the symbol.
## 

HyperTerm 0.7.1
Electron 1.2.8
darwin x64 15.6.0

matheuss:
kapture 2016-11-16 at 15 22 03

## Summary

This PR introduces an attempt to fix all the following issues:

Major:
* **Users on some "foreign" keyboard layouts (Portuguese and Norwegian, for example) weren't able to type characters like `~`, `'`, `"` and so on**
* **Characters that need a composition event couldn't be typed – `á`, `é`, `ü` and so on**

Minor:
* After the bell, the cursor would go back to a `BLOCK` – #674 
* Selecting pane on blurred window requires two clicks – #861 

Other than that, this changes _should_ fix the following:
* Some people can't paste unless Hyper receives a click – #897
* Issues with copy-pasting – #484 #479 

## Prebuilt binaries

Check [here] for a `.dmg`, a `.exe`, a `.deb` and a `.rpm` 😄

## How

The work I did on this PR was just an improvement on top of @rauchg's [method] – huge thanks to him for developing it! Also, thanks @nfcampos for your work on #214 🙌 

We inject a `hyper-caret` div that has `contenteditable="true"` inside `hterm`'s cursor. With that, we're able to show an intermediate state of composition – so you can see (and type) the character you want. 
When the char is ready, `hterm` automatically captures it – that means we only need to worry about showing that intermediate state.

Closes #29; Closes #66; Closes #466; Closes #518; Closes #579;

[method]: https://github.com/zeit/hyper/commit/6ef13a2c3b525910250bc86f252eb3dc5f0f8744
[here]: https://drive.google.com/drive/folders/0BzynZtaOcjwMNGVaVlFXeFBwdFE?usp=sharing

### DIFF
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -258,7 +258,6 @@ hterm.Terminal.prototype.focusHyperCaret = function () {
 };
 
 hterm.Screen.prototype.syncSelectionCaret = function () {
-  console.log(this);
   const p = this.terminal.hyperCaret;
   const doc = this.terminal.document_;
   const win = doc.defaultView;
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -212,6 +212,63 @@ hterm.ScrollPort.prototype.setBackgroundColor = function () {
   this.screen_.style.backgroundColor = 'transparent';
 };
 
+// will be called by the <Term/> right after the `hterm.Terminal` is instantiated
+hterm.Terminal.prototype.onHyperCaret = function (caret) {
+  this.hyperCaret = caret;
+
+  // we can ignore `compositionstart` since chromium always fire it with ''
+  caret.addEventListener('compositionupdate', () => {
+    this.cursorNode_.style.backgroundColor = 'yellow';
+    this.cursorNode_.style.borderColor = 'yellow';
+  });
+
+  // at this point the char(s) is ready
+  caret.addEventListener('compositionend', () => {
+    this.cursorNode_.style.backgroundColor = '';
+    this.setCursorShape(this.getCursorShape());
+    this.cursorNode_.style.borderColor = this.getCursorColor();
+    caret.innerText = '';
+  });
+
+  // we need to capture pastes, prevent them and send its contents to the terminal
+  caret.addEventListener('paste', e => {
+    e.stopPropagation();
+    e.preventDefault();
+    const text = e.clipboardData.getData('text');
+    this.onVTKeystroke(text);
+  });
+
+  // here we replicate the focus/blur state of our caret on the `hterm` caret
+  caret.addEventListener('focus', () => this.cursorNode_.setAttribute('focus', true));
+  caret.addEventListener('blur', () => this.cursorNode_.setAttribute('focus', false));
+
+  // this is necessary because we need to access the `document_` and the hyperCaret
+  // on `hterm.Screen.prototype.syncSelectionCaret`
+  this.primaryScreen_.terminal = this;
+  this.alternateScreen_.terminal = this;
+};
+
+// ensure that our contenteditable caret is injected
+// inside the term's cursor node and that it's focused
+hterm.Terminal.prototype.focusHyperCaret = function () {
+  if (!this.hyperCaret.parentNode !== this.cursorNode_) {
+    this.cursorNode_.appendChild(this.hyperCaret);
+  }
+  this.hyperCaret.focus();
+};
+
+hterm.Screen.prototype.syncSelectionCaret = function () {
+  console.log(this);
+  const p = this.terminal.hyperCaret;
+  const doc = this.terminal.document_;
+  const win = doc.defaultView;
+  const s = win.getSelection();
+  const r = doc.createRange();
+  r.selectNodeContents(p);
+  s.removeAllRanges();
+  s.addRange(r);
+};
+
 // fixes a bug in hterm, where the shorthand hex
 // is not properly converted to rgb
 lib.colors.hexToRGB = function (arg) {
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -213,6 +213,16 @@ hterm.Screen.prototype.syncSelectionCaret = function () {
   s.addRange(r);
 };
 
+// fixes a bug in hterm, where the cursor goes back to `BLOCK`
+// after the bell rings
+const oldRingBell = hterm.Terminal.prototype.ringBell;
+hterm.Terminal.prototype.ringBell = function() {
+  oldRingBell.call(this);
+  setTimeout(() => {
+    this.restyleCursor_();
+  }, 200);
+};
+
 // fixes a bug in hterm, where the shorthand hex
 // is not properly converted to rgb
 lib.colors.hexToRGB = function (arg) {
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -183,7 +183,7 @@ hterm.Terminal.prototype.onHyperCaret = function (caret) {
     this.restyleCursor_();
   });
   caret.addEventListener('blur', () => {
-    this.cursorNode_.setAttribute('focus', false)
+    this.cursorNode_.setAttribute('focus', false);
     this.restyleCursor_();
   });
 
@@ -216,7 +216,7 @@ hterm.Screen.prototype.syncSelectionCaret = function () {
 // fixes a bug in hterm, where the cursor goes back to `BLOCK`
 // after the bell rings
 const oldRingBell = hterm.Terminal.prototype.ringBell;
-hterm.Terminal.prototype.ringBell = function() {
+hterm.Terminal.prototype.ringBell = function () {
   oldRingBell.call(this);
   setTimeout(() => {
     this.restyleCursor_();
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -1,5 +1,5 @@
 import {hterm, lib} from 'hterm-umdjs';
-import fromCharCode from './utils/key-code';
+// import fromCharCode from './utils/key-code';
 
 const selection = require('./utils/selection');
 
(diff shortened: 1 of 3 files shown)


rauchg:
👊

sergiovilar:
:+1:

nfcampos:
great to see this finally resolved! :)

kauegimenes:
Finally i can start using hyper!

YuriBrunetto:
Great work, everybody!
👏 👏 👏 👏 👏 👏

bdc-stripe:
zomg this makes me so happy 😍

frdmn:
Well done! 👍

albinekb:
@jbergstroem there isn't the hterm code is hosted and maintained over at googles [git](https://chromium.googlesource.com/apps/libapps/+/HEAD/hterm)

But i think it's an easy fix and even possible to do just by some more prototype hacking 😏 
I'll have a look later this week.

Great job @matheuss

matheuss:
@jbergstroem there's already a `setTimeout` in `hterm`'s code 😄 I just created another one to add that call 👌 Other than that, the bell will not play more than once per second (guaranteed by `hterm`'s code), so we're good 👌 

Thank you @albinekb 🙌

jbergstroem:
@matheuss thanks for elaborating.

lib/hterm.js
```diff
@@ -212,6 +151,102 @@ hterm.ScrollPort.prototype.setBackgroundColor = function () {
...
+};
+
+// fixes a bug in hterm, where the cursor goes back to `BLOCK`
+// after the bell rings
+const oldRingBell = hterm.Terminal.prototype.ringBell;
+hterm.Terminal.prototype.ringBell = function () {
+  oldRingBell.call(this);
+  setTimeout(() => {
+    this.restyleCursor_();
+  }, 200);
```
rauchg: We need to add some sort of `if` check or clear this timeout somewhere :P

matheuss: @rauchg I'm not sure if it's necessary – it's just an extension of the timeout that `hterm` creates:

```javascript
hterm.Terminal.prototype.ringBell = function() {
  this.cursorNode_.style.backgroundColor =
      this.scrollPort_.getForegroundColor();

  var self = this;
  setTimeout(function() {
      self.cursorNode_.style.backgroundColor = self.prefs_.get('cursor-color');
    }, 200);
...
```

matheuss: What do you think?

jbergstroem:
@albinekb that means bug tracker lives here: https://bugs.chromium.org/p/chromium/issues/list

kentcdodds:
Hmmm... Was this supposed to resolve the emoji issues? I did a fresh download/install and it appears to still not be working well for me... Am I missing something?

hyper

noamcore:
This pull request has not been released yet.

danilomiranda:
Great man! Thank you very much!

karlpokus:
Is [1.0.0](https://github.com/zeit/hyper/releases/tag/1.0.0) supposed to have support for å, ä and ö? I'm on 1.0.0.1303 and it does not work.

noamcore:
@karlpokus yes.

screen shot 2016-12-13 at 4 05 22 pm

karlpokus:
@noamcore Wierd. Still don't work. My <kbd>ö</kbd> returns `(arg: 6)`. Is my keyboard f**ked?

jbergstroem:
@karlpokus works for me too. Perhaps check your locale env?

karlpokus:
@jbergstroem 
```
~ $ locale
LANG="sv.UTF-8"
```
OS X 10.11.5

jbergstroem:
@karlpokus shouldn't that be sv_SE.UTF-8?

jmml97:
hi! I have the same problem as @karlpokus I cannot type á or ñ on 1.0. It returns (arg: 2). What is wrong? My locale shows `LANG="es.UTF-8"`. Tried changing it to `LANG="es_ES.UTF-8"` but it didn't work...

noamcore:
```
$ locale
LANG=
LC_COLLATE="C"
LC_CTYPE="UTF-8"
LC_MESSAGES="C"
LC_MONETARY="C"
LC_NUMERIC="C"
LC_TIME="C"
LC_ALL=
```

screen shot 2016-12-15 at 1 19 02 pm

Which keyboard layout are you using?

negatron99:
On Windows 10, attempting to type Japanese characters using the IME. 

When typing the first character, the IME immediately disables itself (a cross in it's notification icon) and only the next character is outputted to the terminal. The IME then appears to enable itself (Hiragana symbol) but it continues this cycle of being disabled and enabled.

noamcore:
@negatron99 Japanese characters **are not working** even in macOS.

I tried right now, with no success.

alayii:
Chinese input method also not working in macOS 10.12.2

if using macOS Pinyin input method, when I type `a`, nothing appeared in terminal, type `a` again, character `a` appeared, I cannot type spacebar to choose chinese character, and  candidate box always blink.

if using [Squirrel](https://github.com/rime/squirrel) input method, every typing will directly send a chinese character to terminal, cannot choose.

and character width is not right.

karlpokus:
@jbergstroem Yes it should.
```
# hyper
$ locale
LANG="sv.UTF-8"

# terminal
$ locale
LANG="sv_SE.UTF-8"
```
Are they suppose do differ?
@noamcore I am using `Svensk - Pro` keyboard layout.

jbergstroem:
@karlpokus you should set it (correctly) as part of your shell init. Terminal probably does that for you.

karlpokus:
@jbergstroem So you're saying this setting is on the user - not a bug in hyper? or a temp solution?

jbergstroem:
@karlpokus setting a proper locale should be done in the shell, not in the terminal. Some terminals does it for you though. Anyway, I don't even know if this fixes it for you :)

karlpokus:
@jbergstroem Didn't know the terminal on macOS wasn't a shell :/

Anyways. `sv.UTF-8`is not a valid option it seems and maybe hyper should consider changing it?
```
# in both terminal and hyper
~ $ locale -a | grep sv -i
sv_SE
sv_SE.ISO8859-1
sv_SE.ISO8859-15
sv_SE.UTF-8
```

kleinfreund:
@karlpokus If the wrong value `sv.UTF-8` is causing your issue then this is on you to change, not Hyper.

karlpokus:
@kleinfreund cool. No Problemo. FWIW I can't remember changing this value myself so I thought this might be a bug in hyper considering my terminal got it right.

jmml97:
Typing `locale` on macOS terminal shows me `es_ES.UTF-8` but on Hyper I get `es.UTF8`

jmml97:
Doing what @karlpokus says above resolves the problem.

Simon-L:
Hi, I've done my best to help with improving this as I agree it's an absolute maximum priority issue.
All my research was done on Linux (Mint 18 Cinnamon) set to **French** language and without looking at Hyper's code, *at least for now*.

Here is what I have:
On **Debian** based distros, as per [Debian wiki](https://wiki.debian.org/Locale#Standard), one place to look at would be `/etc/defaults/locale`:
```
$ cat /etc/defaults/locale
#  File generated by update-locale
LANG="fr_FR.UTF-8"
```
More specific tools exist on distros using `systemd`, ie. `localectl`:
```
$ localectl
fr_FR.UTF-8
```
`localectl` provides a reliable output inside Hyper, as opposed to `locale` which is set to a wrong value inside Hyper and a correct one in gnome-terminal and in TTYs (ctrl+alt+F1...+F2 etc.)

I have then looked at Node packages used for locale/intl'ization, though they seem to be using LC and LANG environment variables: https://github.com/sindresorhus/os-locale/blob/master/index.js#L9
It should be noted that Node obviously gets it right outside of Hyper: `process.env.LANG` has the correct value.

It seems Hyper is getting the locale value from the wrong source. I'm still looking for gnome-terminal's way of doing this. Editing either .bashrc or .bash_profile fixes it but I believe that similarly to the many other applications one is running, Hyper should get it right from the start.

PR #1017, opened 2016-11-19
Add shell menu feature

Specro:
Adds right click 'Open Hyper here' shell menu feature on Windows. 

Open Hyper here

Depending on the shellMenu option in config it adds or removes the entries from the registry.

* Need tests on macOS and Linux to make sure it doesn't brake anything.
* Maybe a better name than shell menu?
* Add & remove methods have empty callbacks so far. Maybe returning something could be useful for notifications, etc.


matheuss:
Amazing, @Specro 😄 

Maybe we could go option-less on this one 🤔 Do you think someone would like to disable this? It'd be annoying to not be able to disable it?

For the name: maybe `systemContextMenu` or something 🤔

Specro:
I also thought the option wasn't necessary but left it just in case. Came close to that with `contextMenu` just couldn't figure out a modifier 😅 so there wouldn't be any confusion with terminal context menu. Thanks, gonna apply these with next commit :smile:

levrik:
Could you please ensure that the menu item gets removed on uninstall of Hyper? Would be annoying to remove it manually after an uninstall.

matheuss:
That's a good catch, @levrik. I think that to accomplish that, we'll have to handle the Squirrel command for uninstall 🤔

Specro:
I rewrote this to add/remove the context menu during installation/uninstallation. Seems to work fine 😃

matheuss:
Perfect, thank you! 😄

PR #1018, opened 2016-11-19
🙌 unicode/emoji fix 

dotcypress:
Emoji fix.

Related PR: https://github.com/zeit/hyper/pull/740

Before:
screen shot 2016-11-19 at 2 47 39 am

After:
screen shot 2016-11-19 at 2 48 27 am

### DIFF
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -51,6 +51,84 @@ hterm.Terminal.prototype.onMouse_ = function (e) {
 function containsNonLatinCodepoints(s) {
   return /[^\u0000-\u00ff]/.test(s);
 }
+
+// hterm Unicode patch
+hterm.TextAttributes.splitWidecharString = function (str) {
+  const context = runes(str).reduce((ctx, rune) => {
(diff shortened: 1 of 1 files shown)


matheuss:
Amazing, @dotcypress 😌 

We still need to fix the insertion, removal, and the behavior of more complex emojis:

kapture 2016-11-19 at 10 15 43

cc @rauchg @rginda

rauchg:
@dotcypress this is amazing stuff

matheuss:
So, @dotcypress: I think the last thing we should do before merging is finding a way to add `line-height: 1em` to `<x-row>` if there's any emoji on the screen.
This is not the best solution in the world, but I think it's enough until we manage to fix `line-height` on `hterm` 👌

matheuss:
Ping @dotcypress @rauchg: should we merge as it is?

rauchg:
Yes. I'd love to fix the other issue (maybe @dotcypress has good ideas), but for now this is an improvement

PR #1036, opened 2016-11-23
Remove contributors from MIT license

CodeTheory:
@leo Updated here as well. 😄

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 # MIT License
 
-Copyright (c) 2016 Zeit, Inc. and contributors
+Copyright (c) 2016 Zeit, Inc.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal


leo:
😊

caesar:
Why? (Unless you have a contributor agreement which assigns you copyright on all contributions, the previous version was correct and the new version is factually false…)
Just curious :-)

CodeTheory:
Surely the old version assigned responsibility, legally, to all contributors @caesar? This would require ZEIT, who this project actually belongs to, to include contributors in a lot of things legally?

caesar:
But that's not really how copyright works… I'm not sure what you mean by 'responsibility', but anyone who writes any code (or anything else) owns the copyright in it, unless they explicitly assign that copyright to someone else. So technically, the copyright in each line of the code belongs to the individual who wrote it, not to Zeit.

Of course, since it's all MIT licensed, this isn't a practical problem – though it does mean that you can't change the license later, without permission from every contributor.

But still, technically, the old version was right and the new version is wrong, unless every contributor explicitly assigns their copyright to Zeit.

Anyway – as I haven't contributed (yet?) I have no stake here, I was just browsing the changelog and thought it was a bit odd.

PR #1048, opened 2016-11-28
Fix a bug related to the alpha hex calculation

### LINKED ISSUE #873: Transparency breaks until config is reloaded
Currently, whenever I open hyper, the window is drawn fully opaque regardless of the `background` setting. If I do a full reload, transparency on the window is re-enabled. This does not work with a normal reload.

I'm currently running hyper with `--enable-transparent-visuals` and `--disable-gpu` as recommended and my system is running:

```
OS:         Arch Linux x64
WM:         bspwm
Compositor: compton
```

I suspect this is a problem with my window manager moving or resizing the window after it's been created, although, after a full reload, I can move and scale the window without a problem, floating or not. 

My `.hyper.js` is:

```
module.exports = {
  config: {
    fontSize: 13,
    fontFamily: 'Menlo, "DejaVu Sans Mono", "Lucida Console", monospace',
    cursorColor: 'rgba(248,28,229,0.75)',
    cursorShape: 'BLOCK',
    foregroundColor: '#fff',
    backgroundColor: 'rgba(0,0,0,0)',
    borderColor: '#333',
    css: '',
    termCSS: '',
    padding: '0px',

    colors: {
      black: '#000000',
      red: '#ff0000',
      green: '#33ff00',
      yellow: '#ffff00',
      blue: '#0066ff',
      magenta: '#cc00ff',
      cyan: '#00ffff',
      white: '#d0d0d0',
      lightBlack: '#808080',
      lightRed: '#ff0000',
      lightGreen: '#33ff00',
      lightYellow: '#ffff00',
      lightBlue: '#0066ff',
      lightMagenta: '#cc00ff',
      lightCyan: '#00ffff',
      lightWhite: '#ffffff'
    },
    shell: ''
  },
  plugins: [],
  localPlugins: []
};
```

dkniffin:
I was running into an issue on OSX, where I'd set the backgroundColor to something like `'rgba(0, 0, 0, 0.8)'`, and everything would look great, until I closed Hyper, and re-opened it. Once I did that, it would still have transparency, but it'd be more transparent.

I tracked this down to the change I made. Basically, what was being done before is `0.8` would be converted to the decimal number 80, which is not correct. What should happen is that `0.8` should be a percentage of 100% opacity, which is 255, then converted to a hex string (via `toString(16)`). Hopefully that makes sense.

I'm not 100% sure, but I think this might also resolve #873, but I'd need @Lindenk to confirm.

### DIFF
--- a/app/utils/to-electron-background-color.js
+++ b/app/utils/to-electron-background-color.js
@@ -8,6 +8,8 @@ module.exports = bgColor => {
   if (color.alpha() === 1) {
     return color.hexString();
   }
-  // (╯°□°）╯︵ ┻━┻
-  return '#' + Math.floor(color.alpha() * 100) + color.hexString().substr(1);
+
+  // http://stackoverflow.com/a/11019879/1202488
+  const alphaHex = Math.round(color.alpha() * 255).toString(16);
+  return '#' + alphaHex + color.hexString().substr(1);
 };


dkniffin:
Not sure if I should add this comment or not: `// ┬──┬ ノ( ゜-゜ノ)`

Lindenk:
Well, the behavior is different this time. Rather than starting with an all white background, it looks normal until a reload, which might be what this fixes for me. Unfortunately transparency doesn't work at all now.

I'm running the default config with only the following `hyper.js` changes:

```
backgroundColor: 'rgba(0,0,0,0.6)',

showHamburgerMenu: true,
``` 

Well it was worth a try :smile: 

hyper_transparancy_bug

Lindenk:
Wait wait, apparently this only seems to be a problem when running in dev mode. Also, it looks like `--enable-transparent-visuals` and `--disable-gpu` weren't passed to the app when running in dev mode. It seems to work correctly when fully built for release.

dkniffin:
@Lindenk So does this PR fix #873?

Lindenk:
This seems to fix (the more important) half of the issue. I'll make a new issue on the remaining problem with `full-reload`, but otherwise, yes this PR does fix the majority of the problem. Thanks!

SirTimmyTimbit:
+1 happens in Windows too.

SirTimmyTimbit:
Please can someone review this? I've been having the same problem on Windows 10 x64, Hyper 1.0.0.

dkniffin:
Not exactly sure who has write access to this project, but looking at some recently merged PRs, @rauchg seems to. Would you mind taking a look over this and merging it? Thanks!

ppot:
Can you rebase to latest master? @dkniffin

dkniffin:
@ppot Done.

PR #1052, opened 2016-11-29
Added condition for update failing (FEED_URL)

vaporwavie:
I thought it would be nice some return from Hyper when failing the update searching process, so I added an condition to check if the website returned some package. If not, the console.error will appear. Simple as that. =)

### DIFF
--- a/app/auto-updater.js
+++ b/app/auto-updater.js
@@ -10,6 +10,9 @@ const platform = process.platform === 'darwin' ?
   'osx' :
   process.platform;
 const FEED_URL = `https://hyper-updates.now.sh/update/${platform}`;
+if (FEED_URL = false) {
+  console.error('Error while looking for updates. Website is not available.')
+}
 let isInit = false;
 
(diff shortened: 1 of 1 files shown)


matheuss:
@vaporwavie your code makes no sense – and it throws 😰 

```javascript
const FEED_URL = '...';
if (FEED_URL = false)
```

You're trying to assign a value to a constant variable – you can't do this 😅 

Anyways, even if you change it to `if (FEED_URL === false)`, the code would still continue to make no sense: `FEED_URL` will never be `false` – it will always be a `string`.

Other than that, we already have something to show errors during the _check for updates_ process: [check here](https://github.com/zeit/hyper/pull/1052/files#diff-b5fec21ff458ca905c3ae9aaf28685ebR19)

Thanks for trying to help 🙌

vaporwavie:
Oh, I see... Well, time to study more about Hyper's code then 😄 
Thanks for clarifying!

PR #1065, opened 2016-12-01
Fix "Update Plugins" for shells that don't have `&&`

dfrankland:
Some shells (like `fish`) do not support `&&`, rather they use other syntax like `foo; and bar`. This fixes that issue.


dfrankland:
There's precedent for using `bash` in this way, see https://github.com/zeit/hyper/blob/31721e5c29524742af26387dc87238d403db3e2d/lib/actions/ui.js#L208

Issue #1030, opened 2016-11-23
Not able to type more than 77 characters from 2nd tab onwards


emaillenin:
- ✔️ I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version - **0.8.3**
- ✔️ I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**:  OS X El Capital 10.11.6
- **Hyper.app version**: 0.8.3
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/emaillenin/3f28ad9ea2075f816c56757dd2d8a3e8
- **Relevent information from devtools** _(CMD+SHIFT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: 

```
bundle.js:620 (re)loading renderer plugins
index.html:34 total init time 194.35500000000005
bundle.js:36273 Unknown OSC code: "1"
bundle.js:36273 Unknown OSC code: "1"
```

- **The issue is reproducible in vanilla Hyper.app**: Is Vanilla

## Issue
Only for tabs created after the first tab, whatever I type after the 77th character goes back to the first character. I am just pasting "1234567890" continuously in the first tab (which pastes continuously as expected) but while pasting in the 2nd tab, during the 8th paste, the paste goes back to the first character. GIF below:

![hyper-1](https://cloud.githubusercontent.com/assets/1235853/20548641/4c8b7afe-b179-11e6-8358-0ff37dd9c761.gif)

CodeTheory:
I can't seem to reproduce this. 🤔 
Anyone else? CC/ @matheuss

matheuss:
Same, can't reproduce 😥

PR #1075, opened 2016-12-02
Specify cols key when spawning pty shell

breber:
The column size param in pty.js is named ‘cols’, but the shorthand notation treats it as ‘columns’, causing the column count to not be correct.

I was seeing the issue described in zeit/hyper#1030 both on Mac and PC, and this fixed the issue.


matheuss:
Thanks 🙌

Issue #1039, opened 2016-11-24
Win32 error code 267


urec:
- [true] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [true ] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: 
- **Hyper.app version**: 

No work.(( pyton to PATH (true). I have error after command(npm run app)
![-1](https://cloud.githubusercontent.com/assets/12876772/20605561/0c7dad42-b27d-11e6-95ac-016910d0a817.jpg)

DanielRamosAcosta:
Exact same error here.
![image](https://cloud.githubusercontent.com/assets/11427028/20607488/bb2b6e72-b270-11e6-9370-ecaac3118e4c.png)

* Windows 10 compilation 14393.447
* Latest in `master`, In about Hyper says "Hyper 810e0a9"

stefanivic:
@urec @DanielRamosAcosta i had the same issue, after removing the directory and cloning it again it works fine. Try that, if it doesn't work report back please.

DanielRamosAcosta:
I completly deleted the repo, I cloned it again freshly and I still have the same issue 😢

DanielRamosAcosta:
I have python version 2.7.6, just in case:

```
D:\...\hyper [master ≡]> python --version
Python 2.7.6
```

urec:
@stefanivic no work

Stanzilla:
Hey there, please try again with the new Hyper v2 release and open a new issue if the problem stills exists. Thank you!

PR #1076, opened 2016-12-03
Use homedir() from os package instead of env vars

breber:
I was seeing the issue described in zeit/hyper#1039 on a Windows machine, and it seemed to start happening after 0c30fc0bcfe1b6667b958896bf0bdbf801afe077 (where HOME and HOMEPATH were reversed). In any case, switching to use os.homedir() in place of the environmental variables worked for me.


leo:
Thanks!

PR #1098, opened 2016-12-08
Unicode magic

### LINKED ISSUE #154: Emoji push the line they're on down into the line below
screen shot 2016-07-16 at 12 08 46 am

In this screenshot, the top two lines and the bottom two lines are evenly spaced. But the middle line with the emoji is pushed downward and collides with the line below it. You can even see it when the row is highlighted in the Inspector:

screen shot 2016-07-16 at 12 12 00 am

This is the output of Homebrew's `brew upgrade` command, if you want to try and replicate it.

dotcypress:
Fix line height for `x-row` containing emoji.

Fixes #154

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -231,6 +231,9 @@ export default class Term extends Component {
       .cursor-node[focus="false"] {
          border-width: 1px !important;
       }
+      x-row {
+        line-height: 1em;
+      }
       ${hyperCaret}
       ${osSpecificCss}
       ${css}


pablopunk:
The bug above was tested on a monitor. This doesn't seem to be an issue on my macbook display 

screen shot 2016-12-09 at 6 45 28 pm

CodeTheory:
@dotcypress @matheuss Perhaps we could bump up the line-height to 1.2em to protect from this ^ and for more readability too?

dotcypress:
@CodeTheory @matheuss absolutely agree 👍

And `1.2em` will fix this: 
screen shot 2016-12-09 at 4 57 42 pm

ppot:
@dotcypress Do it i will merge when done.

dotcypress:
@ppot Done
https://github.com/zeit/hyper/pull/1107

PR #1111, opened 2016-12-11
🚀 Performance tuning & bugfix

dotcypress:
Fix unicode performance (cause separated html element per character is too slow)

**🙌 Make Hyper fast again!**

UPD: 🚨 this PR also will fix > 5 opened issues

@matheuss @CodeTheory @ppot @rauchg

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -231,9 +231,6 @@ export default class Term extends Component {
       .cursor-node[focus="false"] {
          border-width: 1px !important;
       }
-      x-row {
-        line-height: 1.2em;
-      }
       ${hyperCaret}
       ${osSpecificCss}
       ${css}
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -43,10 +43,10 @@ hterm.TextAttributes.splitWidecharString = function (str) {
       return ctx;
     }
     if (ctx.acc) {
-      context.items.push({str: ctx.acc});
+      ctx.items.push({str: ctx.acc});
       ctx.acc = '';
     }
-    context.items.push({str: rune, wcNode: true});
+    ctx.items.push({str: rune, wcNode: true});
     return ctx;
   }, {items: [], acc: ''});
   if (context.acc) {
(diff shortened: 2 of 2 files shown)


matheuss:
@dotcypress just tested here and the performance looks like the same as before 😰 Each braille char is still alone inside its own `span` 🤔 
Also, if I change the font size via <kbd>Cmd</kbd> + <kbd>+</kbd> and/or <kbd>Cmd</kbd> + <kbd>i</kbd> I get some weird behavior 😩 

screen shot 2016-12-11 at 8 14 19 am

dotcypress:
@matheuss yes, we still need to wrap each `non latin symbols` to `span`.

But now, huge amount of spans created if string contains at least 1 non latin character, see screenshots below, all `spans` are red 😂 

Before:
screen shot 2016-12-11 at 9 51 19 am

After:
screen shot 2016-12-11 at 9 53 43 am

Currently i'm trying to fix issue with increasing/decreasing font size, stay tuned 😎

dotcypress:
Update:

I can reproduce bug with `vtop` and changing font size even on `master` branch.
screen shot 2016-12-11 at 4 13 57 pm


After font change, some characters are gone,  for example:
`" CPU USAGE " -> "CPU USAGE"`
screen shot 2016-12-11 at 11 33 39 am

But if you try this:  `curl https://gist.githubusercontent.com/dotcypress/29ed6c767e824b6241b7465f9d036567/raw/72e75feba4a2a4fbe20dc79a8af6f204c1b096c5/vtop-snapshot.txt | cat` you can change font size without that weird behavior.  🤔 

@matheuss

dotcypress:
Bump 😎

@matheuss

dimitrieh:
will this fix slow scrolling in for example micro-editor.github.io and slow performance with cat long files?

dotcypress:
@dimitrieh not fully fixed but speed of scrolling is better than in `1.0.0`

b3ll:
@dotcypress whilst this is a clever way of speeding things up, not all characters are displayed as <terminal-font-size>px, which causes layout to break. In my case, I have an emoji in the prompt (OS X, 10.12), and this causes all subsequent characters to be displayed wrongly by about half a character's width (or more). 

For this solution to work, any special spans should have their width set by either the DOM automatically, or using canvas's text measuring functions, rather than a global special span width.

In the screenshot, the cursor is supposed to be right after the -la, but it ends up being an entire character's width off.

span

rauchg:
We should look into adding a character measurement cache perhaps?

Thanks for your analysis Adam

b3ll:
perhaps? I'm not entirely sure what the best fix is, I'm not super familiar with current web technologies. I just do know that setting a global CSS style for all specially marked unicode characters won't work.

A CSS style for each character should work, but I can see other languages / character sets causing performance to take a nose dive.

rauchg:
Also, my understanding is that we only need two character classes: single
and double width, since we are dealing with a grid.

alexandernst:
This sounds way too familiar with what the guys from Atom have been working
on for ages...

Issue #828, opened 2016-10-09
select Up, Down, Left and Right panes


iamstarkov:
right now its only prev/next pane, the same as tabs. Its enough for tabs, because tabs are one dimensional array.

But as far as panes are two dimensional feature, we need to navigate between them in all four sides: go to up/down/left/right pane.

OJFord:
I was going to suggest numbers when I found this - I don't use prev/next for tabs, and the Ctrl-Option[-Shift]-Tab shortcut to do so for panes is particularly awkward.

Cmd-N to choose a tab and Option-(*numbers appear on panes*)-N to choose a pane would be great though.

I'm not suggesting it instead of directions, but directions instead of tabbing, and by-numbers available for panes too would be :+1:.

iam4x:
I like the iTerm2 shortcuts:

* Cmd+left arrow, Cmd+right arrow navigates among tabs. So does Cmd-{ and Cmd-}.
* Cmd+number navigates directly to a tab.
* Cmd+Option+Number navigates directly to a window.
* Cmd+Option+Arrow keys navigate among split panes.
* Cmd+] and Cmd+[ navigates among split panes in order of use.

source: https://www.iterm2.com/documentation-one-page.html

spirosikmd:
I think the last point is the most important regarding UX

`Cmd+] and Cmd+[ navigates among split panes in order of use`

Now we need to use 

```
selectNextPane: 'Ctrl+Alt+Tab',
selectPreviousPane: 'Ctrl+Shift+Alt+Tab'
```

I don't know about others, but for me this is difficult.

spro:
I like how tmux does it, where there is a single shortcut to toggle between the latest 2 panes.

joshschmelzle:
Just to add, under Windows 10, Ctrl + Alt + Tab is another alternative for Alt + Tab. It is a cross between Windows + Tab and Alt + Tab. It keeps the task switcher on the screen but does not include any ability to manage desktops. Ctrl+Alt+Tab. 

So, for me under Windows 10 (version 1511 build 10586.679), Ctrl + Alt +Tab brings up the task switcher and does not navigate between split panes.

langri-sha:
Ctrl+Alt+Tab and Ctrl+Shift+Alt+Tab do not work under Gnome, as they're assigned by the window manager for switching panels.

If anyone is interested in releasing theirs to use Hyper, run `dconf-editor` and disable these two keybindings by overriding the defaults and configuring `[]`:

[...]

PR #1113, opened 2016-12-11
Use the same accelerator as iterm for navigation between split panes

spirosikmd:
This fixes `Cmd+] and Cmd+[ navigates among split panes in order of use` in #828. Should be ready to be merged, but I am not sure if the linux accelerator is the ideal one, can you please verify this?

### DIFF
--- a/app/accelerators.js
+++ b/app/accelerators.js
@@ -41,8 +41,8 @@ const applicationMenu = { // app/menu.js
   minimize: 'M',
   showPreviousTab: 'Alt+Left',
   showNextTab: 'Alt+Right',
-  selectNextPane: 'Ctrl+Alt+Tab',
-  selectPreviousPane: 'Ctrl+Shift+Alt+Tab',
+  selectNextPane: isMac ? ']' : 'Ctrl+Alt+]',
+  selectPreviousPane: isMac ? '[' : 'Ctrl+Alt+[',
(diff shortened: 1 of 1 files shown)


ekmartin:
The reason we didn't use Cmd+][ in the first place wasn't technical, if I remember correctly it was because it doesn't really work on European keyboards where you don't have a specific key for [ and ]. I agree that Cmd+Alt+Tab isn't as convenient as it should be though, but we already use Cmd+Tab for cycling between tabs.

spirosikmd:
It is your call guys. I still believe that using (in most cases) the default iterm shortcuts would be really useful. I know most (if not all) of my colleagues use iterm. Also those who don't have the english keyboard, they do change to an english keyboard layout while writing code. Which solves the issue in the first place.

kleinfreund:
> they do change to an english keyboard layout while writing code.

Switching keyboard layouts is not solving the issue at hand, it's a mere workaround.

MartinDelille:
+1 for *iTerm* shortcut. I personnaly have a AZERTY physical keyboard but I switch to QWERTY layout when writing code and using the terminal. 

Some people think it's crazy but I think they don't realize the confort it brings!

spirosikmd:
Closing this PR as a similar [plugin](https://github.com/chabou/hyper-pane) has been made.

PR #1138, opened 2016-12-12
Open Preferences in Default JS Editor on Windows

### LINKED ISSUE #1126: Preferences are opened in notepad.exe on Windows, mangling line endings
On Windows, the settings are opened in notepad.exe which does not support line breaks. You should either use `\r\n` in your settings or open the settings with wordpad or any other app that supports Linux-style link breaks.

jongalloway:
Notepad doesn't handle `\n` line breaks, so config file is unreadable. 

Updated PR does the following:
* If not Windows, follows same logic as before
* If Windows, checks for default JS editor. If set and not notepad, opens hyper.js in default JS editor.
* If default JS editor not set or notepad, shows file in folder using electron shell command.

PR is ready to be merged, I'm not aware of any related PRs, and you're free to edit in any way.

Fixes #1126.

### DIFF
--- a/lib/actions/ui.js
+++ b/lib/actions/ui.js
@@ -213,12 +213,11 @@ function getDefaultJsEditor() {
     if (err) {
       // In case of error, return default editor
       return 'notepad.exe';
-    } else {
-      // Default is "C:\Windows\System32\Notepad.exe %1"
-      entry = entry.substring(0, entry.indexOf(' %1'));
-      entry = entry.substring(entry.lastIndexOf('\\'), entry.length);
-      return entry;
-    }
+    } 
+    // Default is "C:\Windows\System32\Notepad.exe %1"
+    entry = entry.substring(0, entry.indexOf(' %1'));
+    entry = entry.substring(entry.lastIndexOf('\\'), entry.length);
+    return entry;
   });
 }
 
--- a/lib/actions/ui.js
+++ b/lib/actions/ui.js
@@ -1,5 +1,5 @@
 import os from 'os';
-import shell from 'electron';
+import {shell} from 'electron';
 import path from 'path';
 
 import * as shellEscape from 'php-escape-shell';
--- a/lib/actions/ui.js
+++ b/lib/actions/ui.js
@@ -213,7 +213,7 @@ function getDefaultJsEditor() {
     if (err) {
       // In case of error, return default editor
       return 'notepad.exe';
-    } 
+    }
     // Default is "C:\Windows\System32\Notepad.exe %1"
     entry = entry.substring(0, entry.indexOf(' %1'));
     entry = entry.substring(entry.lastIndexOf('\\'), entry.length);
--- a/lib/actions/ui.js
+++ b/lib/actions/ui.js
@@ -1,4 +1,4 @@
-import {EOL} from 'os';
+import os from 'os';
 
 import * as shellEscape from 'php-escape-shell';
 import last from '../utils/array';
@@ -29,10 +29,10 @@ import {
 } from '../constants/ui';
 
 import {setActiveGroup} from './term-groups';
+import {shell} from 'electron';
+import path from 'path';
 
 const {stat} = window.require('fs');
-const {shell} = window.require('electron');
-const {Registry} = window.require('winreg');
 
 export function increaseFontSize() {
   return (dispatch, getState) => {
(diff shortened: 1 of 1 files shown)


citycide:
I'm curious - would it be preferable to the team & other users to allow a configurable editor on Windows, like VS Code? Since we're editing a JS file (even minimally) it'd be nice to get highlighting, and neither Notepad or Wordpad support that.

jongalloway:
@citycide We discussed a little on #1126. Windows has a setting for a configurable editor for text files, and for .js files as well (`HKEY_CLASSES_ROOT\JSFile\Shell\Edit\Command`) but they're rarely set.

My preference would be to start with the extremely simple fix of switching to wordpad, since it's at least readable, then add an option to configure it or use default text / js editors. It's currently just about unusable on notepad, since no linebreaks are shown.

jonathandelgado:
`shell.showItemInFolder()` in electron would be a good alternative for windows. This opens the folder in explorer and allows users to open it however they want.

 I also agree with @jongalloway, people rarely set their default text editor. Ultimately, I think defaulting to notepad/wordpad is a bad idea.

citycide:
@jongalloway Ah, I do see that in the issue now. I think I'm on board with the proposal from @jonathandelgado that Hyper should use the `shell.showItemInFolder()` method. This removes the need to run a command to do this.

This would also have the benefit of working if the user has changed the default shell to PowerShell for example, since `start` doesn't work there.

null-reference:
I like the idea of allowing Windows users to set an environment variable to override the editor (ex. %editor%). If the environment variable is not present, fallback to notepad (or wordpad).

The code to change is in "lib/actions/ui.js" and could be modified as follows:
```
const command = isWin ? ' IF EXIST "%editor%" (start "%editor%" "%userprofile%\\.hyper.js") ELSE (start notepad "%userprofile%\\.hyper.js")' : ' bash -c \'exec env ${EDITOR:=nano} ~/.hyper.js\'';
  const message = [];
  message.push(isWin ?
    ' echo Attempting to open ^%userprofile^%\\.hyper.js with ^%editor%^ (if set) or notepad' :
    ' echo Attempting to open ~/.hyper.js with your \\$EDITOR');
```

If this is desired, I can put together a PR.

citycide:
@null-reference I think opening the folder is a better, more universal approach. Running a command requires that Hyper needs to either 1) take into account all the possible shells it could be using, or 2) fail to run the command due to running eg. PowerShell instead of cmd.exe.

Not only that but it's also cross-platform thanks to Electron so there wouldn't even be a need to check if it's Windows or not.

jongalloway:
I'd like to update my pull to essentially do this:

```javascript
if(isWin) {
  shell.showItemInFolder("%userprofile%\\.hyper.js")
} else
{
  //open in nano for *nix
}
```

However, after several hours of messing around with c++, python, git eol settings, etc., [I still don't have it building on Windows](https://github.com/zeit/hyper/issues/1027) so I can't test. Hopefully someone else who does can submit a pull request that does the above.

jonathandelgado:
@jongalloway Push the change and I'll test it for you!

jongalloway:
I pushed an update which continues to work for me on OS X, haven't been able to build for Windows yet.

To test:
* Without having overridden the registry value (so JSFile default editor is notepad.exe), open preferences on Windows. Verify that the directory `%userprofile%` directory is opened, hopefully with `hyper.js` highlighted.
* With a non-default editor associated with JS files (e.g. VS Code), open preferences. Verify that your `hyper.js` is opened in your custom editor. 

ping @jonathandelgado who volunteered to test

lib/actions/ui.js
```diff
@@ -202,37 +204,69 @@ export function moveTo(i) {
...
+    if (!jsEditor || jsEditor.toLowerCase() === 'notepad.exe') {
+      handled = true;
+      shell.showItemInFolder('%userprofile%\\.hyper.js');
+    } else {
+      // eslint-disable-next-line no-template-curly-in-string
+      command = ' start ${jsEditor} "%userprofile%\\.hyper.js"';
+    }
+  }
+
+  if (!handled) {
```
Specro: You should handle code in this if block in a callback since line 241 may happen before `getDefaultJsEditor()` has returned anything as `winreg` is async.

lib/actions/ui.js
```diff
@@ -202,37 +204,69 @@ export function moveTo(i) {
...
+  new Registry({
+    hive: 'HKCR',
+    key: regKey
+  }).get(Registry.DEFAULT_VALUE, (err, entry) => {
+    if (err) {
+      // In case of error, return default editor
+      return 'notepad.exe';
+    }
+    // Default is "C:\Windows\System32\Notepad.exe %1"
+    entry = entry.substring(0, entry.indexOf(' %1'));
```
Specro: Here entry is a `RegistryItem`. The string you need is in `entry.value` 😄

lib/actions/ui.js
```diff
@@ -202,37 +204,69 @@ export function moveTo(i) {
   };
 }
 
+function getDefaultJsEditor() {
+  const regKey = '\\JSFile\\Shell\\Edit\\Command';
+  new Registry({
```
jonathandelgado: Uncaught TypeError: Registry is not a constructor (Line 209)

lib/actions/ui.js
```diff
@@ -202,37 +204,69 @@ export function moveTo(i) {
...
-                ...message,
-                command,
-                ''
-              ].join(EOL)
-            ));
+  if (isWin) {
+    const jsEditor = getDefaultJsEditor();
+    if (!jsEditor || jsEditor.toLowerCase() === 'notepad.exe') {
+      handled = true;
+      shell.showItemInFolder('%userprofile%\\.hyper.js');
```
jonathandelgado: This doesn't work, modifiers like `%userprofile%` must be resolved in node unless you are including it in some kind of a command (like line 237).

A working alternative would be `shell.showItemInFolder(path.join(os.homedir(), '.hyper.js'));` where `path` and `os` are from the NodeJS API.

lib/actions/ui.js
```diff
@@ -202,37 +204,69 @@ export function moveTo(i) {
...
 export function showPreferences() {
   const isWin = process.platform === 'win32';
   // eslint-disable-next-line no-template-curly-in-string
-  const command = isWin ? ' start notepad "%userprofile%\\.hyper.js"' : ' bash -c \'exec env ${EDITOR:=nano} ~/.hyper.js\'';
-  const message = [];
-  message.push(isWin ?
-    ' echo Attempting to open ^%userprofile^%\\.hyper.js with notepad' :
-    ' echo Attempting to open ~/.hyper.js with your \\$EDITOR');
-  message.push(' echo If it fails, open it manually with your favorite editor!');
+  let command = ' bash -c \'exec env ${EDITOR:=nano} ~/.hyper.js\'';
```
jonathandelgado: In my opinion, `shell.showIteminFolder` should be used across all platforms. I'm sure there are a lot of unix users that don't have an editor set and will have to screw around with nano because they are unaware of this line.

End of the day, I don't even know if we should check the registry on windows; this is such an infrequent command it's unfortunate it requires so much complication. My recommendation is to overhaul all of this with a single `shell.showItemInFolder` for all platforms.

jongalloway: Okay, I'm convinced. I'll update my pull request. 

I'm building a Windows VM to test locally too so I can test - my bare metal machines all have multiple versions of Visual Studio and Python, which is not playing well with the build.

jongalloway:
Tested and pushed update. 

I think this is a lot better for Windows, not sure if it's better for *nix. I think the eventual state should default to show-in-folder for Windows, nano for *nix, and include a hyper config setting to specify default config editor. My personal OSS philosophy leans towards pushing this and improving later.

jonathandelgado:
@jongalloway Just to clarify, this doesn't change the live update system. In `app/config.js` Hyper has a constant watch on that file, so everything will work as it did previously in that regard.

This looks like a step in the right direction!

lib/actions/ui.js
```diff
@@ -1,4 +1,6 @@
-import {EOL} from 'os';
+import os from 'os';
+import shell from 'electron';
```
jonathandelgado: `...showItemInFolder is not a function` is thrown because of this line. This isn't importing the way you think it is, this should be `import {shell} from 'electron';`. Right now, it is importing the default electron export, which doesn't lineup to what you are calling below.

paulcbetts:
Can we also write out the default .hyper.js to have CRLF on Windows? We could dodge a lot of silliness this way

jongalloway:
@paulcbetts Definitely - then we could use notepad and nano as the default editors and add a preference to set another editor if desired.

PR #1141, opened 2016-12-12
Add VibrancyType Config Option

IanMitchell:
This is ready to be merged, although it might be good to shorten the config URL to a `hyper.is` alias.

Exposes the VibrancyType config option so it will work with any theme. Reloads the view when a user changes it. On non macOS machines it returns an empty string, which causes [BrowserWindow to ignore it](https://github.com/electron/electron/blob/master/docs/api/browser-window.md#winsetvibrancytype-macos).


albinekb:
Could you rebase? @IanMitchell

IanMitchell:
Out on vacation for a week, but will when I have access to a computer again

app/index.js
```diff
@@ -165,6 +167,9 @@ app.on('ready', () => installDevExtensions(isDev).then(() => {
       // notify renderer
       win.webContents.send('config change');
 
+      // Update Vibrancy
+      win.setVibrancy(toElectronVibrancy(cfg_.vibrancyType));
```
IanMitchell: Seems the background color change moved elsewhere, but I'm having a hard time finding it - did it go into the reducer? I'm not sure I can put this into state; also not sure it belongs. Would appreciate advice!

IanMitchell:
After the update, the "light" vibrancy type doesn't appear to work. I'm really unsure as to why - the rest seemed fine during testing. It's set correctly, but appears as a solid grey background color. Am I missing anything obvious?

matheuss:
@IanMitchell first of all, sorry for the delay.

I'm sorry but we decided to not accept this kind of feature/config option/etc into Hyper. If we decide to accept all of them, Hyper's codebase and `.hyper.js` would be both enormous.

We still need to document that, but our policy is basically: if you can do it as a plugin, it shouldn't be on Hyper's core. We're, for example, moving the `webview` feature (you type a URL and Hyper renders it) into a core plugin very soon.

Ps: make sure to check [hyper-simple-virancy](https://github.com/matheuss/hyper-simple-vibrancy) 🙌

Issue #1166, opened 2016-12-13
Error updating plugins


urkopineda:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 8.1 x64, v6.3.9600
- **Hyper.app version**: 1.0.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: [LINK](https://gist.github.com/urkopineda/7e48770b52b0fa18897237053dcde730)
- **Relevent information from devtools** _(CMD+SHIFT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: N/A
- **The issue is reproducible in vanilla Hyper.app**: Is Vanilla

## Issue
I have an error at start, says that plugins cannot be updated, see log (and the log isn't where is pointing to, '~\.hyper-plugins\npm-debug.log'. I can reproduce the bug not only restarting the app but going to menu, 'Plugins' -> 'Update All Now'.

In my particular case, I have a Windows 8.1 **without administrator rights** at work so I think that limitation is causing the problem.

And, of course, amazing app, it's superb!

sallar:
I have the same issue. Everything is broken on all my Mac machines.

jonathandelgado:
Same issue on Windows 10 with admin rights, I'll try to do some more digging.

jonathandelgado:
Found the problem.

I was getting the error:
```
{ Error: Command failed: npm prune && npm install --production
bash: /s: No such file or directory

    at ChildProcess.exithandler (child_process.js:218:12)
    at emitTwo (events.js:106:13)
    at ChildProcess.emit (events.js:191:7)
    at maybeClose (internal/child_process.js:877:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:226:5)
```

Looking into the `install` function, I noticed the path was correct, but the shell was interpreting it incorrectly. Commenting out the `shell` argument resolved the issue ([line 243 currently](https://github.com/zeit/hyper/blob/a5b457ed13d649be41255310f08f59ac3464769b/app/plugins.js#L242)). I had my shell set to `bash.exe`, but the default was also causing issues.

Not going to throw in a PR because i'm not sure how modifying the shell will effect other operating systems, but hopefully this comment helps.

sallar:
[...]

PR #1181, opened 2016-12-13
Add support for the fish shell

### LINKED ISSUE #1142: Plugin-Update does not work when Shell is set to Fish

- **OS version and name**: macOS 10.12.1
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: [My .hyper.js](https://gist.github.com/hoffi/7c4277ddfa4dc893fed2ad2d668f60b4)
- **Relevent information from devtools** _(CMD+SHIFT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: None

## Issue
When i configure fish as my shell in `.hyper.js` the plugin update always fails. 
The problem is [this line](https://github.com/zeit/hyper/blob/master/app/plugins.js#L239) which executes `npm prune && npm install --production` in the configured shell. When i change this to `npm prune; and npm install --production` it works for me.

Maybe the best would be to not run it directly on the configured shell and use something like for the preferences: `bash -c "exec env npm prune && npm install --production"`

knewter:
Fixes #1142

The fish shell is not POSIX-compliant.  As a result, the installation
command's use of `&&` caused `updatePlugins` to fail, erroneously
claiming there would be details in `~/.hyper_plugins/npm-debug.log`.

They of course weren't there because the command it tried to run was an
invalid command.  I've added an object to choose the install command to
run based on the shell you're in, and a very basic test to determine if
we're in fish.  Most shells should be able to be handled by the
'default' key, so for now it just checks to see if it's fish by doing a
regex on the configured `shell` option.

### DIFF
--- a/app/plugins.js
+++ b/app/plugins.js
@@ -236,7 +236,15 @@ function install(fn) {
     env.npm_config_target = process.versions.electron;
     env.npm_config_disturl = 'https://atom.io/download/atom-shell';
     /* eslint-enable camelcase  */
-    exec('npm prune && npm install --production', {
+    // Shell-specific installation commands
+    const installCommands = {
+      fish: 'npm prune; and npm install --production',
+      default: 'npm prune && npm install --production'
+    };
+    // determine the shell we're running in
+    const whichShell = shell.match(/fish/) ? 'fish' : 'default';
+    // Use the install command that is appropriate for our shell
+    exec(installCommands[whichShell], {
       cwd: path,
       env,
       shell
--- a/app/plugins.js
+++ b/app/plugins.js
@@ -239,10 +239,10 @@ function install(fn) {
     // Shell-specific installation commands
     const installCommands = {
       fish: 'npm prune; and npm install --production',
-      default: 'npm prune && npm install --production'
+      posix: 'npm prune && npm install --production'
     };
     // determine the shell we're running in
-    const whichShell = shell.match(/fish/) ? 'fish' : 'default';
+    const whichShell = shell.match(/fish/) ? 'fish' : 'posix';
     // Use the install command that is appropriate for our shell
     exec(installCommands[whichShell], {
       cwd: path,


app/plugins.js
```diff
@@ -236,7 +236,15 @@ function install(fn) {
     env.npm_config_target = process.versions.electron;
     env.npm_config_disturl = 'https://atom.io/download/atom-shell';
     /* eslint-enable camelcase  */
-    exec('npm prune && npm install --production', {
+    // Shell-specific installation commands
+    const installCommands = {
+      fish: 'npm prune; and npm install --production',
+      default: 'npm prune && npm install --production'
```
Asuza: Would it be more appropriate to change this from `default` to `posix`?

knewter: I believe so, and I've just pushed a commit making that change.  Thanks!

jonathandelgado:
I did some research on this is #1166 - this is a larger issue than just fish. Rather than splitting shells up, I think the better idea would be adding an npm script to `~/.hyper-plugins/package.json` (since npm scripts would be able to handle the `&&` across multiple operating systems) and calling `npm run sync` instead.

Additionally, there are some problems with explicitly setting a shell while running these commands that need to be addressed as well.

mike-engel:
+1 for either creating an npm run script, or just using `;` since I know bash, zsh, and fish all support that syntax.

dfrankland:
Maybe a duplicate of https://github.com/zeit/hyper/pull/1065

mike-engel:
Definitely a duplicate, but with a different solution. I think an npm run-script is the most cross-platform way to do this

OJFord:
I thought this was going to be why I have the same issue using xonsh - but xonsh _does_ support `&&` (though it's not POSIX-compliant in general). Is there somewhere else that might need a fish/xonsh switch too?

matheuss:
Great solution @knewter, thank you!

PR #1218, opened 2016-12-15
Fix text deselect

### LINKED ISSUE #819: Text pasted has white background on white

ppot:
Close #819

### DIFF
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -213,19 +213,21 @@ hterm.Keyboard.prototype.onKeyDown_ = function (e) {
 
   if (e.altKey || e.metaKey || isAccelerator(e)) {
     // hterm shouldn't consume a hyper accelerator
+    // // Was the hyperCaret removed for selectAll
+    if (e.key === 'v' && !this.terminal.cursorNode_.contains(this.hyperCaret)) {
+      this.terminal.focusHyperCaret();
+    }
     return;
+  }
+
+  // Test for valid keys in order to accept clear status
+  if (e.code !== 'Control' && e.key !== 'Shift' && e.code !== 'CapsLock' && e.key !== 'Dead') {
+    selection.clear(this.terminal);
   }
 
   // Was the hyperCaret removed for selectAll
   if (!this.terminal.cursorNode_.contains(this.hyperCaret)) {
     this.terminal.focusHyperCaret();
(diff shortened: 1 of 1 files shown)


ppot:
I can't review myself. @matheuss @albinekb @rauchg

naturalethic:
This is still not fixed on MacOS at least.  First of all, e.code is `ShiftLeft`, `ControlLeft`, etc.  But even if I hack it to check these, and ensure that `selection.clear` is not being called, the selection is still being cleared by something else.

ppot:
On MacOS this work perfectly. This is not released yet. If you still have a problem. feel free to make a gif. With your specific use case.

naturalethic:
I was hacking on it in dev mode using the repo master branch.  A production build to dist doesn't work for me, it won't open a new window.  Anyhoo, I'm fine waiting for the next official release.  Thanks again.

lib/hterm.js
```diff
@@ -213,20 +213,22 @@ hterm.Keyboard.prototype.onKeyDown_ = function (e) {
 
   if (e.altKey || e.metaKey || isAccelerator(e)) {
     // hterm shouldn't consume a hyper accelerator
+    // // Was the hyperCaret removed for selectAll
+    if (e.key === 'v' && !this.terminal.cursorNode_.contains(this.hyperCaret)) {
```
matheuss: @ppot why this is here? :thinking:

Yliaho:
spans that have white background color are still obfuscated... I had to google this issue, sorry if this isn't the right place to moan about this.  
  
This fixes *all* of the white-against-white issues:  
  
`termCSS: 'span[style*="background-color: rgb(255, 255, 255);"] {color: black!important;}'`

bmherold:
> This fixes all of the white-against-white issues:
> termCSS: 'span[style*="background-color: rgb(255, 255, 255);"] {color: black!important;}'

with "backgroundColor: 'rgba(0,0,0,0.6)'," I'm still seeing white on white, even with the termCSS setting above.

ppot:
If you have issue. We recommend creating new one and linking it to the closed one.

PR #1220, opened 2016-12-15
Add a shortcut to toggle fullscreen mode. Fix #1187

### LINKED ISSUE #1187: Add ability to toggle pane fullscreen
iTerm allows this when using command-shift-enter. Probably the thing I miss the most.

evanxd:
This pull request fixes #1187 .

Add a shortcut(`Super+Enter`) to toggle fullscreen mode.

Could you help review the patch? Thanks.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -29,7 +29,7 @@ const {homedir} = require('os');
 
 // Packages
 const {parse: parseUrl} = require('url');
-const {app, BrowserWindow, shell, Menu} = require('electron');
+const {app, BrowserWindow, shell, Menu, globalShortcut} = require('electron');
 const {gitDescribe} = require('git-describe');
 const uuid = require('uuid');
 const fileUriToPath = require('file-uri-to-path');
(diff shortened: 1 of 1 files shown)


tbodt:
What's Super?

citycide:
@tbodt `super` is just a universal name for either the <kbd>Windows</kbd> key on Windows / Linux or the <kbd>Command</kbd> key on Mac.

cthedot:
Normal keyboard shortcut for Fullscreen on Windows is F11 or sometimes ALT+ENTER too. WIN+ENTER toggles narrator and guess should not be replaced.
I have no idea how it works but is it possible to have OS specific and different keyboard shortcuts in Electron/Hyper?

albinekb:
I think this should be configurable? cc: @matheuss

matheuss:
I don't think it's a good idea to set a global shortcut like that 🤔 

Other than that, that's doable via a plugin, and as a recent policy we're not merging this kind of change (simple change, doable as a plugin) into the core codebase 🙌

MEDIUM-ch:
I *do* think it's a good idea to set a global shortcut like that  ;)

More seriously, this is in the main menu so should have a Keyboard shortcut too and as it is common probably F11  (or/and maybe platform specific to Windows ALT+ENTER). A core functionality should be Keyboard accessible and be part of the core and not be implemented in a plugin.

PR #1247, opened 2016-12-16
[Windows] Add right click copy/paste

### LINKED ISSUE #1019: Right click to copy and paste
Hi, that's what we can do on powershell, it's be nice if we can do that too on hyper, and another one, can you please add a shortcut for copy and paste: like Ctrl+Shift+C/V. 

Thank you

Specro:
Adds the usual Windows behavior of copying or pasting when right clicking in the terminal. Would be nice if more people would test that this works correctly 😄 Besides that it's ready to be merged.

Closes #1019

### DIFF
--- a/app/config-default.js
+++ b/app/config-default.js
@@ -77,7 +77,11 @@ module.exports = {
     bell: 'SOUND',
 
     // if true, selected text will automatically be copied to the clipboard
-    copyOnSelect: false
+    copyOnSelect: false,
+
+    // if true, on right click selected text will be copied or pasted if no
+    // selection is present
+    quickEdit: false
 
     // URL to custom bell
     // bellSoundURL: 'http://example.com/bell.mp3',
--- a/lib/components/term-group.js
+++ b/lib/components/term-group.js
@@ -73,7 +73,8 @@ class TermGroup_ extends Component {
       onTitle: this.bind(this.props.onTitle, null, uid),
       onData: this.bind(this.props.onData, null, uid),
       onURLAbort: this.bind(this.props.onURLAbort, null, uid),
-      borderColor: this.props.borderColor
+      borderColor: this.props.borderColor,
+      quickEdit: this.props.quickEdit
     });
 
     // This will create a new ref_ function for every render,
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -1,4 +1,5 @@
 /* global Blob,URL,requestAnimationFrame */
+import {clipboard} from 'electron';
 import React from 'react';
 import Color from 'color';
 import uuid from 'uuid';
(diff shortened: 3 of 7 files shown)


kleinfreund:
Would this lead to copying with rightclick on Windows by default?

That said, is there a main issue regarding cross-platform default keybindings? The various ways different operating system and terminal programs use shortcuts is an incredible mess. Copying via <kbd>Mouse 2</kbd> (rightclick)? Or even <kbd>Shift</kbd><kbd>Ins</kbd>? Or <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>C</kbd>? Or only <kbd>Ctrl</kbd><kbd>C</kbd>? I'm sure there are even more ways.

In general, should the Hyper keybindings match the OS default? In this case, at this point I would argue that the default on Windows is <kbd>Ctrl</kbd><kbd>C</kbd>.

albinekb:
I would like this an an option. It's weird for users if Hyper acts different on based on OS...

Specro:
@kleinfreund It seems Windows cmd has a QuickEdit mode which if enabled, lets the user select text in the window and copy/paste it on right click. Without it enabled on right click a context menu is shown and then cmd actually considers <kbd>Enter</kbd> the default key for copying. 

So one more key to that mess you mentioned 😕

I think we should go with @albinekb idea of making an option for this.

ppot:
Appart form this. This type of action should be handled by hterm directly in the onMouseDown_

paulcbetts:
> Would this lead to copying with rightclick on Windows by default?

If there is a selection, right-click == Copy, if not, right-click == Paste

> In general, should the Hyper keybindings match the OS default? In this case, at this point I would argue that the default on Windows is CtrlC.

Not in a terminal, because of the obvious conflict between Ctrl-C as SIGINT. The old-school DOS shortcut Shift-Insert applies here.

QuickEdit is a deeply Windows Thing but should indeed be an option in Hyper if it wants to be a terminal on Windows (tbh, if you got used to it on macOS or Linux, I bet you'd get value out of it there too tho). I suspect it should be opt-in though.

> Also you paste on right-click, what if you want to use the context-menu? 🤔

When QuickEdit is enabled, there is no context-menu

Specro:
@paulcbetts Very well explained 👌 
@ppot Is this ok? 🤔 

I named the config option quick edit for now. If anyone has any better suggestions they're welcome to share 😄

paulcbetts:
> I would like this to be an opt-in-option. It's weird for users if Hyper acts different on based on OS...

I think that this should be enabled by default on Windows, since it is also enabled by default in the default terminal on Win10

matheuss:
Thanks @Specro 🙌

Issue #1042, opened 2016-11-25
Weird white screen issue


rainerborene:
- **OS version and name**: elementary OS 0.4 Loki
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://github.com/rainerborene/dotfiles/blob/master/.hyper.js

## Issue

Sometimes when I type a command that does not exist it shows that blank page.

![screenshot from 2016-11-25 12 57 21](https://cloud.githubusercontent.com/assets/101501/20628813/5e6b5aa4-b30f-11e6-9c77-0b06588fad47.png)

Is it possible to disable web preview stuff?

thejmazz:
heh, I also have random white screen issues (esp. when using tmux). you can hide it by doing

```javascript
store.dispatch({
  type: 'SESSION_URL_UNSET',
  uid: store.getState().sessions.activeUid
})
```

if you want to disable all web stuff,
just intercept the `SESSION_URL_SET` action in a plugin middleware and don't call `next`:

```javascript
// ~/.hyper_plugins/disableweb/index.js

exports.middleware = (store) => (next) => (action) => {
  if (action.type !== 'SESSION_URL_SET') {
    next(action)
  } else {
    console.log('Intercepted SESSION_URL_SET')
  }
}
```

then add `disableweb` to your local plugins in `~/.hyper.js`. You can use cmd+r to refresh/reload plugins while the terminal is open.

This is the same pattern I used, but the other way around, in my `imgcat` plugin:

https://github.com/thejmazz/hyper-imgcat/blob/6d9410a62f378ec2d62daf55d943968ac3f690fb/index.js#L1-L22

ppot:
@rainerborene Does your command include a URL?

rainerborene:
@ppot not

jrop:
I am running into this when entering a non-existent command within TMUX.  For example, if I enter the command "asdf", then the `isUrlCommand(shell, data) {...}` function erroneously interprets `data` as a URL:

```js
//
// Variable capture from within isUrlCommand(shell, data) {...}
//
shell = "zsh"
data = "\u001b[1;84r\u001b[H\r\nzsh: command not found: asdf\u001b[1;85r\u001b[H\u001b[2B"
matcher = "zsh: (?:(?:command not found)|(?:no such file or directory)): ((?:https?:\/\/)|(?:file:\/\/)|(?:\/\/))?([^\n]+)"
[...]

Issue #910, opened 2016-10-20
Can't use r.js (RequireJS)


markzrt:
when I write r.js or like the prompt disappears, and then displays an error
## 

HyperTerm 0.7.1
Electron 1.2.8
darwin x64 16.0.0

ekmartin:
This is because it's recognized as a website URL. We should probably have some kind of check to make sure that the URL isn't a valid command before opening it in a webview.

markzrt:
Exactly, you're absolutely right...

rauchg:
URL handling has been vastly [improved on `1.0.1`](https://twitter.com/rauchg/status/815010770618437632)! Thanks for the reports 🙏

PR #1259, opened 2016-12-17
Fix url interpretation

ppot:
Previous `url-command.js` used domainRegex
`export const domainRegex = /([0-9]+[:.]*)+|([a-zA-Z0-9.]+[.][a-zA-Z0-9]+([:][0-9]+)*){1}/`

```
    if (domainRegex.test(domain)) {
      const result = path.match(domainRegex)[0];
      return `http://${result}`;
    }
```
The forced http made command starting with 3 open in a `<webview>`

Exemple of supported url:
```
http://www.hyper.is
https://www.hyper.is
http://hyper.is/over/there?name=ferre
http://hyper.is/hello
http://www.q.hyper.is/hello
http://hyper.qc.ca/hello
http://localhost:1337/
http://hyper:1337/load?now=serve
http://localhost:1337/#/now
http://192.143.134/over/ok?there=hello
https://192.143.134:80/
```
Close : #1042 , #910, #594, #97

### DIFF
--- a/lib/actions/sessions.js
+++ b/lib/actions/sessions.js
@@ -1,5 +1,5 @@
 import rpc from '../rpc';
-import getURL from '../utils/url-command';
+import isUrl from '../utils/url-command';
 import {keys} from '../utils/object';
 import findBySession from '../utils/term-groups';
 import {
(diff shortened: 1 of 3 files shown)


rauchg:
Thank you @ppot. Looks beautiful

ppot:
@albinekb isUrl is better interpreted in that case. since we test the url. But maybe matchURL would be a better word?

Issue #1212, opened 2016-12-14
Add --help command line args


Tyriar:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Ubuntu 16.10
- **Hyper.app version**: 1.0.0

## Issue

`/opt/Hyper/hyper --help` simply runs the app, it should tell you what command line arguments are available (if any) as well as app name/version/etc.

radix:
I suppose this should be closed since a PR has been merged adding --help support

PR #1278, opened 2016-12-20
Add `--help` and `--version` command line arguments.

supersam654:
As requested in #1212, it would be really nice if Hyper recognized the `--help` command line argument. 

Seeing as Hyper doesn't support any actual arguments, I opted to just print the version, location of the config file, and a brief message explaining that there are no command line arguments.

It felt weird supporting `--help` without `--version` so `--version` and `-v` do the same thing.

I've been trying to upload a screenshot but Github isn't cooperating right now so this will have to do:

```
$ ./Hyper.exe --help

Hyper version 1.0.1
Hyper does not accept any command line arguments. Please modify the config file instead.
Hyper configuration file located at: C:\Users\Sam\.hyper.js
```

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -1,3 +1,15 @@
+// Remove the full path to electron and the name of this file from the arguments list.
+const args = process.argv.slice(2);
+// Print diagnostic information for a few arguments instead of running Hyper.
+if (args[0] === '--help' || args[0] === '-v' || args[0] === '--version') {
+  const {version} = require('./package');
+  const configLocation = process.platform === 'win32' ? process.env.userprofile + '\\.hyper.js' : '~/.hyper.js'
+  console.log('Hyper version ' + version);
+  console.log('Hyper does not accept any command line arguments. Please modify the config file instead.');
+  console.log('Hyper configuration file located at: ' + configLocation);
+  process.exit();
+}
+
 // handle startup squirrel events
 if (process.platform === 'win32') {
   // eslint-disable-next-line import/order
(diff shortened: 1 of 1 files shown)


app/index.js
```diff
@@ -1,3 +1,15 @@
+const firstArg = process.argv[1];
+// Print diagnostic information for a few arguments instead of running Hyper.
+if (firstArg === '--help' || firstArg === '-v' || firstArg === '--version') {
```
anonrig: What about?

```['--help', '-v', '--version'].indexOf(firstArg) != -1```

File: app/index.js [hunk 2]
```diff
@@ -1,3 +1,14 @@
+// Print diagnostic information for a few arguments instead of running Hyper.
+if (['--help', '-v', '--version'].indexOf(process.argv[1]) !== -1) {
+  const {version} = require('./package');
+  const configLocation = process.platform === 'win32' ? process.env.userprofile + '\\.hyper.js' : '~/.hyper.js';
+  console.log('Hyper version ' + version);
```
albinekb: use template-strings
```js
console.log(`Hyper version  ${version}`)

File: app/index.js [hunk 3]
```diff
@@ -1,3 +1,14 @@
+// Print diagnostic information for a few arguments instead of running Hyper.
+if (['--help', '-v', '--version'].indexOf(process.argv[1]) !== -1) {
```
albinekb: use [.includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) instead 🎉

matheuss:
Neat, thank you @supersam654 💯

Issue #1275, opened 2016-12-19
Change to xterm.js


mofux:
Hi, after evaluating several JS libs for terminal emulation to improve the performance of hyper, I found that https://github.com/sourcelair/xterm.js would probably better suite this project. 

- 👍 It is actively maintained.
- 👍 Used by Visual Studio Code and maintained by @Tyriar , who would probably welcome some combined efforts on fixing bugs and improving xterm.js
- 👍 It doesn't require to sign an NDA to contribute
- 👍 It doesn't isolate the terminal DOM inside an iframe, which makes it much easier to hack and style
- 👍 It has solved some of the problems we are having with hterm currently, e.g. IME and CJK character support
- 👎 It is currently being migrated to typescript, which might not be everyone's favourite flavour
- 👎 ~~It feels slightly less responsive at the moment, I think this is caused by the queue mechanism that tries to schedule big ui updates and introduces some ms lag - but it can be optimized~~

I did some quick tests replacing hterm with xterm.js and the performance seems level. I was also able to quick and dirty put in some of the features we get from hterm, e.g. changing the cursor color on bell.

If you like I can start working on pull request that replaces the current implementation with xterm.js, but I'd like to see some thumbs up for that first :neckbeard:

rauchg:
When I first tested it it had some serious problems in terms of support of terminal features and slower performance. Would be useful to revisit this now, though

mofux:
Agree, would actually be good to have some reproducible test cases to compare both against each other. Like the `vtop` or `emoji` test for non standard-width characters (which is broken in xterm and hterm) and locale specific tests, as well as tests for unicode, performance (`seq 1000000`) etc... I will try to create a couple of test files and run them to compare the output.

tbodt:
It also may fix #270.

tbodt:
Actually it won't, it has no support for cursor shapes at all. It would be a lot easier to fix #270 though.

MatthiasWinkelmann:
@mofux I recently generated the stats at https://github.com/zeit/hyper/issues/687#issuecomment-268141860. It's a really somewhat terrible ruby script but I could put it in a gist if you're interested.

It's actually a bit involved to measure output lag (from the outside) because the actual commands finish long before the output arrives.

tbodt:
https://github.com/zeit/hyper/pull/1286#issuecomment-268610797
[...]

PR #1286, opened 2016-12-21
Add support for cursor styles 5 and 6 to hterm

### LINKED ISSUE #270: Vim (neovim) cursor shape does not change (BLOCK to VLINE shape).
On the terminal itself the cursor works fine and changes to BLOCK shape to VLINE shape but inside VIM its simple does not work.

Follow my configuration for the cursor shape on my .zshrc file:

```
        vicmd)      print -n -- "\E]50;CursorShape=0\C-G";;  # block cursor
        viins|main) print -n -- "\E]50;CursorShape=1\C-G";;  # line cursor

```

I've try all different bellow configurations on VIM without success:
http://stackoverflow.com/questions/6488683/how-do-i-change-the-vim-cursor-in-insert-normal-mode
http://vim.wikia.com/wiki/Change_cursor_shape_in_different_modes

For example, I can change the shape on iTerm2/VIM with the follow setup:

```
let $NVIM_TUI_ENABLE_CURSOR_SHAPE=1
let &t_SI = "\<Esc>]50;CursorShape=1\x7"
let &t_SR = "\<Esc>]50;CursorShape=2\x7"
let &t_EI = "\<Esc>]50;CursorShape=0\x7"
```

Appreciate any help at all.

Thanks!

tbodt:
This fixes #270, cursor shape changing in Neovim.

### DIFF
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -427,5 +427,31 @@ lib.colors.hexToRGB = function (arg) {
   return arg;
 };
 
+// add support for cursor styles 5 and 6, fixes #270
+hterm.VT.CSI[' q'] = function (parseState) {
+  const arg = parseState.args[0];
+  if (arg === '0' || arg === '1') {
+    this.terminal.setCursorShape(hterm.Terminal.cursorShape.BLOCK);
+    this.terminal.setCursorBlink(true);
+  } else if (arg === '2') {
+    this.terminal.setCursorShape(hterm.Terminal.cursorShape.BLOCK);
+    this.terminal.setCursorBlink(false);
+  } else if (arg === '3') {
+    this.terminal.setCursorShape(hterm.Terminal.cursorShape.UNDERLINE);
+    this.terminal.setCursorBlink(true);
+  } else if (arg === '4') {
+    this.terminal.setCursorShape(hterm.Terminal.cursorShape.UNDERLINE);
+    this.terminal.setCursorBlink(false);
+  } else if (arg === '5') {
+    this.terminal.setCursorShape(hterm.Terminal.cursorShape.BEAM);
+    this.terminal.setCursorBlink(true);
+  } else if (arg === '6') {
+    this.terminal.setCursorShape(hterm.Terminal.cursorShape.BEAM);
+    this.terminal.setCursorBlink(false);
+  } else {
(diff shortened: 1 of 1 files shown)


colepeters:
This looks awesome, thanks for your work on this @tbodt. Can’t wait to try it out. This has been one of the things stopping me from making the full switch from iTerm to Hyper.

tbodt:
There are a number of other styling things I'd like to fix, such as underline/i-beam cursors not blinking, but the root cause is hterm. SO CAN WE PLEASE FOR THE LOVE OF GOD USE A JAVASCRIPT TERMINAL LIBRARY THAT DOES NOT REQUIRE SIGNING AN NDA TO CONTRIBUTE TO!!! #1275

babenzele:
@tbodt I can only find reference to changing the cursor permanently from within `hyper.js`. 

```javascript
module.exports = {
  cursorShape: 'BLOCK'
}
```

But do you know how one would trigger this from vim, such that it is BEAM in Insert mode and BLOCK in Normal mode?

colepeters:
@babenzele Not sure if possible in Vim, but for Neovim, drop this in your `init.vim`:

```vimL
let $NVIM_TUI_ENABLE_CURSOR_SHAPE=1	" Cursor is pipe in Insert, block in Normal
```

tbodt:
It certainly is possible in Vim:

    let &t_SI = "\<Esc>]50;CursorShape=1\x7"
    let &t_EI = "\<Esc>]50;CursorShape=0\x7"

~Theodore

vapier:
FTR, hterm has never required a NDA and never will.  an NDA doesn't even make sense when the project is open source.

tbodt:
...I think I meant CLA...I swear I saw NDA somewhere though.

vapier:
then i don't understand your complaint.  yes, hterm has a CLA, but xterm.js also has a CLA.  you have read their CONTRIBUTING.md right ?

albinekb:
Problem with hterm is that it's not on github @vapier @tbodt, makes it really hard to contribute and the effort to set it all up and learn the whole system google has just isn't worth it, that's why we were maintaining all of these hot-patches over hterm. Not an issue anymore with xterm 👌

vapier:
this was only talking about @tbodt's claim about NDA/CLA

PR #1287, opened 2016-12-21
Fix ctrl+c to close WebView 🍾 

### LINKED ISSUE #392: Can't always close browser with ctrl-c
When the built in browser has focus, I'm unable to close it with `ctrl-c`. 

I have to click the header first, which took me a while to figure out when first using it :)

ppot:
Add Keybinding
Keep caretComponent To prevent doom error.

close #392


lib/components/term.js
```diff
@@ -139,6 +132,18 @@ export default class Term extends Component {
...
+  handleFocus() {
+    // This will in turn result in `this.focus()` being
+    // called, which is unecessary.
+    // Should investigate if it matters.
+    this.props.onActive();
+  }
+
+  handleKeyDown(e) {
+    if (e.ctrlKey && e.key === 'c') {
+      this.props.onURLAbort();
```
rauchg: shouldn't we check here that there's an actual URL open?

ppot: We could. But it's enabled only if there is one oppened.

PR #1322, opened 2016-12-30
fix line height layout

dotcypress:
Before:
screen shot 2016-12-29 at 4 11 47 pm

After:
screen shot 2016-12-29 at 4 11 30 pm

Related info about [Inline-Block](https://blog.mozilla.org/webdev/2009/02/20/cross-browser-inline-block/)


strilliams:
So keen on this

limichange:
🛠
```js
    // .hyper.js
    termCSS: `
      .unicode-node {
        vertical-align: middle !important;
      }
    `,
```

rauchg:
Yep. Middle is the right answer. Thanks @limichange

PR #1326, opened 2017-01-01
Maximize / restore pane

alexfedoseev:
This PR implements maximize / restore active pane feature.

screen recording 2017-01-01 at 04 47 pm

Shortcut is identical to iTerm: `Cmd+Shift+Enter`

More details in review below.


app/menu.js
```diff
@@ -244,6 +244,18 @@ module.exports = ({createWindow, updatePlugins}) => {
             focusedWindow.rpc.emit('decrease fontSize req');
           }
         }
+      },
+      {
+        type: 'separator'
+      },
+      {
+        label: 'Maximize / Restore Pane',
```
alexfedoseev: This text should depend on the current state, but I'm not sure how to handle it here.

lib/components/split-pane.js
```diff
@@ -123,7 +142,6 @@ export default class SplitPane extends Component {
         display: 'flex',
         flex: 1,
         outline: 'none',
-        position: 'relative',
```
alexfedoseev: I removed it so each nested maximized pane is able to fill whole tab's space (not only parent group container). I could do this using `position: fixed` for maximized pane, but it overlaps tab labels.

NB! I'm not sure that it doesn't break anything. It seems that it works fine, but might be some edge cases, that I'm not aware of.

lib/components/split-pane.js
```diff
@@ -138,8 +156,7 @@ export default class SplitPane extends Component {
 
       pane: {
         flex: 1,
-        outline: 'none',
-        position: 'relative'
```
alexfedoseev: ^ Same comment.

lib/components/tabs.js
```diff
@@ -29,35 +29,35 @@ export default class Tabs extends Component {
...
+              })
+            }
+          </ul>,
+          isMac && <div
+            key="shim"
+            style={{borderColor}}
+            className={css('borderShim')}
+            />
+        ] :
+        null
```
alexfedoseev: Nothing changed here except indentation (linter).

lib/components/split-pane.js
```diff
@@ -76,28 +76,47 @@ export default class SplitPane extends Component {
...
               >
               { child }
+              { maximized &&
+                <div className={css('maximized_icon')}>
+                  <div
+                    className={css('maximized_icon_inner_dot')}
+                    style={{backgroundColor}}
+                    />
+                </div>
+              }
```
alexfedoseev: This icon indicates if pane is maximized and there're more panes are hidden (see gif above).

lib/components/tabs.js
alexfedoseev: Looks like it's deps issue. `yarn` is kinda broken or outdated on master as `yarn install` changes current `yarn.lock` and adds new one in `app` folder.

lib/actions/term-groups.js
```diff
@@ -171,3 +173,19 @@ export function exitActiveTermGroup() {
...
 }
+
+export function requestPaneMaximizedStateToggle() {
+  return (dispatch, getState) => {
+    dispatch({
+      type: TERM_GROUP_TOGGLE_MAXIMIZED,
+      effect() {
+        const {sessions, termGroups} = getState();
+        const {uid} = findBySession(termGroups, sessions.activeUid);
+        dispatch({
```
chabou: Finding the active TermGroup to maximize should be done on the reducer, I think.
Two actions for this feature seems to much no ?

Implementation of this feature should be like moveToNextPane() action, imo.

alexfedoseev: > Finding the active TermGroup to maximize should be done on the reducer, I think.

I don't have access to `state.sessions` inside `termGroups` reducer, so it's easier to use thunk for it. Also this pattern is used in many other action creators (incl `moveToNextPane`).

> Two actions for this feature seems to much no ?

I agree in general. Actually I did it this way b/c of the `effect` method, if it's not necessary here, I can refactor to single action.

> Implementation of this feature should be like moveToNextPane() action, imo.

Another way is to use `MAXIMIZE` / `RESTORE_MAXIMIZED` actions instead of `TOGGLE`, leaving it up to reviewers.

chabou: > Also this pattern is used in many other action creators (incl moveToNextPane).

You're right ! My bad.

app/menu.js
```diff
@@ -244,6 +244,18 @@ module.exports = ({createWindow, updatePlugins}) => {
...
+        accelerator: accelerators.maximizePane,
+        click(item, focusedWindow) {
+          if (focusedWindow) {
+            focusedWindow.rpc.emit('toggle pane maximized state req');
```
ppot: This is to long. Can you shorten it up? `toggle pane state` ?

lib/actions/term-groups.js
```diff
@@ -171,3 +173,19 @@ export function exitActiveTermGroup() {
     });
   };
 }
+
+export function requestPaneMaximizedStateToggle() {
```
ppot: `requestPaneMaximizedStateToggle` to `requestTogglePaneState`

app/menu.js
alexfedoseev: Done.

lib/actions/term-groups.js
alexfedoseev: Done.

alexfedoseev:
Also added prevention of move to the neighbour pane if active pane is maximized.

ppot:
@alexfedoseev is your `maximized` state supposed to be fullscreen

ppot:
Here is one bug: Opening a `webview` in `maximazed` state
screen shot 2017-01-03 at 5 37 34 pm

lib/components/split-pane.js
ppot: you need to keep this to prevent `webview` going `maximized`

egeriis:
This is the one feature I'm waiting for, until I'll be able to use Hyper. Can I do anythubg to assist to get this merged?

alexfedoseev:
@egeriis Hi! Sorry, I'm on vacation atm and haven't done anything about `webview` case. Here's my quick thoughts on this: if it's not possible to handle `webview` w/o `position: 'relative'` being set on containers, then we must switch to `position: fixed` for maximized pane, but it has an issue: when it's `fixed`—it overlaps tabs. To fix this we can use `calc` to figure out proper height or something, but tab's height is required in this case.

/cc @ppot

egeriis:
What I'm hearing is that there are core functionality blocking this from working properly?

hedefalk:
I need this to leave iTerm2 - I use it every day. Typical use case is to read the logs of some background thing that normally is kept in a very small pane.

ppot:
@alexfedoseev Currently reworking the core of the lib function. Should make this possible later zsh on.

alexfedoseev:
@ppot You mean after rework you mentioned it will be possible to handle `webview` w/o `position: 'relative'` being set on containers ([this change](https://github.com/zeit/hyper/pull/1326/files#diff-390fb2095059e1398302011147a3bdaeL126)), right?

ppot:
@alexfedoseev Yes, since Pane will be a component in it's own.

alexfedoseev:
@ppot Ok cool, waiting for this then 👍

freeall:
Any news on this one? Highly useful feature if you've ever used a pane workflow

iamstarkov:
@alexfedoseev can it be implemented as a plugin, like @chabou's hyper-pane? https://github.com/chabou/hyper-pane

alexfedoseev:
@iamstarkov I tried that at the first place, but no luck, so I implemented it in the core. TBH from my Atom experience, such core things better to have in the core, b/c otherwise those can be easily broken on the app update. I removed a dozen of broken and abandoned plugins from Atom and it's really frustrating experience when you got used to something and suddenly it doesn't work. 

We should be almost there with this PR and if the blocker is not supposed to be resolved anytime soon, I can try to find workarounds to make it work w/ webview. @ppot Can you update ETA when we can do something with this PR?

BTW How can I open this webview thing lol?

ppot:
@alexfedoseev Just type any website: such as www.hyper.is

nshoes:
Any progress on this? :]

chabou:
I add this feature to [hyper-pane](https://github.com/chabou/hyper-pane).
Feedbacks welcome!

hyper-pane-maximize

chabou:
Closing this PR.
[hyper-pane](https://github.com/chabou/hyper-pane) has this functionality now.
Feel free to try it and open issue in its own repo.

PR #1347, opened 2017-01-06
implemented support for mutliple shell definitions

AlmirKadric:
I've implemented an additional menu item in File which allows you to select other shells that may have been configured inside your config file. My use case was for Windows (Bash vs Powershell), but this would apply to macos and linux for anyone who uses more than a single shell flavour. I wanted to make this a plugin, however the work was too integrated for this to be possible (or rather so simple enough).

Possible enhancements to this PR:
 - [x] don't show menu item if shells is not defined in config
 - [ ] add possibility of accelerator shortcut definition from config file

Please let me know if any of the above or other fixes are required, and I'll jump on it as soon as I can.

P.S. PR checks are failing due to external code which I didn't touch. Seems someone pushed bad code to master...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,3 @@ npm-debug.log
 # optional dev config file and plugins directory
 .hyper.js
 .hyper_plugins
-
--- a/app/index.js
+++ b/app/index.js
@@ -49,7 +49,6 @@ const isDev = require('electron-is-dev');
 // Ours
 const AutoUpdater = require('./auto-updater');
 const toElectronBackgroundColor = require('./utils/to-electron-background-color');
-const createMenu = require('./menu');
 const createRPC = require('./rpc');
 const notify = require('./notify');
 const fetchNotifications = require('./notifications');
@@ -61,6 +60,7 @@ const config = require('./config');
 
 config.init();
 
+const createMenu = require('./menu');
 const plugins = require('./plugins');
 const Session = require('./session');
 
@@ -198,7 +198,8 @@ app.on('ready', () => installDevExtensions(isDev).then(() => {
       // If no callback is passed to createWindow,
       // a new session will be created by default.
       if (!fn) {
-        fn = win => win.rpc.emit('termgroup add req');
+        const shellOpts = options.shellOpts || {};
+        fn = win => win.rpc.emit('termgroup add req', {shellOpts});
       }
 
       // app.windowCallback is the createWindow callback
@@ -217,9 +218,9 @@ app.on('ready', () => installDevExtensions(isDev).then(() => {
       }
     });
 
-    rpc.on('new', ({rows = 40, cols = 100, cwd = process.argv[1] && isAbsolute(process.argv[1]) ? process.argv[1] : homedir(), splitDirection}) => {
-      const shell = cfg.shell;
-      const shellArgs = cfg.shellArgs && Array.from(cfg.shellArgs);
+    rpc.on('new', ({rows = 40, cols = 100, shellOpts = {}, cwd = process.argv[1] && isAbsolute(process.argv[1]) ? process.argv[1] : homedir(), splitDirection}) => {
+      const shell = shellOpts.path || cfg.shell;
+      const shellArgs = shellOpts.args || (cfg.shellArgs && Array.from(cfg.shellArgs));
 
       initSession({rows, cols, cwd, shell, shellArgs}, (uid, session) => {
         sessions.set(uid, session);
--- a/app/menu.js
+++ b/app/menu.js
@@ -1,12 +1,16 @@
+'use strict';
+
 const os = require('os');
 const path = require('path');
 const {app, shell, dialog} = require('electron');
 
 const {accelerators} = require('./accelerators');
-const {getConfigDir} = require('./config');
+const {getConfig,getConfigDir} = require('./config');
 
 const isMac = process.platform === 'darwin';
 const appName = app.getName();
+const shells = getConfig().shells || {};
+const shellKeys = Object.keys(shells);
 
 // based on and inspired by
 // https://github.com/sindresorhus/anatine/blob/master/menu.js
@@ -59,6 +63,56 @@ module.exports = ({createWindow, updatePlugins}) => {
     ]
   };
 
+  const otherShellsMenu = {
+    label: 'Other Shells',
+    submenu: []
+  };
+
+  for (const shellName of shellKeys) {
+    const shellOpts = shells[shellName];
+
+    otherShellsMenu.submenu.push({
+      label: shellName,
+      submenu: [
+        {
+          label: 'New Window',
+          click() {
+            createWindow(null, {shellOpts});
+          }
+        },
+        {
+          label: 'New Tab',
+          click(item, focusedWindow) {
+            if (focusedWindow) {
+              focusedWindow.rpc.emit('termgroup add req', {shellOpts});
+            } else {
+              createWindow(null, {shellOpts});
+            }
+          }
+        },
+        {
+          type: 'separator'
+        },
+        {
+          label: 'Split Vertically',
+          click(item, focusedWindow) {
+            if (focusedWindow) {
+              focusedWindow.rpc.emit('split request vertical', {shellOpts});
+            }
+          }
+        },
+        {
+          label: 'Split Horizontally',
+          click(item, focusedWindow) {
+            if (focusedWindow) {
+              focusedWindow.rpc.emit('split request horizontal', {shellOpts});
+            }
+          }
+        }
+      ]
+    });
+  }
+
   const shellOrFileMenu = {
     label: isMac ? 'Shell' : 'File',
     submenu: [
@@ -101,6 +155,11 @@ module.exports = ({createWindow, updatePlugins}) => {
           }
         }
       },
+      ...(
+        shellKeys.length > 0 ?
+          [{type: 'separator'}, otherShellsMenu] :
+          []
+      ),
       {
         type: 'separator'
       },
--- a/lib/actions/term-groups.js
+++ b/lib/actions/term-groups.js
@@ -12,14 +12,16 @@ import getRootGroups from '../selectors';
 import {setActiveSession, ptyExitSession, userExitSession} from './sessions';
 
 function requestSplit(direction) {
-  return () => (dispatch, getState) => {
+  return options => (dispatch, getState) => {
+    const {shellOpts} = options || {};
     const {ui} = getState();
     dispatch({
       type: SESSION_REQUEST,
       effect: () => {
         rpc.emit('new', {
           splitDirection: direction,
-          cwd: ui.cwd
+          cwd: ui.cwd,
+          shellOpts
         });
       }
     });
@@ -37,8 +39,9 @@ export function resizeTermGroup(uid, sizes) {
   };
 }
 
-export function requestTermGroup() {
+export function requestTermGroup(options) {
   return (dispatch, getState) => {
+    const {shellOpts} = options || {};
     const {ui} = getState();
     const {cols, rows, cwd} = ui;
     dispatch({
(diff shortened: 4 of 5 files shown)


.gitignore
```diff
@@ -11,3 +11,5 @@ npm-debug.log
 .hyper.js
 .hyper_plugins
 
+# IDEs
+.idea
```
albinekb: Add this to your global .gitignore

timneutkens: @AlmirKadric https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

AlmirKadric: ah good point, thanks for that 👍

.gitignore
```diff
@@ -10,4 +10,3 @@ npm-debug.log
 # optional dev config file and plugins directory
 .hyper.js
 .hyper_plugins
-
```
matheuss: Could you put that blank line back again? 😅

app/index.js
```diff
@@ -50,6 +49,7 @@ const config = require('./config');
 
 config.init();
 
+const createMenu = require('./menu');
```
matheuss: Could you move it up there where it was before? There's no need for this change 🙌

app/menu.js
```diff
@@ -1,11 +1,15 @@
+'use strict';
```
matheuss: We don't need `use strict` 😅

lib/actions/term-groups.js
```diff
@@ -12,14 +12,16 @@ import getRootGroups from '../selectors';
 import {setActiveSession, ptyExitSession, userExitSession} from './sessions';
 
 function requestSplit(direction) {
-  return () => (dispatch, getState) => {
+  return (options) => (dispatch, getState) => {
```
matheuss: Please remove the parenthesis around `options` – it's not needed for a single function argument

lib/index.js
```diff
@@ -64,16 +64,16 @@ rpc.on('session clear req', () => {
   store_.dispatch(sessionActions.clearActiveSession());
 });
 
-rpc.on('termgroup add req', () => {
-  store_.dispatch(termGroupActions.requestTermGroup());
+rpc.on('termgroup add req', (options) => {
```
matheuss: Please remove the parenthesis around `options` – it's not needed for a single function argument

lib/index.js
```diff
@@ -64,16 +64,16 @@ rpc.on('session clear req', () => {
...
+  store_.dispatch(termGroupActions.requestTermGroup(options));
```
matheuss: Same

lib/index.js
```diff
@@ -64,16 +64,16 @@ rpc.on('session clear req', () => {
...
 });
 
-rpc.on('split request horizontal', () => {
-  store_.dispatch(termGroupActions.requestHorizontalSplit());
+rpc.on('split request horizontal', (options) => {
```
matheuss: Same

lib/index.js
```diff
@@ -64,16 +64,16 @@ rpc.on('session clear req', () => {
...
+  store_.dispatch(termGroupActions.requestHorizontalSplit(options));
```
matheuss: Same

lib/index.js
```diff
@@ -64,16 +64,16 @@ rpc.on('session clear req', () => {
...
 });
 
-rpc.on('split request vertical', () => {
-  store_.dispatch(termGroupActions.requestVerticalSplit());
+rpc.on('split request vertical', (options) => {
```
matheuss: Same

lib/index.js
```diff
@@ -64,16 +64,16 @@ rpc.on('session clear req', () => {
...
+  store_.dispatch(termGroupActions.requestVerticalSplit(options));
```
matheuss: Same

matheuss:
Besides that:

- Could you please explain what I should have in my `.hyper.js`? I tried a couple of combinations for `shells`, but couldn't get it to work – the new session was always opened with my default shell
- You would need to add checks to the `ui` reducer for the possible values of `shell` – we wouldn't want people to set a empty `''` path for the shell, for example. Check [here](https://github.com/zeit/hyper/commit/325b90b61936809f545500529174108460ad5ac0#diff-7da6d687f5a3beed5725bb8324e4b1ad) for an example of how to do it
- `don't show menu item if shells is not defined in config` definitely? 😄

.gitignore
AlmirKadric: sure sorry for not noticing that XD

AlmirKadric: actually it was the other way around, the new line was missing in the repo
my IDE inserted it according to the .editorconfig

app/index.js
AlmirKadric: Actually there is a need for this change, the createMenu file now relies on configuration options, therefore must be required after the config is initialised.

app/menu.js
AlmirKadric: To use ES6 features on node properly it is required
e.g. my use of `for (const shellName of Object.keys(shells)) {` will break without it. Other than node 7 the above behaviour change with or without `use strict`. Without it `shellName` is set on the first iteration and remains constant for all other iterations. With it shellName is a constant to the scope of the loops inner block, however changes properly with each iteration.

lib/index.js
AlmirKadric: what do you mean??? that's a function call

lib/index.js
AlmirKadric: what do you mean??? that's a function call

lib/index.js
AlmirKadric: what do you mean??? that's a function call

AlmirKadric:
My `.hyper.js` shells configuration looks like this:
```
    shell: 'C:\\cygwin64\\bin\\bash.exe',

    // for setting shell arguments (i.e. for using interactive shellArgs: ['-i'])
    // by default ['--login'] will be used
    shellArgs: ['--login'],

    // for environment variables
    env: {},

    // Additional shells
    shells: {
        bash: {
            path: 'C:\\cygwin64\\bin\\bash.exe',
            args: ['--login'],
            env: {}
        },
        powershell: {
            path: 'C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe',
            args: [],
            env: {}
        }
    },
```

AlmirKadric:
@matheuss I'll fix the other suggestions as well. Caught up with something else right now, will jump on it as soon as I'm done with that.

AlmirKadric:
@matheuss after taking another look at the code, a reducer isn't required.
Currently if an empty shell path is provided it will be falsey causing it to fall back to the default shell.
If an invalid shell path is provided...well we can't really check that (unless we check if that file really exists?)
Also for the other options, empty entries are fine, if invalid options well it will break, though not sure how i can validate environment variables and shell startup arguments.

Any other suggestions or corrections to my above statements welcome.

AlmirKadric:
BUMP @matheuss @albinekb 
any movement on this?
are there any actionable items left on my side?

app/index.js
```diff
@@ -186,7 +186,8 @@ app.on('ready', () => installDevExtensions(isDev).then(() => {
       // If no callback is passed to createWindow,
       // a new session will be created by default.
       if (!fn) {
-        fn = win => win.rpc.emit('termgroup add req');
+        const shellOpts = options.shellOpts;
```
ppot: Need to test if `options.shellOpts` exist

AlmirKadric:
bump????

lib/index.js
```diff
@@ -64,16 +64,16 @@ rpc.on('session clear req', () => {
...
 
-rpc.on('termgroup add req', () => {
-  store_.dispatch(termGroupActions.requestTermGroup());
+rpc.on('termgroup add req', options => {
+  store_.dispatch(termGroupActions.requestTermGroup(options));
 });
 
-rpc.on('split request horizontal', () => {
-  store_.dispatch(termGroupActions.requestHorizontalSplit());
+rpc.on('split request horizontal',options => {
```
henrikdahl: Lack of space `rpc.on('split request horizontal', options => {` ✌️

albinekb:
Sorry @AlmirKadric have not had time to look at this, will look trough it tonight 🤘

Stanzilla:
@albinekb did you have time yet? Really would love to get this in.

lib/index.js
albinekb: @AlmirKadric

app/menu.js
```diff
@@ -61,6 +65,56 @@ module.exports = ({createWindow, updatePlugins}) => {
     ]
   };
 
+  const otherShellsMenu = {
+    label: 'Other Shells',
+    submenu: []
+  };
+
+  for (const shellName of Object.keys(shells)) {
```
albinekb: I would save `Object.keys(shells)` as a const since you're using it twice in this file? 🤔

app/menu.js
AlmirKadric: If anyone else is confused try this in node 5
```
var a = [1,2,3,4,5,6]; for (const b of a) { console.log(b); }
```
vs 
```
'use strict'; var a = [1,2,3,4,5,6]; for (const b of a) { console.log(b); }
```

Removing the `'use strict';` would essentially mean dropping support for anything less than node 6.

AlmirKadric:
Pulled from master and squashed into a single commit

app/menu.js
albinekb: yes, it's OK now since hyper is running on node 7.4 since https://github.com/zeit/hyper/pull/1848 👌

albinekb:
ping @matheuss

Hypercubed:
I just gave this a try... works well.  Needs a rebase and some error trapping (for example bad paths).

MartyGentillon:
One other minor change: this needs to be documented (probably in the default config file), otherwise people won't find the new feature.

jayce-leathers:
Is this going to be merged in anytime soon?

harangue:
Anything specifically holding this PR up? I want this badly enough that I'd take it on if @AlmirKadric doesn't have the time to move forward with it.

AlmirKadric:
@harangue from what I'm reading above, the remaining items are
 - error trapping for bad paths (though I would prefer he commented on that on the actual lines)
 - documentation in the default config file
 - rebase again if any changes are made

@albinekb @matheuss can you please confirm this is the last of the changes
I'm happy to make them, but in all fairness this PR has been sitting here for a terribly long time (6+ months)

I've actually made my own terminal multiplexer using similar technologies as a result since I couldn't wait for my changes to be merged, and branching from hyper didn't make sense as I wanted to make some fundamental design changes (would have been a complete rewrite of the underlying architecture)

cameronmaxwell:
Has this PR stalled / what it's current status? This would be an amazing feature to have on windows.

eamodio:
I would really love to see this in as well -- what is still left to be done, since this was started about a year ago?

AlmirKadric:
I'm still waiting on feedback from the core team as to finalizing what is needed to get this PR merged. I don't fancy fixing things if only to have more comments made for fixes. Once they confirm what's required, I'll updated the PR for merging.

Alternatively feel free to take a look at my own terminal multiplexer here: https://github.com/AlmirKadric-Published/exTerm-electron
Its not as advanced/clean/pretty in some regards, but well ahead in others, when compared to hyperterm. If it suits your needs but you need this or some other minor features, I'm happy to make the changes for you. Its been more than enough for my current needs, albeit one really annoying bug that I haven't had the time to fix completely. I'm more than happy to be more active on it if there is a community of users.

jordandrako:
Did this die with version 2's release?

Stanzilla:
@AlmirKadric willing to pick this up again?

AlmirKadric:
@Stanzilla unfortunately too much time has passed and I have since moved await from hyper and build my own terminal multiplexer using similar tech. The current beta version fills all my needs including the above feature. Even if I were to pick this up again, I believe enough code has changed that I would be essentially re-implementing the feature. As such I would suggest anyone who needs the feature to start again.

PR #1349, opened 2017-01-06
Prevent loosing focus after a char composition ended by Tab key. Fix #1341

### LINKED ISSUE #1341: ~ + Tab -> Focus lost (onKeyDown_ triggered twice for Tab key)

- **OS version and name**: MacOS 10.11.6
- **Hyper.app version**: Master but present in 1.0.1
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: default
- **Relevent information from devtools** _(CMD+SHIFT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: See below

## Issue
The problem is : If a "composition", beginning with `~` for example, is ended with `Tab` and not `Space`, Term loose focus (hyperCaret receive a `blur` event).

After blur event, if I press `Tab` twice, focus come back. (same as https://github.com/zeit/hyper/issues/1273)

More interesting: If I have another pane, focus goes to the other pane (without highlight menu) !! After that, If a do the same on the other pane, focus disappear, press 'tab' twice and he goes back to first pane.

I tracked down this bug and it drove me nuts...

Here is what I found:
```
hterm.js?8d82:204 onKeyDown_ KeyboardEvent {isTrusted: true, key: "Alt", code: "AltLeft", location: 1, ctrlKey: false…}
hterm.js?8d82:204 onKeyDown_ KeyboardEvent {isTrusted: true, key: "Dead", code: "KeyN", location: 0, ctrlKey: false…}
hterm.js?8d82:204 onKeyDown_ KeyboardEvent {isTrusted: true, key: "Tab", code: "Tab", location: 0, ctrlKey: false…}
index.js?bb83:6001 No definition for keyCode: 229
hterm.js?8d82:122 onTextInput_ TextEvent {isTrusted: true, data: "~", view: Window, detail: 0, which: 0…}
core.js?dfa1:97  action @ 23:57:52.611 SESSION_USER_DATA
hterm.js?8d82:204 onKeyDown_ KeyboardEvent {isTrusted: true, key: "Tab", code: "Tab", location: 0, ctrlKey: false…}
hterm.js?8d82:206 onKeyDown emitted twice
hterm.js?8d82:370 -> BLUR
```

`onKeyDown_` function is called twice with the same event (same e.timeStamp value). If preventDefault() is called on it, blur do not happen.

I tried to debug step by step BUT if I add a breakpoint in `onTextInput_`function, this bug disappear 😱

Difference I noticed:
Without breakpoint HyperCarret  `input` eventListener is called right after `SESSION_USER_DATA` redux action and before `SESSION_ADD_DATA` action.
But with a breakpoint, the call occurs after `SESSION_ADD_DATA` and `SESSION_PTY_DATA` actions

Maybe, webview think that I have `Alt` pressed with `Tab`. And in Chrome `Alt+Tab` give focus to the next Link. So Caret and Tabheader maybe ?

If someone can help me, because I'm on the brink of sending a PR with this ugly `e.preventDefault();` 😂

chabou:
When a char composition ends with a Tab key, after a first `onKeyDown_(e)` call, another one occurs with a different keycode (192) but the same timestamp and same key (Tab).
This second `onKeyDown_` event cause a blur and Term loose its focus.

Notice: A breakpoint in `onTextInput_` prevent this second event. Race condition ?

### DIFF
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -182,11 +182,22 @@ hterm.Terminal.prototype.copySelectionToClipboard = function () {
   }
 };
 
+let lastEventTimeStamp;
+let lastEventKey;
 // passthrough all the commands that are meant to control
 // hyper and not the terminal itself
 const oldKeyDown = hterm.Keyboard.prototype.onKeyDown_;
 hterm.Keyboard.prototype.onKeyDown_ = function (e) {
   const modifierKeysConf = this.terminal.modifierKeys;
+  if (e.timeStamp === lastEventTimeStamp && e.key === lastEventKey) {
+    // Event was already processed.
+    // It seems to occur after a char composition ended by Tab and cause a blur.
+    // See https://github.com/zeit/hyper/issues/1341
+    e.preventDefault();
+    return;
+  }
+  lastEventTimeStamp = e.timeStamp;
+  lastEventKey = e.key;
 
   if (e.altKey &&
       e.which !== 16 && // Ignore other modifer keys


rauchg:
Really appreciate your time debugging this @chabou

PR #1350, opened 2017-01-07
20% faster bootup time

rauchg:
This PR introduces two improvements to the production build:

- minification (small parsing perf gain)
- v8 specific heuristics for faster parse time

Overall, from my samples I've been able to reduce bootup time on my Macbook 12 by 20%


package.json
```diff
@@ -9,7 +9,7 @@
     "test": "npm run lint",
     "prepush": "npm test",
     "postinstall": "install-app-deps",
-    "pack": "npm run build && build --dir && babel --no-comments --compact --minified --out-file app/dist/bundle.js app/dist/bundle.js",
+    "pack": "npm run build && build --dir && BABEL_ENV=production babel --out-file app/dist/bundle.js --no-comments --minified app/dist/bundle.js",
```
stefanivic: @rauchg just tested this on Windows and i am getting : 
`'BABEL_ENV' is not recognized as an internal or external command, operable program or batch file.`

I added cross-env and i was able to build it like that. Do you think there is a better option for this ?

rauchg: :( not sure. @hzoo can we pass it as an option somehow?

hzoo: I think people just use cross-env for that - the env option just reads from process.env - otherwise you would just make your babelrc a js file not in package.json and do whatever you want from there?

Issue #1310, opened 2016-12-26
Can't install dependencies.


NileshPS17:
I was trying to setup the project in my linux system but i get this.
What could be the issue?

**Error: /usr/bin/nodejs exited with code 1
Error output:
npm ERR! Linux 4.4.0-45-generic
npm ERR! argv "/usr/bin/nodejs" "/usr/share/npm/bin/npm-cli.js" "install" "--production" "--cache-min" "999999999" "--build-from-source"
npm ERR! node v4.2.6
npm ERR! npm  v3.5.2

npm ERR! No compatible version found: gaze@1.1.2
npm ERR! Valid install targets:
npm ERR! 1.1.1, 1.1.0, 1.0.0, 0.6.4, 0.6.3, 0.6.2, 0.6.1, 0.6.0, 0.5.2, 0.5.1, 0.5.0, 0.4.3, 0.4.2, 0.4.1, 0.4.0, 0.3.4, 0.3.3, 0.3.2, 0.3.1, 0.3.0, 0.2.2, 0.2.1, 0.2.0, 0.1.6, 0.1.5, 0.1.4, 0.1.3, 0.1.2, 0.1.1, 0.1.0
npm ERR! 
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /home/nilesh/Projects/hyper/app/npm-debug.log

    at ChildProcess.childProcess.once.code (/home/nilesh/Projects/hyper/node_modules/electron-builder/src/util/util.ts:124:14)
    at ChildProcess.g (events.js:291:16)
    at emitTwo (events.js:106:13)
    at ChildProcess.emit (events.js:191:7)
    at maybeClose (internal/child_process.js:877:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:226:5)
From previous event:
    at spawn (/home/nilesh/Projects/hyper/node_modules/electron-builder/src/util/util.ts:87:3)
    at installDependencies (/home/nilesh/Projects/hyper/node_modules/electron-builder/src/yarn.ts:65:3)
    at /home/nilesh/Projects/hyper/node_modules/electron-builder/src/yarn.ts:11:11
    at next (native)
From previous event:
    at installOrRebuild (/home/nilesh/Projects/hyper/node_modules/electron-builder/out/yarn.js:33:21)
[...]

PR #1353, opened 2017-01-07
Nw/pty fixes

nw:
Should be good to go. An update to the`README.md` might be advised but will wait for feedback before proceeding.

* `pty.js` has been renamed to [node-pty](https://www.npmjs.com/package/node-pty) changes already made in [vscode](https://github.com/Microsoft/vscode/pull/18250/files). 
* Some of the build issues and troubleshooting are masked by [script/install](https://github.com/Tyriar/node-pty/blob/master/scripts/install.js). No handling of `stderr` also `spawn` should probably be used instead of `exec` or at the very least increase [maxBuffer](https://nodejs.org/api/child_process.html#child_process_maxbuffer_and_unicode). Going to issue a PR shortly.
* `node-gyp` builds can fail because of `xcode-select`. The following solution works great if you don't want to wait for a huge download from apple. http://stackoverflow.com/a/17980786 (`sudo xcode-select -s /Applications/Xcode.app/Contents/Developer`)
* Not a huge fan of how `install-app-deps` works. [#5 in quick setup](https://github.com/electron-userland/electron-builder#quick-setup-guide) references [node-gyp-rebuild](https://github.com/electron/electron-rebuild) which I found to be more intuitive, I refrained from adding an additional dep for now. Example: `electron-rebuild -m app -w node-pty`.
* Added `npm run clean` to reset everything for sanity checking.

cc: @rauchg


rauchg:
> node-gyp builds can fail because of xcode-select. The following solution works great if you don't want to wait for a huge download from apple. http://stackoverflow.com/a/17980786 (sudo xcode-select -s /Applications/Xcode.app/Contents/Developer)

We should mention this on the README in the pty section

Tyriar:
@nw you beat me to it! :smile: 

@rauchg FYI I gave up on doing this change because I couldn't build due to https://github.com/zeit/hyper/issues/1310

rauchg:
@Tyriar strange, yet it works on Travis CI (Linux)? ¯\_(ツ)_/¯

Tyriar:
Not sure why, it happened on both my Ubuntu 16.10 machines. I didn't try yarn though

Issue #100, opened 2016-07-15
Ugly Scrollbar


bnb:
Just downloaded and installed, noticed the scroll bar isn't being styled or modified as I would have expected and have seen in previous releases...

<img width="749" alt="screen shot 2016-07-15 at 3 47 24 pm" src="https://cloud.githubusercontent.com/assets/502396/16886699/d7b9384c-4aa3-11e6-9094-f6bd11ff07b1.png">

On hovering, moving from right to left (NOT left to right), of a very small vertical region on the right side of Hyperterm. Normal without hovering.

kinduff:
I'm able to reproduce following the next steps:

1. Open new terminal
2. Hover where the scroll bar is supposed to appear
3. Press Enter key to add a new line until the scroll bar appears
4. Scrollbar should look thicker and uglier than the regular one
5. Hover out and scroll up and the regular scrollbar should appear

josebalius:
Yeah I am seeing the same thing. Something like Visual Code's scrollbar would be nice :)

rauchg:
I can't reproduce this O.o

rauchg:
😿

bnb:
@rauchg Is there anything I can give you to help find it?

raul:
I can reproduce as well. As a stopgap I've added  `::-webkit-scrollbar { display: none; }` to the `termCSS` config.

amilajack:
I can also reproduce this.

MrRio:
This looks like the regular scrollbar to me:

```
::-webkit-scrollbar {
  background: transparent;
  width: 6px;
}
::-webkit-scrollbar-thumb {
  border-radius: 5px;
  background: grey;
}
```

PR #1354, opened 2017-01-07
Scrollbar Fix on Mac. Fixes issue #100

micopc:
Fix Issue #100 

The issue was happening when "Show scroll bars" was set to "Always" on the user's preferences aside from the referenced issue.

The only problem I had was that some linting was not passing, but I didn't edit that file.
screen shot 2017-01-07 at 11 42 45 am


rauchg:
I believe I fixed that in another PR

PR #1356, opened 2017-01-07
fix travis builds

nw:
I'm running into discrepancies with local tests vs travis builds. Can see in #1353 and saw it mentioned in #1354.

The culprit is the lack of a `install` in the `.travis.yml` This is a result of the auto-magic of the install when travis sees a `yarn.lock` file. [More info here](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Travis-CI-supports-yarn)


rauchg:
Fascinating. That said, `yarn` worked for me locally. Why can't we use it in CI? Thanks @nw

nw:
See: https://github.com/electron/electron/pull/8229#issuecomment-268149167

PR #1359, opened 2017-01-08
Pane jumping with `^+⌘+1-9` shortcut.

### LINKED ISSUE #1295: Jumping to split panes & tabs through numeric mappings
- [Y] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [Y] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

## Feature Request
screen shot 2016-12-22 at 6 38 26 am

I've traditionally used iTerm on Mac, and now I'm switching to Hyper. I am aware of the current shortcut keys to switch panes/tabs in Hyper, but there's a feature that I believe would make work much faster for Hyper users. 

If you look at the screenshot, you will see designated shortcuts on every split pane as well as every tab with a number - so Cmd+2 takes you straight to tab 2, and Alt+3 takes you straight to pane 3. 

This makes it much easier and faster to work than the current scheme, especially when you have a large number of tabs and split panes open. The numbers save you the effort of thinking how many times to press the cycle button. It enables the power of multiple split panes, especially if one is using a large screen.

So, I can reach any tab and any split pane in that tab with O(1) taps of my keyboard (max 4, if I want to switch to a different tab and different split pane, usually 2)

On the other hand, with the current shortcut I need to go cycling around and pressing Ctrl+Alt+Tab repeatedly until I reach my pane/tab. This means that if I have X tabs open and Y panes open, in the worst case I need to press a shortcut of 3 buttons (X-1)+(Y/2) times. And I often have 7-8 tabs with around 5 split panes on every tab. This is especially more cumbersome if I have split the panes in an unnatural order - they don't cycle in a clockwise or anticlockwise way.

I'm sure many users already love Hyper, and would love it even more if it gets something like what is there in the screenshot - life is too short to spend cycling through tabs and panes.

chabou:
This PR implements #1295

I've chosen to use `^+⌘+1-9` to jump to a pane because ``⌘+1-9` is used to jump to a tab.
Shortcut appears on top left of each term (in margin) from 2 panes in tab.

capture d ecran 2017-01-08 a 16 05 15

Possible improvements:
- Make shortcut appears only when `^+⌘` keys are down
- Choose an another shortcut?


lib/actions/ui.js
```diff
@@ -88,21 +89,6 @@ export function setFontSmoothing() {
...
-    return [uid];
-  }
-
-  return group
-    .children
-    .reduce((total, childUid) => total.concat(
-      findChildSessions(termGroups, childUid)
-    ), []);
-};
-
```
chabou: Move to `lib/utils/term-groups.js`because now used in `lib/selectors.js` too

lib/selectors.js
```diff
@@ -1,11 +1,20 @@
...
+  termGroups => getRootGroups_(termGroups)
 );
 
-export default getRootGroups;
+// For each rootGroup, it sorts its children
+export const getSortedSessionGroups = createSelector(
+  getTermGroups,
+  termGroups => getRootGroups_(termGroups).reduce((result, {uid}) =>
+    Object.assign(result, {[uid]: findChildSessions(termGroups, uid)}), {})
+);
```
chabou: digit shortcut for each term are not in the store because it is possible to derive them from the global state. It computes children list for each rootGroup (Tab)

lib/reducers/term-groups.js
```diff
@@ -23,16 +23,6 @@ function TermGroup(obj) {
...
-// Recurse upwards until we find a root term group (no parent).
-const findRootGroup = (termGroups, uid) => {
-  const current = termGroups[uid];
-  if (!current.parentUid) {
-    return current;
-  }
-
-  return findRootGroup(termGroups, current.parentUid);
-};
-
```
chabou: This function should not be moved because used only here

lib/components/terms.js
```diff
@@ -84,6 +84,7 @@ export default class Terms extends Component {
           const props = getTermGroupProps(uid, this.props, {
             termGroup,
             terms: this.terms,
+            sortedSessionGroups: this.props.sortedSessionGroups[uid],
```
chabou: Extract sorted children which concern this tab

lib/actions/ui.js
```diff
@@ -202,6 +188,31 @@ export function moveTo(i) {
...
+export function moveToPane(i) {
+  return (dispatch, getState) => {
+    dispatch({
+      type: UI_MOVE_TO_PANE,
+      index: i,
+      effect() {
+        const state = getState();
+        const rootGroupUid = state.termGroups.activeRootGroup;
+        const sortedSessionGroups = findChildSessions(state.termGroups.termGroups, rootGroupUid);
+        const uid = i > sortedSessionGroups.length ? null : (i === 9 ? sortedSessionGroups[sortedSessionGroups.length - 1] : sortedSessionGroups[i - 1]);
```
chabou: Find right pane for digit shortcut pressed

chabou:
Is it better to tranform this PR to a plugin ?

matheuss:
@chabou I think it might be better to make it a plugin 🤔 We're trying to `If something can be done as a plugin, it should be a plugin instead of a core Hyper feature`. Will confirm with @rauchg and get back to you ASAP 🙌

rauchg:
What I'd like to have is a very *simple* way for plugins to configure new keymappings. This would be trivial to write, in that case. So, when we merge the keymap feature we need to keep in mind that it should be easy for @chabou to extend that config from his plugin to accomplish this.

Thoughts?

ppot:
@chabou Can you point me to the pane numbering order?

chabou:
@ppot I made a selector: https://github.com/chabou/hyper/blob/pane_jumping/lib/selectors.js#L16-L20
It use existing `findChildSession()` function (moved in utils/term-groups.js: https://github.com/chabou/hyper/blob/pane_jumping/lib/utils/term-groups.js#L8-L21) used for next/previous pane calculation. It was the same order than iTerm2: "first child descendent".
I needed to add a special meaning of 9 number which is "the last":
here https://github.com/chabou/hyper/blob/pane_jumping/lib/components/term-group.js#L52
and here https://github.com/chabou/hyper/blob/pane_jumping/lib/actions/ui.js#L209

I don't know yet if your work about Session Restoring (and state refactoring) will impact this.

ppot:
@chabou I was saying the docs about the order of the pane.

ppot:
@chabou I will supply pane numing. So you will be able to navigate with pane number!

chabou:
I will adapt my plugin to use your work ;)

chabou:
@rauchg my plugin use mousetrap for keybindings inspired by https://github.com/iamstarkov/hyper-panes-iterm2-hotkeys. I think it is a better way to achieve this than hacking hterm like this plugin: https://github.com/gyandeeps/hypernpm

It would be really great if, in next keymap feature, handlers registered by a plugin for a key combination will receive `dispatch()` function directly.

PR #1366, opened 2017-01-09
Fix decoding race condition with new terminal sessions

rauchg:
See the comment in the source code for details


rauchg:
Was creating the following issue (see my prompt)
image

rauchg:
vs 
image

Issue #1332, opened 2017-01-03
unicode-node css style


enguerran:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: MacOS Sierra 10.12.2
- **Hyper.app version**: 1.0.1.1369
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/enguerran/c7c56adcaca07b9679daad7328d8bbfd
- **Relevent information from devtools** _(CMD+SHIFT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: N/A
- **The issue is reproducible in vanilla Hyper.app**: [zsh](http://www.zsh.org/) + [oh-my-zsh](https://github.com/robbyrussell/oh-my-zsh) + [bullet-train theme](https://github.com/caiogondim/bullet-train-oh-my-zsh-theme)

## Issue

I have a `unicode-node` style issue: on the left, the actual css rule with `unicode-node { display: inline-block; }`, on the right, a modified css rule with `unicode-node { display: inline; }`.
My custom emojis (goat + gust of wind) are better aligned on the right.

Do you think it is a issue on hyper or may it be a issue on another element in my terminal stack (MacOS + hyper + zsh + oh-my-zsh + bullet-train + Powerline font)?

<img width="1440" alt="screen shot 2017-01-03 at 16 33 06" src="https://cloud.githubusercontent.com/assets/701648/21612808/e4240f56-d1d2-11e6-898d-227c22602384.png">

tkisme:
same result,and unicode-node { display: inline; } works

storvik:
Same goes for me, but.. 

Changing `inline-block` to `inline` fixes icons, but the cursor shifts (due to icon widths?). 
<img width="483" alt="screen shot 2017-01-06 at 10 53 57" src="https://cloud.githubusercontent.com/assets/3438604/21714713/ba71fbe8-d400-11e6-9752-58835b179a66.png">

Turning on newline in the prompt makes it bearable. This also makes tmux usable again, see my comment here #1156. As long as you dont split the window vertically 
<img width="958" alt="screen shot 2017-01-06 at 11 09 41" src="https://cloud.githubusercontent.com/assets/3438604/21714888/9979b16e-d401-11e6-9acd-a478c370ecf6.png">

[...]

Issue #1352, opened 2017-01-07
Powerline symbols not rendering correctly


BirkhoffLee:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: macOS Sierra 10.12.2 (16C67)
- **Hyper.app version**: Version 1.0.1 (1.0.1.1369)
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/BirkhoffLee/8a6a812de8395ed4bc0662e9bd8ea424
- **Relevent information from devtools** _(CMD+SHIFT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: N
/A
- **The issue is reproducible in vanilla Hyper.app**: Yes

## Issue

This explains everything. Look at the branch symbol.  
![screen shot 2017-01-07 at 8 02 42 pm](https://cloud.githubusercontent.com/assets/8148095/21741543/4662cd6c-d514-11e6-81ae-7168c1537383.png)

wesbos:
Same issue reported over at 

https://github.com/zeit/hyper/issues/1336

Issue #1227, opened 2016-12-15
Adding padding to x-row breaks cursor position in terminal


manuelbieh:
- [✅] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [✅] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 10, latest stable build
- **Hyper.app version**: 1.0.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: `    termCSS: 'x-row { padding-top: 4px;}',`
- **Relevent information from devtools**: http://imgur.com/a/TooMS
- **The issue is reproducible in vanilla Hyper.app**: ?

## Issue
I added some padding to each line in the term window but now the cursor position is totally broken. With each linebreak the difference between actual cursor position and the position the cursor should be at is increased by the amount of the padding. 3 new lines = 12px, 10 new lines = 40px, etc ...

AaronFlower:
I face the same problem!

spoike:
This issue has been earlier found in #75

timothyis:
`termCSS` is deprecated and `x-row` no longer exists due to the removal of hterm in favour of xterm.js

PR #1376, opened 2017-01-10
Fix display of unicode characters

### LINKED ISSUE #1336: Unicode characters are crowded together, display unreadable
No matter what font I choose, hyper Chinese it is displayed, is crowded with。

- **OS version and name**: Ubuntu 16.04 x86_64

- **Hyper.app version**: Hyper 1.0.1
- **Link of a [.hyper.js](https://gist.github.com/fcharlie/c0f00aa0194d858a96a0f7a19d3688cb) with the contents of your .hyper.js**: 
- **Relevent information from devtools** _(CMD+SHIFT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: 
- **The issue is reproducible in vanilla Hyper.app**:

tibotiber:
Hi!

This fixes #1336 to the best of my knowledge (I'm new to this code base). The PR is ready to be merged.

Thanks.

before:
screen shot 2017-01-10 at 11 14 54 pm

after:
screen shot 2017-01-11 at 1 40 00 am

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -260,6 +260,9 @@ export default class Term extends Component {
       .cursor-node[focus="false"] {
          border-width: 1px !important;
       }
+      x-row {
+        line-height: 1.2em;
+      }
       ${hyperCaret}
       ${osSpecificCss}
       ${css}
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -145,7 +145,7 @@ hterm.Terminal.prototype.setFontSize = function (px) {
   }
   unicodeNodeStyle.innerHTML = `
     .unicode-node {
-      display: inline-block;
+      display: inline;
       vertical-align: top;
       width: ${this.scrollPort_.characterSize.width}px;
     }
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -145,9 +145,8 @@ hterm.Terminal.prototype.setFontSize = function (px) {
   }
   unicodeNodeStyle.innerHTML = `
     .unicode-node {
-      display: inline;
+      display: inline-block;
       vertical-align: top;
-      width: ${this.scrollPort_.characterSize.width}px;
     }
   `;
 };


ohc192:
This is the same fix I suggested in #1332. Are there any ramifications of making this change to the CSS?

tibotiber:
Thanks for the feedback @ohc192, so this PR might close #1332, #1336, and #1352

tibotiber:
It's probably related to #1322 as well, @dotcypress any comment on possible side effects here?

BirkhoffLee:
I'll take a look later and report.

BirkhoffLee:
The cursor isn't in desired position, and Chinese characters can't display normally.

screen shot 2017-01-11 at 5 21 42 pm

tibotiber:
Hi @BirkhoffLee, thanks for testing. Chinese characters indeed get the additional classe `wc-node` with its own `display: inline-block` so I had to explicitly `display: inline` on the `unicode-node` class. Which is basically the fix proposed by @ohc192 👍. It seems to work now. 
screen shot 2017-01-11 at 8 24 02 pm

About the cursor position, I am not able to reproduce here. Can you please test again?

BirkhoffLee:
@tibotiber Still not working

screen shot 2017-01-11 at 9 32 25 pm

```
module.exports = {
  config: {
    // default font size in pixels for all tabs
    fontSize: 13,

    fontFamily: '"Meslo LG L for Powerline", Menlo, "DejaVu Sans Mono", "Lucida Console", monospace',

    // terminal cursor background color and opacity (hex, rgb, hsl, hsv, hwb or cmyk)
    cursorColor: '#ffffff',

    // `BEAM` for |, `UNDERLINE` for _, `BLOCK` for █
    cursorShape: 'BLOCK',

    // color of the text
    foregroundColor: '#fff',

    // terminal background color
    backgroundColor: '#000',

    // border color (window, tabs)
    borderColor: '#333',

    // custom css to embed in the main window
    css: '',

    termCSS: '',

    // set to `true` if you're using a Linux set up
    // that doesn't shows native menus
    // default: `false` on Linux, `true` on Windows (ignored on macOS)
    showHamburgerMenu: '',

    // set to `false` if you want to hide the minimize, maximize and close buttons
    // additionally, set to `'left'` if you want them on the left, like in Ubuntu
    // default: `true` on windows and Linux (ignored on macOS)
    showWindowControls: '',

    // custom padding (css format, i.e.: `top right bottom left`)
    padding: '12px 14px',

    // the full list. if you're going to provide the full color palette,
    // including the 6 x 6 color cubes and the grayscale map, just provide
    // an array here instead of a color map object
    colors: {
      black: '#000000',
      red: '#ff0000',
      green: '#33ff00',
      yellow: '#ffff00',
      blue: '#0066ff',
      magenta: '#cc00ff',
      cyan: '#00ffff',
      white: '#d0d0d0',
      lightBlack: '#808080',
      lightRed: '#ff0000',
      lightGreen: '#33ff00',
      lightYellow: '#ffff00',
      lightBlue: '#0066ff',
      lightMagenta: '#cc00ff',
      lightCyan: '#00ffff',
      lightWhite: '#ffffff'
    },

    shell: '',

    // for setting shell arguments (i.e. for using interactive shellArgs: ['-i'])
    // by default ['--login'] will be used
    shellArgs: ['--login'],

    // for environment variables
    env: {},

    // set to false for no bell
    bell: false,

    // if true, selected text will automatically be copied to the clipboard
    copyOnSelect: false

    // URL to custom bell

    // for advanced config flags please refer to https://hyper.is/#cfg
  },

  // a list of plugins to fetch and install from npm
  // format: [@org/]project[#version]
  // examples:
  //   `hyperpower`
  //   `@company/project`
  //   `project#1.0.1`
  plugins: [
    // 'hyper-solarized-dark',
    'hyperterm-atom-dark',
    'hyperlinks',
    'hyperline',
    "hyperterm-cursor",
    "hyperterm-1password",
    // "hyperterm-sync-settings",
    // "hyperterm-paste",
    "hyperterm-alternatescroll",
    // "hyperfull",
    "hyper-tab-icons"
  ],

  // in development, you can create a directory under
  // `~/.hyper_plugins/local/` and include it here
  localPlugins: [],
  //padding: '12px 14px 18px 14px'
};
```

tibotiber:
@BirkhoffLee, 2 questions: 
1. When in the line do you start having a cursor position issue? before the end of your powerline or after? Also do you have it on other lines as well (like if you `cat` or `echo` something) or only the powerline?
2. Can you reproduce this with vanilla Hyper?

BirkhoffLee:
@tibotiber 1. I don't understand what you mean, but here's the screenplay (it take time to load, 8.2mb):

capto_capture 2017-01-11_10-00-29_pm

2. Yes.

tibotiber:
@BirkhoffLee, I looked a bit more into your issues and I would say there are different issues here! I suggest the following breakdown:

- The first issue, which we all agreed on, is that unicode wasn't displayed properly. It needs to be displayed with adaptive width. This is a simple display issue at the character level and this PR fixes it I believe.
- The second issue is that once you display unicode properly, the cursor location should also be adapted accordingly. This means that the cursor's `left` attribute should account for all the unicode characters in the row. The cursor's `width` should also match the width of the character below it. This is a much more complex issue and I have no idea how to approach it. The only open issue on cursor location I have found is #1227 and it is not exactly the same issue. They may however be adressed together.
- The third issue is about `cat` somehow displaying more Chinese characters than it should (?). I am not able to reproduce this. Maybe you can send me the string that lets you reproduce this bug. I have the feeling it could come from the way unicode characters can be "combined" to create more, and the way it is handled in JS. You can read more about this [here](https://www.contentful.com/blog/2016/12/06/unicode-javascript-and-the-emoji-family/).

As each issue is very different and related to different parts of the code, I would suggest splitting them into more actionable pieces? Any maintainer to recommend on this?

rauchg:
```
The second issue is that once you display unicode properly, the cursor location should also be adapted accordingly. This means that the cursor's left attribute should account for all the unicode characters in the row. The cursor's width should also match the width of the character below it. This is a much more complex issue and I have no idea how to approach it. The only open issue on cursor location I have found is #1227 and it is not exactly the same issue. They may however be adressed together.
```

A way to fix this could be to make all the chars equal `width`, let them show their `overflow`, and _inside_ have a span with the actual `width`? Random theory, not sure if it'd actually work

tibotiber:
Thanks for the idea @rauchg. I’ve tried to look into testing that, which led me into the chromium hterm code. Wide characters support exists there since [this commit](https://chromium.googlesource.com/apps/libapps/+/48f05cb06712f89592c63993243f58167dda8f5c%5E%21/#F3) but we override in part the wide characters support in `lib/term.js`. This has been changed as of `1.0.1` in #1111 (41ded83). A few things on this commit, maybe @dotcypress can help with feedback on the motivations:
- [this change](https://github.com/zeit/hyper/pull/1111/commits/41ded83174fea97ad5985049b8f0d0683b687726#diff-268710614c2274f611be667a5f72bb13L235) removes the `line-height` in the terminal, which is actually the true reason why powerline fonts don’t display properly anymore. I suggest putting it back and leaving `display: inline-block` on `unicode-node`. This is how hterm handles wide unicode characters, it’s probably better to keep the same (?).
- Once you do this, wide characters aren’t displayed properly anymore, they are crowded on the fix width imposed [here](https://github.com/zeit/hyper/pull/1111/commits/41ded83174fea97ad5985049b8f0d0683b687726#diff-d633f176ea335af45d2682a773898091R149). This is because the width of `unicode-node` overrides the width of `wc-node`. I suggest removing that line, which lets wide characters display properly.

The suggested changes fix the display of powerline and wide characters in a better way (imo). It also fixes the cursor issue for CJK characters (the wide characters supported by chromium’s hterm) because hterm keeps track of these characters internally to compute the position of the cursor. However all wide characters (emoji for example) are not supported by hterm. I see that @dotcypress has pushed a fix in #1018 which should override `lib.wc.strWidth`, the internal method used by hterm to detect wide characters. I’ve tested the new method, works great, but somehow it seems that hterm doesn’t use it (?). Is it due to the way hterm imports `lib.wc` on its own? 🤔 If we could ensure hterm detects emojis as wide characters, we might be off with most of the issue I think.

@BirkhoffLee, can you send the Chinese characters that don’t display properly in your screencast?

BirkhoffLee:
@tibotiber The Chinese characters I used: `測試`

rauchg:
@tibotiber your reasoning sounds very solid, and I'm inclined to merge this. Thank you for your work. I'll await on confirmation by @dotcypress who's looked into this problem quite a bit already

tibotiber:
@dotcypress: any feedback on this PR? Would love to merge and fix display issues.

slajax:
@dotcypress - friendly ping, I would really like to have this too.

matheuss:
Amazing, you @tibotiber and sorry for the late merge 😅

wesbos:
Did this make it into the last release? I just upgraded hoping it would fix but I'm still seeing it:

tibotiber:
@wesbos it's not in the last release yet. Also the line-height issue is fine but the unicode part is being reverted and needs rework due to incompatibility with braille characters used in `vtop`.

rauchg:
@wesbos the next release will fix the line-height, and maybe we can find a solution to the width issue ;)

wesbos:
Thanks yall ❤️

drorata:
@radarhere when is the next release planned?

radarhere:
@drorata I believe you meant @rauchg

drorata:
What should I do for this issue to be fixed on my machine? I still get badly rendered symbols....

wesbos:
Are you on the latest Hyper? I think it's fixed for me... fingers crossed.

drorata:
@wesbos I'm using version 1.3.1.1628. Is it the latest?

tibotiber:
@drorata there are still some issues with unicode affecting powerline among others. Working on it in #1536.

PR #1377, opened 2017-01-10
Foolproof showHamburgerMenu

ryzokuken:
Foolproof `showHamburgerMenu` by:
1. Allowing `'true'` and `'false'` as possible values apart from the original `true` and `false`.
2. Rendering the hamburger menu if `showHamburgerMenu` is either `true` or `'true'`

### DIFF
--- a/lib/components/header.js
+++ b/lib/components/header.js
@@ -145,7 +145,7 @@ export default class Header extends Component {
           style={{borderColor}}
           >
           {
-            hambMenu &&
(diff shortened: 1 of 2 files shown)


albinekb:
I don't like this, It's better to be clear in the docs and in the default config that it should be a boolean, `true` of `false`

rauchg:
Agreed. Thanks tho!

PR #1379, opened 2017-01-10
Make default config clearer

albinekb:
- Fixed a small typo (... that doesn't *shows* ...)
- Make it clear that it should be without backticks


timneutkens:
:fire:

albinekb:
related to #1377

PR #1383, opened 2017-01-11
On Win32, write out default .hyper.js as CRLF so Notepad doesn't have a fit

paulcbetts:
Shouldn't have any adverse effects and will make Windows users' first experience a bit nicer

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -84,6 +84,9 @@ function exec(str) {
   return true;
 }
 
+// This method will take text formatted as Unix line endings and transform it
+// to text formatted with DOS line endings. We do this because the default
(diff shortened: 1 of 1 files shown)


app/config.js
```diff
@@ -84,6 +84,10 @@ function exec(str) {
   return true;
 }
 
+function crlfify(str) {
```
rauchg: can we add a comment on what this does and why it's here?

paulcbetts: :+1:

matheuss:
Friendly ping @paulcbetts 🙌

app/config.js
rauchg: @paulcbetts 👍

matheuss:
Thank you @paulcbetts 🙆

PR #1409, opened 2017-01-15
Unit tests for to-electron-background-color

kandros:
As I'm learning the codebase, I'll add some unit test to have a better overall coverage, haven't found much already existing so please let me know if something should be done differently for future PRs.


rauchg:
Thank you!

PR #1412, opened 2017-01-15
Add the cwd to the redux state

### LINKED ISSUE #1357: Add cwd to redux state
Should be added to the current session PID `session.session[pid].cwd`. There is already some code in place for getting the `cwd` from hterm. For example the path in hyper's title.

Would allow plugins like https://github.com/hharnisc/hypercwd to not execute a command, which will be faster.
Will also allow for cross platform consistency (hypercwd doesn't run on windows).

matheuss:
Closes #1357

### DIFF
--- a/lib/reducers/sessions.js
+++ b/lib/reducers/sessions.js
@@ -10,7 +10,8 @@ import {
   SESSION_URL_SET,
   SESSION_URL_UNSET,
   SESSION_RESIZE,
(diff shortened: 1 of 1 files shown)


timneutkens:
Is it really that simple 0.o. Nice!

rauchg:
The little things that change games

rauchg:
cc @hharnisc

PR #1442, opened 2017-01-22
[UI] Make scrollbar look the same on all platforms

stefanivic:
Linux suffers from the same issue as other platforms where the scrollbar is that ugly electron style. 
This commit will fix this and make it look the same on all platforms and cut down code.

How it looks by default.

How it looks after this commit.

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -232,9 +232,8 @@ export default class Term extends Component {
         font-size: ${this.props.fontSize}px;
       }
     `;
-    let osSpecificCss = '';
-    if (process.platform === 'win32') {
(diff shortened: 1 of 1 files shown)


stefanivic:
@timneutkens i do not own a mac so i was unable to test this. If you or @matheuss could test it it would be great. But the css is very similar to the one used for `darwin` so i don't see why it would not work. :D

chabou:
Current scrollbar on MacOS
hyper-scrollbar-old

Scrollbar with this PR on MacOS
hyper-scrollbar-new


Really better ! Great job !

timneutkens:
@chabou thanks for testing! :heart:

matheuss:
amazing

PR #1469, opened 2017-01-31
:memo: Update Windows installation instructions

felixrieseberg:
Hey :wave:! Not sure if you're interested, but you can just use `windows-build-tools`, which will make the whole installation process a lot easier :wink:


srimajji:
agreed but it is nice to list both install processes. Similar to the windows related instructions in [node-gyp readme](https://github.com/nodejs/node-gyp)

matheuss:
That's awesome, thanks @felixrieseberg 🙌

PR #1473, opened 2017-02-01
Restore a behavior to open preference macOS

ringohub:
Restore a behavior to open preference with `EDITOR` using `cmd+,` at the macOS.
It was disabled at #1138.

I can't open the preference smoothly on macOS.
I used to open `~/.hyper.js` via `$EDITOR` when user `Cmd + ,` hotkey on macOS.
However, I can't open the preference using this hotkey.

hyper_open_finder_no_dots

This behavior is specification or rollback?
If it is new specification then please close this PR.

### DIFF
--- a/lib/actions/ui.js
+++ b/lib/actions/ui.js
@@ -212,12 +212,33 @@ export function moveTo(i) {
 }
 
 export function showPreferences() {
+  const isWin = process.platform === 'win32';
+  // eslint-disable-next-line no-template-curly-in-string
+  const command = isWin ? ' start notepad "%userprofile%\\.hyper.js"' : ' bash -c \'exec env ${EDITOR:=nano} ~/.hyper.js\'';
+  const message = [];
+  message.push(isWin ?
+    ' echo Attempting to open ^%userprofile^%\\.hyper.js with notepad' :
+    ' echo Attempting to open ~/.hyper.js with your \\$EDITOR');
+  message.push(' echo If it fails, open it manually with your favorite editor!');
+
   return dispatch => {
     dispatch({
       type: UI_SHOW_PREFERENCES,
       effect() {
-        dispatch(
-          shell.showItemInFolder(path.join(os.homedir(), '.hyper.js')));
+        if (isWin) {
+          dispatch(
+            shell.showItemInFolder(path.join(os.homedir(), '.hyper.js')));
+        } else {
+          dispatch(requestSession());
+          // Replace this hack with an async action
+          rpc.once('session add', ({uid}) => {
+            rpc.once('session data', () => {
+              dispatch(sendSessionData(
+                uid, [...message, command, ''].join(os.EOL)
+              ));
+            });
+          });
+        }
       }
     });
   };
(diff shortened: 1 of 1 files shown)


citycide:
I can see the issue you're having but this PR destroys all the work that went into #1138, which was a definite improvement for Windows.

Instead if there's an issue just with Mac OS that should be fixed separately.

lib/actions/ui.js
```diff
@@ -212,12 +212,33 @@ export function moveTo(i) {
...
+    ' echo Attempting to open ~/.hyper.js with your \\$EDITOR');
+  message.push(' echo If it fails, open it manually with your favorite editor!');
+
   return dispatch => {
     dispatch({
       type: UI_SHOW_PREFERENCES,
       effect() {
-        dispatch(
-          shell.showItemInFolder(path.join(os.homedir(), '.hyper.js')));
+        if (isWin) {
```
ringohub: I tried fix for #1138 and macOS.

ringohub:
And, I have same problem when displayed dots files :cry:.

hyper_open_finder_show_dots

ringohub:
> I can see the issue you're having but this PR destroys all the work that went into #1138, which was a definite improvement for Windows.
I cared to #1138 using ``if statement''. However, I have not windows environment. Just I try to rent windows machine from my friends and verify it 🙂.

Thank you for your review!

citycide:
Thanks for the last gif, since that clears up that `.hyper.js` is there, but sorry I'm still not sure I understand. Hyper opens the directory and you'd be able to open the config file in your editor from there, right? That was the behavior that was decided on.

lib/actions/ui.js
citycide: If this does go further then, you'd be able to remove all the stuff you re-added at the top of this function having to do with the message & command on Windows. I'll put a note on it.

ringohub:
> Hyper opens the directory and you'd be able to open the config file in your editor from there, right? 

That behavior is open the window via Finder.app and to select `.hyper.js` only. It does not opening an editor.  After that, I push the `cmd+O` then open `.hyper.js` with an editor.

lib/actions/ui.js
ringohub: Oh, good point. I'm going to fix it.

citycide:
That's the expected behavior actually. It makes all platforms have the same results and there's no guarantees about what editors people have - especially on Windows. Editing that file in notepad was a pain :laughing:

ringohub:
Oh... 😱 It is new function.

However I always hiding dot files at Finder.app. It is too noisy. And I access to dot files via terminal always.
I like old behavior, thus I resolving this problem using the alias `hyper="${EDITOR} ~/.hyper.js"`.

I should to close this PR?

citycide:
I'd leave that for the team to decide, your most recent change would be more acceptable.

ringohub:
Thank you so much for giving your reviews and lots of advices 😄.

matheuss:
#1510 🕵️

PR #1494, opened 2017-02-09
Fix replaceParent reducer on termgroup exit

chabou:
activeRootGroup is not cleared when a child become a root termgroup
It happens when you exit a term with only two terms opened. The remaining termgroup containing a session replaces its parent and become root.


matheuss:
@rauchg pls review 🙌

PR #1497, opened 2017-02-12
Hyper rebuild

ppot:
After some times talking on where Hyper should be going, we decided the non-core functionality should be lend toward plugins.

The ongoing discussion  toward moving to **xterm.js** #1275 that is actively maintained lend to a slow implementation toward that case. The fact that **hterm** is not actively maintained lead us to lost of monkey-patching to make it compliant to our application. 

Also, The current implementation force us to have the `Term` bonded with multiple `props` which is not ideal. The implementation of the split made the propagation of those harder and less structured. Therefore resulting in some sort of monkey-patching along the way. 

### Keymap
Issue: https://github.com/ppot/hyper/issues/11
On this implementation, I included a base of Keymap following three specific files.
```
darwin.json
win32.json
linux.json
```

This way, the **Keymap** will be easier to manage the using accelerator. Since we want the **Keymap** to be easily modified and not forcing the remplacement of `metaKey` depending of the operating system.  

### Pane
issue: https://github.com/ppot/hyper/issues/13
The introduction of **Pane** is leading toward the ability to make custom Pane that are not a stack of multiple objects such as Webpane and Term. Using Both on the same Component mean also that the **Keymap** assigned to specific action will be mixed up. The same keybinding can lunch different action. For exemple. `Ctrl+C` to close the pty process versus closing the opened window.

This implementation is ongoing and to followup and discussion should be made inside specific issues inside https://github.com/ppot/hyper/issues

### Why ?
Since there is many issues in the current implementation, some are linked into specific issues already. This why it will be easier for me to keep track of the different task.


Tyriar:
Any reason you're committing xterm.js rather than linking to it as a dependency?

ppot:
@Tyriar Yes. Using the node version had lots of warning with webpack most related to addons.
The current version `2.3.0` [http://xtermjs.org/blog/articles/2017-02-08-xtermjs-2.3](url) resolve the warning. So i included it now.

rauchg:
Lots of awesome work in here, but in order to review this we'll have to split it into smaller PRs I think 🙌

ppot:
@rauchg The PR is to keep track. But I don't think it can be made into smaller PR since in the end IT will rework the core 🐳

MartyGentillon:
Hey @ppot, I was wondering what your status is on this.

I ask because I was thinking of working on hypertext terminal support (things like displaying images, video, formatted text, and other things inline), and it would be more sensible to do that after this migration was complete (possibly even writing it into the xterm.js library).

ppot:
@MartyGentillon One Keymaps is integrated I can work on this again. Since they found the status to big for now.

MartyGentillon:
That makes sense, but doesn't really answer the core of my question, which I suppose is more along the lines of "what needs to be done to finish this?" and "how useable is this currently?"

ppot:
From what I remember when working on it. I have some issue with pane reordering when root pane is close. It's usable with xterm which is the purpose of the PR. I should do a copy of it and exclude temporary xterm since the goal of the PR is not including xterm but rebuilding the structure of Hyper.

MartyGentillon:
Thanks.

ppot:
I am deleting this.

PR #1510, opened 2017-02-15
Revert "Open Preferences in Default JS Editor on Windows"

matheuss:
Reverts #1138

#1138 removed the `CMD+,` which wasn't intended.

Other than that, #1383 fixes the `\n` issue on Notepad, so the reasoning from #1138 (`Notepad doesn't handle \n line breaks, so config file is unreadable.`) is _invalid_.

Also, I don't see how the `Default JS Editor on Windows` is being used on `shell.showItemInFolder` 🤔

### DIFF
--- a/lib/actions/ui.js
+++ b/lib/actions/ui.js
@@ -1,6 +1,4 @@
-import os from 'os';
-import {shell} from 'electron';
-import path from 'path';
+import {EOL} from 'os';
 
 import * as shellEscape from 'php-escape-shell';
 import last from '../utils/array';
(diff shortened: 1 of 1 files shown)


lib/actions/ui.js
```diff
@@ -212,12 +210,33 @@ export function moveTo(i) {
...
+  message.push(' echo If it fails, open it manually with your favorite editor!');
+
   return dispatch => {
     dispatch({
       type: UI_SHOW_PREFERENCES,
       effect() {
-        dispatch(
-          shell.showItemInFolder(path.join(os.homedir(), '.hyper.js')));
+        dispatch(requestSession());
+        // Replace this hack with an async action
```
albinekb: `TODO`?

jongalloway:
As the author of the PR that was reverted here, I absolutely agree. That was just a temporary workaround, now that #1383 makes things work fine in Notepad, this is better behavior. Thanks!

matheuss:
@jongalloway 🙌

PR #1516, opened 2017-02-16
fix undefined fish shell match

### LINKED ISSUE #1513: Issues with plugins on Hyper 1.2.0 & 1.2.1

- **OS version and name**: `10.12.3 macOS Sierra`
- **Hyper.app version**: `1.2.1`
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: `Default`
- **Relevant information from devtools** _(CMD+ALT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: `N/A`
- **The issue is reproducible in vanilla Hyper.app**:  `Is Vanilla`

## Issue
Getting `Error updating plugins.` even without any plugins. It says to check for `~/.hyper_plugins/npm-debug.log` but no such file exists.

Seems to be having issues with downloading plugins as well. Getting `Plugin "name" failed to load (Cannot find module '/Users/henrik/.hyper_plugins/node_modules/name')`

Tried installing fresh, simply reinstalling as well. Still remains.

henrikdahl:
Fixes #1513 

Basically if you have no shell specifically set in config then `const shell` is set to `undefined` which match fails on because it can't match on `undefined`.

This commit matches against `cfgShell` instead.

### DIFF
--- a/app/plugins.js
+++ b/app/plugins.js
@@ -241,7 +241,7 @@ function install(fn) {
       posix: 'npm prune && npm install --production'
     };
     // determine the shell we're running in
-    const whichShell = cfgShell.match(/fish/) ? 'fish' : 'posix';
+    const whichShell = (typeof cfgShell === 'string' && cfgShell.match(/fish/)) ? 'fish' : 'posix';
     // Use the install command that is appropriate for our shell
     exec(installCommands[whichShell], {
       cwd: path,
--- a/app/plugins.js
+++ b/app/plugins.js
@@ -241,7 +241,7 @@ function install(fn) {
       posix: 'npm prune && npm install --production'
     };
     // determine the shell we're running in
-    const whichShell = shell.match(/fish/) ? 'fish' : 'posix';
+    const whichShell = cfgShell.match(/fish/) ? 'fish' : 'posix';
     // Use the install command that is appropriate for our shell
     exec(installCommands[whichShell], {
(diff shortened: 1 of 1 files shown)


app/plugins.js
```diff
@@ -241,7 +241,7 @@ function install(fn) {
       posix: 'npm prune && npm install --production'
     };
     // determine the shell we're running in
-    const whichShell = shell.match(/fish/) ? 'fish' : 'posix';
+    const whichShell = cfgShell.match(/fish/) ? 'fish' : 'posix';
```
albinekb: Sure we don't want to check `shell` as a fallback to `cfgShell`?  🤔  Have not had time to check the full code, but tested this branch with fish and it seems to work as expected

albinekb: for extra ⭐️  i would rename this variable to `shellType` to be more descriptive

albinekb: Maybe change this to:
```js
const whichShell = (typeof cfgShell === 'string' && cfgShell.match(/fish/)) ? 'fish' : 'posix';
```
To guard for values that aren't strings

zsol:
I just ran into this. After fixing it for myself I realized this is already addressed in `master` but not in the latest release. Kind of a turnoff for a completely new user to see this right after installing the latest release for the first time. Maybe cut a hotfix release?

PR #1532, opened 2017-02-16
add select color based on cursor color

henrikdahl:
Sets the selection color based on cursorColor instead of the default blue one ✌️


rauchg:
Does it hot reload?

rauchg:
yep it should

PR #1536, opened 2017-02-17
Fix unicode display: CJK, powerline, tmux

tibotiber:
Hi! This is a follow up PR from #1376, taking 2e58e56 into consideration. I have isolated the issue with braille characters for `vtop` by decorating the corresponding nodes with a `braille-node` class additionally to the existing `unicode-node` class. It is ready to merge to the best of my knowledge.
Cc @rauchg 

**WIP summary:**
- [x] powerline wide unicode glyphs display
- [x] CJK chars display, `wc-node` css
- [x] multiple rendering of CJK in some cases, see  https://github.com/zeit/hyper/pull/1376#issuecomment-271876444 / https://github.com/zeit/hyper/pull/1376#issuecomment-272191734
- [x] tmux empty prompt width, see screenshot below:
tmux

### DIFF
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -349,6 +349,115 @@ hterm.Terminal.prototype.focusHyperCaret = function () {
   this.hyperCaret.focus();
 };
 
+hterm.Screen.prototype.insertString = function(str) {
+  var cursorNode = this.cursorNode_;
+  var cursorNodeText = cursorNode.textContent;
+  this.cursorRowNode_.removeAttribute('line-overflow');
+  // We may alter the width of the string by prepending some missing
+  // whitespaces, so we need to record the string width ahead of time.
+  var strWidth = lib.wc.strWidth(str);
+  // No matter what, before this function exits the cursor column will have
+  // moved this much.
+  this.cursorPosition.column += strWidth;
+  // Local cache of the cursor offset.
+  var offset = this.cursorOffset_;
+  // Reverse offset is the offset measured from the end of the string.
+  // Zero implies that the cursor is at the end of the cursor node.
+  var reverseOffset = hterm.TextAttributes.nodeWidth(cursorNode) - offset;
+  if (reverseOffset < 0) {
+    // A negative reverse offset means the cursor is positioned past the end
+    // of the characters on this line.  We'll need to insert the missing
+    // whitespace.
+    var ws = lib.f.getWhitespace(-reverseOffset);
+    // This whitespace should be completely unstyled.  Underline, background
+    // color, and strikethrough would be visible on whitespace, so we can't use
+    // one of those spans to hold the text.
+    if (!(this.textAttributes.underline ||
+          this.textAttributes.strikethrough ||
+          this.textAttributes.background ||
+          this.textAttributes.wcNode ||
+          this.textAttributes.tileData != null)) {
+      // Best case scenario, we can just pretend the spaces were part of the
+      // original string.
+      str = ws + str;
+    } else if (cursorNode.nodeType == 3 ||
+               !(cursorNode.wcNode ||
+                 cursorNode.tileNode ||
+                 cursorNode.style.textDecoration ||
+                 cursorNode.style.backgroundColor)) {
+      // Second best case, the current node is able to hold the whitespace.
+      cursorNode.textContent = (cursorNodeText += ws);
+    } else {
+      // Worst case, we have to create a new node to hold the whitespace.
+      var wsNode = cursorNode.ownerDocument.createTextNode(ws);
+      this.cursorRowNode_.insertBefore(wsNode, cursorNode.nextSibling);
+      this.cursorNode_ = cursorNode = wsNode;
+      this.cursorOffset_ = offset = -reverseOffset;
+      cursorNodeText = ws;
+    }
+    // We now know for sure that we're at the last character of the cursor node.
+    reverseOffset = 0;
+  }
+  if (this.textAttributes.matchesContainer(cursorNode)) {
+    // The new text can be placed directly in the cursor node.
+    if (reverseOffset == 0) {
+      cursorNode.textContent = cursorNodeText + str;
+    } else if (offset == 0) {
+      cursorNode.textContent = str + cursorNodeText;
+    } else {
+      cursorNode.textContent =
+          hterm.TextAttributes.nodeSubstr(cursorNode, 0, offset) +
+          str + hterm.TextAttributes.nodeSubstr(cursorNode, offset);
+    }
+    this.cursorOffset_ += strWidth;
+    return;
+  }
+  // The cursor node is the wrong style for the new text.  If we're at the
+  // beginning or end of the cursor node, then the adjacent node is also a
+  // potential candidate.
+  if (offset == 0) {
+    // At the beginning of the cursor node, the check the previous sibling.
+    var previousSibling = cursorNode.previousSibling;
+    if (previousSibling &&
+        this.textAttributes.matchesContainer(previousSibling)) {
+      previousSibling.textContent += str;
+      this.cursorNode_ = previousSibling;
+      this.cursorOffset_ = lib.wc.strWidth(previousSibling.textContent);
+      return;
+    }
+    var newNode = this.textAttributes.createContainer(str);
+    this.cursorRowNode_.insertBefore(newNode, cursorNode);
+    this.cursorNode_ = newNode;
+    this.cursorOffset_ = strWidth;
+    return;
+  }
+  if (reverseOffset == 0) {
+    // At the end of the cursor node, the check the next sibling.
+    var nextSibling = cursorNode.nextSibling;
+    if (nextSibling &&
+        this.textAttributes.matchesContainer(nextSibling)) {
+      nextSibling.textContent = str + nextSibling.textContent;
+      this.cursorNode_ = nextSibling;
+      this.cursorOffset_ = lib.wc.strWidth(str);
+      return;
+    }
+    var newNode = this.textAttributes.createContainer(str);
+    this.cursorRowNode_.insertBefore(newNode, nextSibling);
+    this.cursorNode_ = newNode;
+    // We specifically need to include any missing whitespace here, since it's
+    // going in a new node.
+    this.cursorOffset_ = hterm.TextAttributes.nodeWidth(newNode);
+    return;
+  }
+  // Worst case, we're somewhere in the middle of the cursor node.  We'll
+  // have to split it into two nodes and insert our new container in between.
+  this.splitNode_(cursorNode, offset);
+  var newNode = this.textAttributes.createContainer(str);
+  this.cursorRowNode_.insertBefore(newNode, cursorNode.nextSibling);
+  this.cursorNode_ = newNode;
+  this.cursorOffset_ = strWidth;
+};
+
 hterm.Screen.prototype.syncSelectionCaret = function () {
   const p = this.terminal.hyperCaret;
   const doc = this.terminal.document_;
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -86,21 +86,17 @@ lib.wc.substr = function (str, start, optWidth) {
   const chars = runes(str);
   let startIndex;
   let endIndex;
-  let width = 0;
+  let width;
 
-  for (let i = 0; i < chars.length; i++) {
-    const codePoint = chars[i].codePointAt(0);
-    const charWidth = lib.wc.charWidth(codePoint);
-    if ((width + charWidth) > start) {
-      startIndex = i;
+  for (startIndex = 0, width = 0; startIndex < chars.length; startIndex++) {
+    width += lib.wc.charWidth(chars[startIndex].codePointAt(0));
+    if (width > start) {
       break;
     }
-    width += charWidth;
   }
 
   if (optWidth) {
-    width = 0;
-    for (endIndex = startIndex; endIndex < chars.length && width < optWidth; endIndex++) {
+    for (endIndex = startIndex, width = 0; endIndex < chars.length && width < optWidth; endIndex++) {
       width += lib.wc.charWidth(chars[endIndex].charCodeAt(0));
     }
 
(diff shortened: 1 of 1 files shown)


albinekb:
Wow, this is awesome! Will try it tonight 👍

tibotiber:
Screenshots of the result here: vtop and unicode + tmux
vtop
unicode

rauchg:
Only issue is: shouldn't we respect the grid size for emojis? Both Terminal.app and iTerm seem to overlap them:

image

image

I wonder if by taking space _wider than a character_ we would be breaking some UIs ?

rauchg:
Also, since other Terminals limit the width of Emojis, I worry that we'd be displaying them _not as nicely_ in Hyper, because they'd essentially have an extra space

tibotiber:
@rauchg: great point about emojis, one important breaking UI is the cursor position. However we need to address at least powerline unicode characters (e.g. https://github.com/zeit/hyper/issues/1336#issuecomment-270670882), which is handled properly by iTerm as well. Emojis being spread over codepoints, it would be easier to isolate the [powerline glyphs](https://apw-bash-settings.readthedocs.io/en/latest/fontpatching.html#glyph-table) and give them more space. We can choose to apply flexible or double spacing, I would suggest double to keep the cursor position in check. From what I can see, this is the way iTerm handles it too.

After that, I think I should remove css decoration for braille, since it will fall in the normal unicode styling (?).

_Sidenote:_ This is getting interesting, powerline glyphs are stored in the [Unicode Private Use Area](https://en.wikipedia.org/wiki/Private_Use_Areas#Usage), along with medieval scripts, Klingon, and various languages of the Middle-earth... Should we display that with fix width or double width like powerline glyphs? 😂

rauchg:
@tibotiber that's a very good question. I think ultimately it would depend on each glyph in the private use area whether they want double or single width.

Let's do this: for now, let's give *all* private use area double (which would encompass Powerline glyphs), and then if a request comes up, we can think about adding a config flag for specifying _ranges_ that you want to be double width

tibotiber:
OK, I'll push changes today (it's Saturday here already). I'll remove the braille part then.

tibotiber:
_Update:_ The issue with powerline glyphs is not a width issue. I couldn't get this to work in a way that supports basic powerline glyphs as well as user defined ones, so I checked iTerm a bit more and realised they do not treat powerline glyphs (new or old) as wide chars. What happens is wide unicode simply overlaps on the next char. See screenshots for iTerm and Hyper:
hyper
iterm
If you look at the `cat` above, both terminals treat the glyphs exactly in the same way. Now, the difference is in the powerline prompt line where there is a background color. See the branch sign being clipped in Hyper? Because Hyper encapsulates each unicode char in its own `<span />`, the wide unicode do not visibly overlap on the char to its right, since that char's background color is on top of whatever overlaps. This is not an issue in iTerm where unicode is probably not encapsulated like this. Why do we encapsulate unicode in individual containers exactly?

I removed my previous changes and only left the low hanging fruit: fixing of CJK wide chars. Still trying to find a way around the powerline issue. Summary of issues i'm trying to address and the ones fixed:

- [x] CJK chars display
- [ ] powerline wide unicode glyphs display
- [ ] tmux empty prompt width, see screenshot below:
tmux

rauchg:
> Why do we encapsulate unicode in individual containers exactly?

As explained above, without the `<span>` with a fixed width, glyphs that are not included in a monospace font (such as braille which uses the "Apple Braille" font would show up in a different width (and this width varies per-character) and break rendering)

> What happens is wide unicode simply overflows on the next char

That's not true for Emojis it seems? Or emojis are not "wide unicode" (why not?)

image

Another thing I'd like to add to your list: ensure chinese characters render correctly

rauchg:
More food for thought: how should `﷽` be rendered?

rauchg:
Personally, I like what Terminal.app does (similar to emoji handling). Take up one column, show overflow.

image

iTerm takes up one column, but only shows partial overflow:

image

tibotiber:
OK, that's what I had gathered about the encapsulation, just wanted to make sure there were no other reasons.

> That's not true for Emojis it seems? Or emojis are not "wide unicode" (why not?)

I don't understand. Emojis are wide unicode, and they do overflow on the next char in iTerm, don't they? This is what your screenshot shows with the heart overflowing from its span onto the next (behind the `a`). Maybe it's a vocabulary issue, I have the feeling we're saying the same thing.

Good job finding ﷽ 😂, I'll target full overflow, like in Terminal.app.

> Another thing I'd like to add to your list: ensure chinese characters render correctly

Which part exactly? The `wc-node` behaviour is done. Are you referring to the 2 chars that displayed as 3 mentioned in https://github.com/zeit/hyper/pull/1376#issuecomment-271876444 / https://github.com/zeit/hyper/pull/1376#issuecomment-272191734? Or anything else?

rauchg:
Just wanted to be 100% sure! Seems like we're on the same page @tibotiber.
Unicode nirvana (for _output) is within reach! 

Next up once we merge this PR is to fix backspace (the behavior of the ANSI escape that is) for surrogate pairs and caret position

rauchg:
Yep I was referring to the report by @BirkhoffLee

rauchg:
One super tricky edge case that'll be difficult to solve (with my current knowledge of CSS):

We want it to overflow, but if it's at the edge, we would want the parent container to trigger scrolling somehow.

No terminal that i know of seems to handle this edge case currently

image

rauchg:
Also not a priority for now considering our broken backspace, but thought I'd mention it

tibotiber:
Took a while to get back to this, sorry. I pushed new changes to try and fix the display of unicode in powerline prompts:
- removed existing unicode patching that was creating new containers for unicode chars. It was making it impossible for unicode to overflow as desired when there is a background color (e.g. powerline).
- instead I override hterm's `insertString` to get back the node in which the string has been added and I decompose its `textContent` when it contains unicode to insert child nodes wrapping unicode in individual spans. Doing so we still manage to get monospaced unicode but it can now overflow as long as the background color doesn't change.

This is partially working and I can't figure out why only "partially". It basically works for general unicode display such as `cat` files or `vtop`, it also work with file editors when re-rendering rows etc, but the node's content is flattened as if nothing happened in powerline prompts, except for the last node in a row.

I think it has something to do with the way hterm handles data decoding in `hterm.VT.prototype.parseUnknown_` but I can't manage to pinpoint exactly what. I could be wrong though and was hoping someone has an idea. I was originally not familiar with hterm's internals so I might miss something.

I left some debug logs that help to understand what the new `insertString` does.

tibotiber:
Wow, I almost forgot about this PR. Got back to it yesterday with fresh eyes and finally found what was missing. hterm's `insertString` was actually overriding my spans content to inject white spaces under some circumstances. Really nothing to do with the parser 😅.

So it is now where I wanted it to be, i.e. it encapsulate unicode nodes in a span within the current container, instead of creating a new container for each unicode node. This fixes the display of legacy powerline icons:
old-powerline

wide chars (CJK) display is fixed, and unicode nodes are displayed with fixed width and full overflow as suggested by @rauchg:
unicode

vtop and others are displaying well:
vtop

and 2-sided powerline prompts in tmux are now fixed:
tmux


The only thing left is the issue where multiple CJK chars are rendered. I'll look into it, but we can also merge this PR first and when I fix this and I'll open another one. I'll let you guys review before I squash the commits (lots of trial and error, boo).

Cc @albinekb @rauchg

tibotiber:
Good news, I found the issue that caused multiple wide chars (e.g. CJK) to be rendered. It was a case where hyper's override of `lib.wc.substr` had `startIndex = undefined`. This PR is complete now 🎉. I fixed it by matching closer to the original code, despite using `runes` to be emoji-friendly.

cjk

albinekb:
Wow @tibotiber 😱

lib/hterm.js
```diff
@@ -388,6 +345,157 @@ hterm.Terminal.prototype.focusHyperCaret = function () {
...
+    }
+    const newNode = this.textAttributes.createContainer(str);
+    this.cursorRowNode_.insertBefore(newNode, cursorNode);
+    this.cursorNode_ = newNode;
+    this.cursorOffset_ = strWidth;
+    this.wrapUnicode(newNode);
+    return;
+  }
+  if (reverseOffset === 0) {
+    const nextSibling = cursorNode.nextSibling;
```
albinekb: nit:
```js
const { nextSibling } = cursorNode;

tibotiber: I'd rather keep it like the original `hterm` code to make diffs easier. Makes sense?

tibotiber:
@albinekb thank you for the review. Most of the voodoo magic is not mine actually, it's mainly the original `hterm` code that I patched as minimally as possible:
- replaced `var` by `const` or `let` to match local code style
- removed comments (it was extremely verbose and still available in the original code)
- made sure to call `wrapUnicode` in `insertString` and `splitNode_`

The `wrapUnicode` method is mine, so that might be a good place to focus on in reviews 😛.

lib/hterm.js
albinekb: Absolutely! 👍

albinekb:
After running `npm run dist` (macOS 10.12.4 (16E195)) I get 
image this error 🤔  Is this just me? @rauchg  @tibotiber

I'm gonna try master now as we speak..

rauchg:
I just got the same thing

rauchg:
Reverting

albinekb:
I'll have a look for a fix

tibotiber:
Hey, sorry about this, didn't try to `run dist`. I'll have a look as well.

tibotiber:
I'm not able to reproduce. I'm only getting a warning for an unknown OSC code.

screen shot 2017-05-20 at 12 37 56 pm

Might be coming from different prompts, what is yours like?

albinekb:
I use https://github.com/brandonweiss/pure.fish and i think rauch is using https://github.com/sindresorhus/pure @tibotiber

tibotiber:
Mmmmh, I'm on sindresorhus pure as well actually. I tried a few other prompts (antigen themes) and I'm getting no error. Not understanding this.

albinekb:
🤔 I'll diag & give you something to reproduce with tomorrow, need to get some 💤

mengzhuo:
Why don't we just let the webkit do this work instead of calculate the width of the char?

albinekb:
Sorry for not coming back yet, I'll get this sorted tomorrow 👼

mengzhuo:
@albinekb 

I use "Cousine for Powerline"



After I remove all unicode-node style.

I think it's a "web-based" app, we should due with it more "web-way" by using more css instead of messing with DOM.

tibotiber:
@mengzhuo, the issue with this approach is that you break the rendering for unicode that is not monospaced (see [this comment](https://github.com/zeit/hyper/pull/1536#issuecomment-280972848)). You can observe a bad example by running `vtop` after removing unicode styling. More generally, terminals expect monospace chars and handle exceptions such as CJK with "wide-chars" which are doubled-spaced.

albinekb:
Sorry for the delay @tibotiber 

the problem is that not all nodes has `className`, for example text-nodes (see https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType)

I tried two things:

- adding a check in `hterm.Screen.prototype.wrapUnicode` to only handle element nodes (type 3), no errors but emojis get scrambled.
- adding a check for `node.className`, same problem

tibotiber:
@albinekb thanks for posting more details. I looked into it and here are some thoughts:
1. we seem to have an issue with cursor position. Basically the cursor moves with double spacing on single spaced unicode. Since `hterm.Terminal.prototype.eraseToRight` uses the cursor position to calculate how many chars should be deleted, it creates an issue with unicode deletion. I'm thinking this might be the source of the unknown char that appears when we delete emojis. Note: this issue exists in master so it is not new to this PR and could be tackled in another PR.
2. nice find on "the problem is that not all nodes has className", I completely missed it. I fixed this locally and will open a new PR soon since this one is now frozen.

tibotiber:
Quite honestly, `hterm` is horrible to patch because the code is extremely imperative and convoluted. Do you know of any other terminal emulator that would be more "react-ish"? That would be really cool! @rauchg?

LeonBlade:
I stumbled upon this PR and wanted to ask here before creating an issue first, so sorry in advance for posting this here.

By any chance is this rendering issue related to the problem produced in this screenshot: Terminal and Hyper

On the left is Terminal and on the right is Hyper.  I'm speaking of course about the box side which seems to warp as well as contain some overlap between the characters.

tibotiber:
@LeonBlade this will need to be checked once the follow up PR #1873 lands, but it should fix your issue I think.

uetchy:
I've tested it with Japanese and Korean characters on macOS El Capitan and there are some visual errors appeared.
There is an iTerm2 screen next to Hyper to clarify differences.
It looks like the width of Japanese characters is too narrow and Korean characters are weirdly broken.

screenshot 2017-06-02 22 01 32

ELLIOTTCABLE:
Hi! I don't necessarily have a horse in this race, besides the ones that the contributors above have (Pretty Unicode Interfaces [for english text] And Maybe Sometimes Emoji) — but I want to point out that this *heavily* affects users from other countries and cultures? This seems like a question that's ideal for a ‘reach out and ask users’ kind of moment.

If nothing else, I suspect we need to get a ~Unicode expert~ to come in and answer some questions about character ranges to treat as ‘wide’ (our implementation gives them double width, automatically consuming an extra column) versus ones to ignore, treat as single-wide, and leave for userspace handling (“What? This emoji looks funny? I guess I'd better spit out an extra space after it.” *aka* The Yarn Approach).

ELLIOTTCABLE:
An aside — copied [from a friend](https://twitter.com/FakeUnicode/status/886854168736092162), as an example of why the emoji in particular are A Hard Problem, and “display all private-use-area characters as two columns” isn't enough:

> 👨‍❤️‍💋‍👨 1f468-200d-2764-fe0f-200d-1f48b-200d-1f468 KISS: MAN, MAN
> 8 codepoints, 11 UCS-2 units, one Wide character.
> 
> — @‌FakeUnicode‬

sachin21:
Great!! You saved my life. Thank you. I'll looking forward to be merged :+1:

sachin21:
@tibotiber Great!! Thank you. I'll looking forward to be merged :+1: 

you saved my life :dango:

lucasqueiroz:
Seems like 1.4.x fixed this issue.
Vtop with 1.4.2 (PragmataPro Font):
hyper_001

PR #1541, opened 2017-02-17
Add perf optimizations

rauchg:
- No longer rely on the `Write` record being assigned to state by Redux
  - It was kind of odd and ultimately it's best to skip Redux to make imperative API calls
  - Fewer object allocations
- Term components are now cached when mounted / unmounted so that we can reach them from the new middleware
- Add a store middleware to perform the term writes when we get a `SESSION_PTY_DATA` action
- Cache measurement of codepoints (`strWidth`), which avoids calling `runes` too much which V8 has de-opted.
- Improve reducers logic to avoid merges that create a new object that won't get merged anyways for this critical path


rauchg:
@ppot re-based.

matheuss:
:nice:

PR #1544, opened 2017-02-17
windows: fix the missing dependency

stefanivic:
As seen here, the builds brake on Windows due to a missing dependency. https://ci.appveyor.com/project/appveyor-zeit/hyper/build/1.0.564

Adding it will fix the issue.


rauchg:
Thanks @stefanivic

PR #1547, opened 2017-02-17
add hterm cursor blink support

henrikdahl:
Add support for hterm blinking cursor with config options default set to `false`. 

kapture 2017-02-18 at 0 45 12


unluisco:
Hi Henrik, I think that the blink is a bit long. I compared it with atom.

henrikdahl:
@unluisco We're relying on hterm for the blink support so it's basically what they have set as a default: https://github.com/macton/hterm/blob/master/src/hterm_terminal.js#L2046-L2047

It's entirely possible to expose this interval as a config option but ultimately not my decision, cc @rauchg

PR #1549, opened 2017-02-18
add configurable uifontfamily

### LINKED ISSUE #1401: Tabs font stack unavailable to plugins

- **OS version and name**: 10.12.2 macOS Sierra
- **Hyper.app version**: 1.1.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: default
- **Relevent information from devtools** _(CMD+SHIFT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: 

## Issue

Tabs font stack isn't available through any variable:

https://github.com/zeit/hyper/blob/12ae25a60b800efdd19205a0f259bd53c1668273/lib/components/tabs.js#L70-L73

Would be neat to be able to access it easily so that you can rely on it in plugins. Currently if you just copy the stack manually it wouldn't update if Hyper ever decides to change it.

henrikdahl:
Fixes #1401 

Removing static font stacks used for UI and replacing them with a config var so that plugins can easily access it.

### DIFF
--- a/app/config-default.js
+++ b/app/config-default.js
@@ -6,9 +6,6 @@ module.exports = {
     // font family with optional fallbacks
     fontFamily: 'Menlo, "DejaVu Sans Mono", Consolas, "Lucida Console", monospace',
 
-    // UI font family with optional fallbacks
-    uiFontFamily: '"-apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif"',
-
     // terminal cursor background color and opacity (hex, rgb, hsl, hsv, hwb or cmyk)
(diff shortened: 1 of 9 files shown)


app/config-default.js
```diff
@@ -6,6 +6,9 @@ module.exports = {
     // font family with optional fallbacks
     fontFamily: 'Menlo, "DejaVu Sans Mono", Consolas, "Lucida Console", monospace',
 
+    // UI font family with optional fallbacks
+    uiFontFamily: '"-apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif"',
+
```
rauchg: I think it's a bit too much for the default. Most people would expect system font to be the UI font and call it a day

PR #1551, opened 2017-02-18
only show tab title on overflow

henrikdahl:
Now only shows the tabs titles if the text is actually overflowing.


rauchg:
This will probably have concerning performance implications

henrikdahl:
Possibly. I don't think there's a way to make this efficient enough ✌️

henrikdahl:
Looks like terminal always shows the title as well so I guess we're fine as is.

PR #1560, opened 2017-02-20
add always on top in menu

henrikdahl:
Adds a `Float on Top` menu toggle item that makes Hyper stay above every other window when enabled.

skarmavbild 2017-02-20 kl 13 35 29

It's a very handy yet simple feature to have. It's the same that can be found in [Quicktime](http://www.iclarified.com/53150/how-to-keep-the-quicktime-video-player-always-on-top)/VLC and other apps.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -145,6 +145,7 @@ app.on('ready', () => installDevExtensions(isDev).then(() => {
       height,
       minHeight: 190,
       minWidth: 370,
+      alwaysOnTop: focusedWindow ? focusedWindow.isAlwaysOnTop() : false,
       titleBarStyle: 'hidden-inset', // macOS only
       title: 'Hyper.app',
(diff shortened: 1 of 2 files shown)


rauchg:
Seems like you could write a plugin for this:

https://github.com/henrikdahl/hyper/blob/2f8bd25737260c58619501c587734e9e42d89403/app/index.js#L163

henrikdahl:
I think its far too simple to be a plugin.

chabou:
A plugin is never too simple 😄 
`If something can be done as a plugin, it should be a plugin instead of a core Hyper feature`

henrikdahl:
Fair enough, if someone is interested in turning this into a plugin feel free to use this code as reference ✌️

chabou:
Done: https://github.com/chabou/hyper-always-on-top 🎉

Issue #1480, opened 2017-02-03
As a Windows Hyper user, I should be able to install packages regardless of which shell I am using


belthesar:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x ] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 10 build 15014
- **Hyper.app version**: 1.1.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/belthesar/b8a8f608a5ed47becd96068706eb6648
- **Relevant information from devtools** _(CMD+ALT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: N/A
- **The issue is reproducible in vanilla Hyper.app**: Yes
## Issue

Problem:  When running a shell other than 'cmd' on Windows, packages do not install successfully.  
Steps to Reproduce:  

1. With a functional Hyper install (vanilla, or with packages installed), modify your .hyper.js config to change the shell to a different shell (EG: "C:\\Windows\\System32\\bash.exe" or "%SystemRoot%\system32\WindowsPowerShell\v1.0\powershell.exe"). 
2. Add or remove packages.

Expected Result:  

Packages are installed or uninstalled as requested by adjusting the config. 

Actual Result:  

Packages are not installed.  Error suggest to look at %userprofile%\.hyper_plugins\npm-debug.log
This log file is never generated.  

Additional info:  Running Hyper within a cmd window provides additional logging that suggests that there is some sort of hard coding that is attempting to run whatever the current shell is with switches that only work with 'cmd'.  

'''
plugins have changed / not init, scheduling plugins installation
running in prod mode
electron will open file://C:\Users\belthesar\AppData\Local\hyper\app-1.1.0\resources\app\index.html
Checking for notification messages
No matching notification messages
Error: Command failed: npm prune && npm install --production
/bin/bash: /s: No such file or directory

    at ChildProcess.exithandler (child_process.js:218:12)
    at emitTwo (events.js:106:13)
    at ChildProcess.emit (events.js:191:7)
    at maybeClose (internal/child_process.js:877:16)
    at Socket.<anonymous> (internal/child_process.js:334:11)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at Pipe._handle.close [as _onclose] (net.js:493:12)
[...]

PR #1565, opened 2017-02-21
Falls back to default shell to install plugin on win32

kwonoj:
related to #1480.

This PR try to fix plugin installation fails on windows machine, if user specifies custom shell other than `cmd.exe`. The crux of this issue is due to behavior of `child_process::exec` (https://nodejs.org/dist/latest-v6.x/docs/api/child_process.html#child_process_child_process_exec_command_options_callback), node.js requires any shell specified in windows should compliant to command line parsing with `cmd.exe` to understand `/s /c` (or `/d /s /c` on later version) and does not provide any way to override it.

There isn't seems stable manner of drop-in replacement to child process supports this , so in this PR simply let `exec` falls back to default shell instead of custom shell when install plugins.

### DIFF
--- a/app/plugins.js
+++ b/app/plugins.js
@@ -242,12 +242,20 @@ function install(fn) {
     };
     // determine the shell we're running in
     const whichShell = (typeof cfgShell === 'string' && cfgShell.match(/fish/)) ? 'fish' : 'posix';
-    // Use the install command that is appropriate for our shell
-    exec(installCommands[whichShell], {
+    const execOptions = {
       cwd: path,
-      env,
-      shell
-    }, err => {
+      env
+    };
+
+    // https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback
+    // node.js requires command line parsing should be compatible with cmd.exe on Windows, should able to accept `/d /s /c`
+    // but most custom shell doesn't. Instead, falls back to default shell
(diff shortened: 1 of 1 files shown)


rauchg:
Are you sure you meant to test for _not windows_? I don't understand the logic 100%

kwonoj:
Changed logic is 

1. if platform is windows
2. fall back to shell to `cmd.exe`, by ignoring any custom configuration in hyper.js

if platform is non-windows, it'll still honor config as-is. Main issue in here is `node.js` force inject `/s /c` command for windows, assumes all shell specified compliant to `cmd.exe`, in opposite for non-windows it always inject `-c`, assumes shell complaint to those switch. 

In windows, it is possible to have non-cmd compliant shell. For non windows, afaik it is not possible to have those shell since most known shell even including `fish` which isn't POSIX-compliant honors `-c` switch, doesn't need to fall back to default system shell.

kwonoj:
This is reason in code it test against if `platform isn't win32, specify shell as config file`.

dcinzona:
For the sake of conversation, I was having this same issue on my install on Windows 10 with Bash -> zsh set up as my default shell.  Including the changes that were part of this PR solved the issue for me.

rauchg:
Thanks @dcinzona

fbm-static:
This worked for me. Thank you!

PR #1595, opened 2017-03-02
Support XDG and other user config dirs

skeate:
Tries to find a config file in the following order:

- `$XDG_CONFIG_HOME`, if it's set
- Electron's `app.getPath('userData')`. This is:
  - `%APPDATA%` on Windows
  - `$XDG_CONFIG_HOME` on Linux
  - `~/Library/Application Support` on macOS
- User's home directory

For all but the home directory, files are in a subfolder of the directory which is named after the `packageName` in app/package.json.

For creation of a new config, again, `$XDG_CONFIG_HOME` is preferred, if set. Otherwise, it will use the `app.getPath('userData')` path.

This is largely a rewrite of #170, since that was too old to make rebasing/patching practical. Accordingly, merging this should close #170.

### DIFF
--- a/app/config-paths.js
+++ b/app/config-paths.js
@@ -1,0 +1,48 @@
+const {app} = require('electron');
+const {join} = require('path');
+const {homedir} = require('os');
+const {existsSync} = require('fs');
+const {productName} = require('./package.json');
+
+const xdgDir = process.env.XDG_CONFIG_HOME ?
+  join(process.env.XDG_CONFIG_HOME, productName) : false;
+const userDataDir = app.getPath('userData');
+const legacyDir = homedir();
+
+const xdgConfig = xdgDir ? join(xdgDir, 'init.js') : '';
+const userDataConfig = join(userDataDir, 'init.js');
+const legacyConfig = join(legacyDir, '.hyper.js');
+
+const xdg = existsSync(xdgConfig);
+const userData = existsSync(userDataConfig);
+const legacy = existsSync(legacyConfig);
+
+let paths;
+
+if (xdg || (xdgDir && !userData && !legacy)) {
+  paths = {
+    root: xdgDir,
+    config: xdgConfig,
+    plugins: join(xdgDir, 'plugins'),
+    localPlugins: join(xdgDir, 'plugins', 'local'),
+    joinConfigPath: (...paths) => join(xdgDir, ...paths)
+  };
+} else if (userData || !legacy) {
+  paths = {
+    root: userDataDir,
+    config: userDataConfig,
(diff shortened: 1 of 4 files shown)


skeate:
Caveat: I largely just followed the example in #170, but I'm not super familiar with the codebase. I'm not sure if there are any other places in the code that would refer to the config and/or plugin paths that would also need to be updated.

LinusU:
@skeate Could I recommend using the [application-config-path](https://github.com/LinusU/node-application-config-path) library? It would save you from having to implement all this yourself inside this project.

The resolution is also a bit different, but I think it's an improvement.

1. `XDG_CONFIG_HOME` is not considered on the macOS platform. I think this is better since users running macOS is more likely to want it in the `Application Support` but some legacy applications only made for Linux might need `XDG_CONFIG_HOME` to be present.
1. On Linux, if `XDG_CONFIG_HOME` is not present `$HOME/.config/<name>` is used, which is the proper XDG defined fallback
1. Support for older versions of windows (XP, 2000)
1. Uses `LOCALAPPDATA` instead of `APPDATA` on Windows > Vista. I believe this is more correct since the hyper settings is local to the specific user.

Platform | Location
--- | ---
OS X | `~/Library/Application Support/<name>`
Linux (XDG) | `$XDG_CONFIG_HOME/<name>`
Linux (Legacy) | `~/.config/<name>`
Windows (> Vista) | `%LOCALAPPDATA%/<name>`
Windows (XP, 2000) | `%USERPROFILE%/Local Settings/Application Data/<name>`

MindTooth:
Not sure I would want to have my configs under Application Support.  That would be more suited if we were not forced to edit a .js file and could use a GUI.

LinusU:
@derimagia Some apps, Inkscape comes to mind, have macOS releases which really just is the Linux app packaged to work. They don't really have any platform integration and uses X11 do draw windows. These applications read `XDG_CONFIG_HOME` which is why some people have it set on macOS.

But macOS isn't an XDG platform, and the correct way for a macOS app to behave is to follow Apple's recommendations, and put the stuff in `~/Library/Application Support/<name>`.

skeate:
FWIW, I set `$XDG_CONFIG_HOME` on macOS because I use the same configs on both macOS and Linux. All the programs whose configs I care about sharing support it.

That said, I no longer use Hyper and don't really want to resolve those conflicts, so I'm going to close this.

chabou:
I totally agree to have some overriding mechanism but I really don't want to change default location without a real need.

I suggest using `$XDG_CONFIG_HOME` if set, or home directory if not.

chabou:
Related subject:
https://github.com/zeit/hyper/issues/2932

PR #1601, opened 2017-03-05
Fix decorators check and documentation

chabou:
Every component can be decorated. It is a generic feature without component name checking.
We only check if a plugin exposes at least one valid method name.
This list of valid method names was not up to date as well as the website.

I updated website's Readme to reflect changes since repositories merge.

### DIFF
--- a/app/plugins.js
+++ b/app/plugins.js
@@ -22,10 +22,10 @@ const availableExtensions = new Set([
   'reduceUI', 'reduceSessions', 'reduceTermGroups',
   'decorateMenu', 'decorateTerm', 'decorateHyper',
   'decorateHyperTerm', // for backwards compatibility with hyperterm
-  'decorateTab',
+  'decorateHeader', 'decorateTerms', 'decorateTab',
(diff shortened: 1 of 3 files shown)


app/plugins.js
```diff
@@ -22,10 +22,10 @@ const availableExtensions = new Set([
   'reduceUI', 'reduceSessions', 'reduceTermGroups',
   'decorateMenu', 'decorateTerm', 'decorateHyper',
   'decorateHyperTerm', // for backwards compatibility with hyperterm
-  'decorateTab',
+  'decorateHeader', 'decorateTerms', 'decorateTab',
```
ppot: `decorateTabs` missing?

chabou: Was already there : https://github.com/chabou/hyper/blob/c94dc41809a8e3d3e82462a64ed3bf39136c4b3c/app/plugins.js#L27

chabou: I was about reorder declarations, but I preferred have fewest changes

Issue #1579, opened 2017-02-23
Calling "Command + Shift + R" on macOS causes Hyper to freak out


ryan4888:
Pretty self explanatory. I found this because I use "Command + Shift + R" a lot for development, and I found that when hitting it, by accident, in Hyper made it freak out. The following is a screen cap of it's behavior during this.

![image](https://cloud.githubusercontent.com/assets/10523324/23278886/fda9fad6-f9e0-11e6-9e04-b57811526266.png)

chabou:
Same for me : only when you have opened devtools before.
Closing/opening/closing devtools (dock mode) fixes this glitch.
Maybe an electron/devtools issue ?

gsklee:
@chabou I can reproduce this without opening the devtools.

landsman:
Same issue

matheuss:
#1602

Issue #1575, opened 2017-02-22
UI gets repeated all over the place?


polds:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: macOS Sierra 10.12.4 Beta
- **Hyper.app version**: 1.2.1
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/polds/b195ca548c0901425b8bc0898e86048c
- **Relevant information from devtools** _(CMD+ALT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: N/A?
- **The issue is reproducible in vanilla Hyper.app**: Yes

## Issue
I know my `.hyper.js` says otherwise but I can reproduce this entirely on vanilla Hyper.app, sometimes when starting a new tab or interacting with it the UI gets repeated all over the place. The gif below shows me adding a plugin which encourage the issue to happen faster

![fun](https://cloud.githubusercontent.com/assets/2372558/23232649/4fc3386a-f909-11e6-9a17-ba9ebb35ae52.gif)

gsklee:
Here's a wild guess... plugins have authored states / middleware compositions in a wrong way such that the existence of peer plugins is not considered.

Does this error happen when you only have one plugin installed at all? Does it happen when you have more than two plugins?

polds:
Yeah I can see it with no plugins installed, one plugin, two, etc.

polds:
Interestingly, I've just discovered as long as I have the `hyperborder` plugin installed this doesn't happen. Without that plugin it almost always happens.

Blindmikey:
I see something similar, when I ssh into a linux machine from windows' hyper install. And nano any file, or run `top` Lots of distortion, lots of capitol `B`s and `~`s too, hard to edit files through hyper this way. File scrolling ends up displaying only half a block of text, sometimes text position is incorrect until I resize hyper or move cursor over text. 

Will see if hyperborder solves any of my issues.

Rawnly:
I've seen the same effect when full reloading with new theme, i'll check `hyperborder` too

matheuss:
#1602

PR #1602, opened 2017-03-05
Fix flicks after a full reload or when close tab button (dis)appear. (Fix #1490, #1575, #1579)

### LINKED ISSUE #1490: Colored text flashes on hover/blur of term window
hyper text color artefacts

- **OS version and name**: Win10 64 
- **Hyper.app version**: 1.1.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: I haven't changed it

## Issue
Gif attached

chabou:
This PR fixes #1490.

Adding `-webkit-backface-visibility: hidden` does the trick. I didn't notice any perf issue after adding it.
Adding it to main body or iframe body does the trick too. But I think it is better to colocate the trick with the incriminated transition.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -9,6 +9,7 @@
     <style>
       body {
         color: #fff;
+        -webkit-backface-visibility: hidden;
       }
 
       * {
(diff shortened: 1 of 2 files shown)


chabou:
Ok! put ` -webkit-backface-visibility: hidden` on main body fixes flick after a full reload too (#1575 and #1579) 🎉

matheuss:
amazing

PR #1191, opened 2016-12-14
Keyboard shortcut prefix fix for Windows, Linux (Ctrl+Shift). Fixes #1069.

nbolten:
This pull request fixes #1069 (and related issues like #1100).

### Overview

Interactive command line applications commonly bind to <kbd>Ctrl</kbd>+<kbd>&lt;key&gt;</kbd> keyboard shortcuts. For example, `nano` uses <kbd>Ctrl</kbd>+(<kbd>G</kbd>, <kbd>O</kbd>, <kbd>W</kbd>, <kbd>K</kbd>, <kbd>J</kbd>, <kbd>C</kbd>, <kbd>X</kbd>, <kbd>R</kbd>, <kbd>\\</kbd>, <kbd>U</kbd>, <kbd>T</kbd>, <kbd>_</kbd>, <kbd>V</kbd>) for essential functionality like reading and writing files and exiting the program. Other applications are similar, like `tmux` and `screen`, relying on keyboard shortcuts to manage the program's basic functionality.

This isn't an issue when using `hyper` on Macs, as the modifier prefix for interacting with the `hyper` windows is <kbd>Cmd</kbd>. On non-Macs (like Windows and Linux machines), the default prefix is `Ctrl`, which causes many conflicts.

This pull request proposes to address these conflicts by setting the default prefix on non-Macs to <kbd>Ctrl</kbd>+<kbd>Shift</kbd>, which is common for terminal emulators in linux. The <kbd>Super</kbd> key has been suggested as an alternative, but this would conflict with several shortcuts built into the Windows OS as well as many linux-focused desktop environment like Unity and Gnome. To avoid conflicts with defaults that already used <kbd>Ctrl</kbd>+<kbd>Shift</kbd>, some shortcuts were changed to (hopefully) reasonable alternatives. This pull request does not change any shortcuts on Macs.

### Summary of changes

1. Default shortcut prefix on non-Macs changed from <kbd>Ctrl</kbd> to <kbd>Ctrl</kbd>+<kbd>Shift</kbd>.

2. Some shortcuts that already used <kbd>Ctrl</kbd>+<kbd>Shift</kbd> were changed to be more similar to their Mac counterparts:
    - split vertically was changed from <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>E</kbd> to <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>D</kbd>

[...]

PR #1616, opened 2017-03-09
Improve Linux keyboard shortcuts

### LINKED ISSUE #1178: Change realod accelerator
- **OS version and name**: Ubuntu 14.04 LTS 64-bit
- **Hyper.app version**: 1.0.0

## Issue
Could we change the reload accelerator to something different than `Ctrl+R`? Because it's overwriting the reverse-i-search bash feature.

[https://www.gnu.org/software/bash/manual/html_node/Commands-For-History.html](https://www.gnu.org/software/bash/manual/html_node/Commands-For-History.html)

hharnisc:
This PR makes it a little nicer to use Linux (no idea if these shortcuts work on Windows) but it frees up `ctrl+r` for searching through previous commands and uses what seems to be the linux convention of cycling through tabs: `ctrl+PageUp/Down`. In it's current state I cannot use Hyper on Linux.

Sensitive to other open PRs (like #1509) that change a lot more, but it seems like there's still much to discuss there. Also this closes #1178.

### DIFF
--- a/app/accelerators.js
+++ b/app/accelerators.js
@@ -27,8 +27,8 @@ const applicationMenu = { // app/menu.js
   emojis: isMac ? 'Ctrl+Cmd+Space' : '',
 
   // View menu
-  reload: 'R',
-  fullReload: 'Shift+R',
+  reload: isMac ? 'R' : 'Alt+R',
+  fullReload: isMac ? 'Shift+R' : 'Alt+Shift+R',
   toggleDevTools: isMac ? 'Alt+I' : 'Shift+I',
   resetZoom: '0',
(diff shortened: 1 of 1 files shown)


rauchg:
@hbp would you agree with these shortcuts?

hharnisc:
Been using this pretty happily on Ubuntu since posting this PR. No issues so far.

MartyGentillon:
This is also under consideration in https://github.com/zeit/hyper/pull/1509 and https://github.com/zeit/hyper/pull/1191

ppot:
Cloing in favor of #1509

PR #1621, opened 2017-03-10
add custom inner children

henrikdahl:
Adds `customInnerChildren` prop to `decorateHyper` so that plugins can reside inside the `hyper-main` such as `hyper-statusline` plugin.

It's necessary to easily inherit styles applied to `hyper-main` such as the new `uiFontFamily` ✌️


rauchg:
Super +1. A note on `website` would be neat though.

Issue #1619, opened 2017-03-09
Echo text being included on .hyper.js edit using VIM


rizowski:
- [X] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [X] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: OSX Sierra V 10.12.3
- **Hyper.app version**: 1.3.0.1614 
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**:  https://gist.github.com/rizowski/dc846b7d73313ba7b4ee250d45fabf89 
- **Relevant information from devtools**: `Mac OS: CMD + ,`
- **The issue is reproducible in vanilla Hyper.app**: Not Vanilla

## Issue
When I hit `CMD + ,` to edit the preferences I get some of the echo statements prepended on top of my config. See screenshot:

![screenshot 2017-03-09 14 06 52](https://cloud.githubusercontent.com/assets/574835/23771038/b590b5f8-04d2-11e7-903b-9817cbcb9dda.png)

I am using `vim` as my editor. If I execute the command the hyper uses to open `.hyper.js` I do not get this. The command being `bash -c 'exec env ${EDITOR:=nano} ~/.hyper.js'`

This issue didn't start happening until I upgraded to 1.3.0.

boobiq:
same thing is happening with default EDITOR being nano
<img width="919" alt="screen shot 2017-03-10 at 11 08 58" src="https://cloud.githubusercontent.com/assets/429225/23791076/fcf331f0-0581-11e7-98df-676905d167ef.png">

henrikdahl:
Solved https://github.com/zeit/hyper/pull/1633 ✌️

Issue #1626, opened 2017-03-10
Error on loading configuration file


damienbidaud6:
Hello

Since the new reales 1.3.0 when I try to change my configuration file I have a error after saving the new file who say: 

`An error occurred loading your configuration (/Users/bidaudd/.hyper.js): Unexpected identifier`

And the change are not apply after that.

I try to reinstall the app and I still have the error 

I am on OSX 10.11.6
The version of the app is the 1.3.0
Link of a gist of my `.hypper.js` : https://gist.github.com/damienbidaud6/722ae817784e92ad10a0d76e2b9c033e 

Thanks for the help.

boobiq:
I believe it's because #1619 which adds text at the beginning of your configuration and it can't be parsed correctly.

damienbidaud6:
I remove the text and it's work but after I save it's add again. It's possible to disable this behaviour?

boobiq:
You have to remove that text it adds every time (they have to fix this bug) or you can edit **.hyper.js** file manually by running `vim ~/.hyper.js` (or any other editor), this bug only happens when you press **CMD+,** to open preferences

damienbidaud6:
Ok thanks

henrikdahl:
Closing since this is fixed in 1.3.1 ✌️

PR #1633, opened 2017-03-10
fix echo issue in config pref

henrikdahl:
Fixes https://github.com/zeit/hyper/issues/1619 issue with echo being prepended to config pref ✌️


henrikdahl:
Related https://github.com/zeit/hyper/issues/1626

matheuss:
🙏 ❤️

Stanzilla:
Even with 1.3.1 the echo appears twice 
hyper_2017-03-10_21-16-45

henrikdahl:
@Stanzilla that wasn't the bug this merge approached ✌️

PR #1644, opened 2017-03-12
Disable pop-up accent window on OS X

### LINKED ISSUE #1137: Press and hold keys on OSX causes suggestions popup
- **OS version and name**: OSX 10.12.1
- **Relevent information from devtools** _(CMD+SHIFT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: NA

## Issue
Just updated to 1.0.0, now when I'm in neovim if i press and hold down the 'L' key instead of repeating the movements, it also flickers the popup to enter 'ł'. It also seems to affect all keys as pressing and holding jkli keys seems sluggish. Was fine before this and is still fine in iterm. See attached gif which captures a subset of the popups flashing

press hold neovim

roolebo:
Before:
flickering-popus

After:
no-popups

Fix #1137

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -40,7 +40,7 @@ const {homedir} = require('os');
 
 // Packages
 const {parse: parseUrl} = require('url');
-const {app, BrowserWindow, shell, Menu} = require('electron');
+const {app, BrowserWindow, shell, Menu, systemPreferences} = require('electron');
 const {gitDescribe} = require('git-describe');
 const uuid = require('uuid');
 const fileUriToPath = require('file-uri-to-path');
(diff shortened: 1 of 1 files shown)


matheuss:
@roolebo for now I'll leave this as `blocked` because I'd like to fix this – I had it working while implementing the `hyper-caret`, as you can see in this gif i recorded back last year:
gif

chabou:
It would be great to make this accent keyboard working and maybe let user optionaly disabling it.
But for now, it is better to disabling it like iTerm.

app/index.js
```diff
@@ -215,6 +215,10 @@ app.on('ready', () => installDevExtensions(isDev).then(() => {
       } else {
         console.log('ignoring auto updates during dev');
       }
+      // Disable accent pop-up
+      if (process.platform === 'darwin') {
+        systemPreferences.setUserDefault('ApplePressAndHoldEnabled', 'boolean', false);
```
chabou: It is not the right place to do this. `setUserDefault`should be called only once and not at every window init.
Can you fix merge conflicts and move your block directly in `app.on('ready',...)` callback?

chabou:
This is not relevant anymore.
BTW thank you so much for your help and really sorry for answer delays 🙏

PR #1656, opened 2017-03-14
Fixed emoji spacing

douvaughn:
Fixed issue mentioned near the end of #154.  I am new to this codebase, so as far as I can tell this is ready to be merged.

Before:
before

After:
after

### DIFF
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -154,9 +154,8 @@ hterm.Terminal.prototype.setFontSize = function (px) {
   }
   unicodeNodeStyle.innerHTML = `
     .unicode-node {
-      display: inline-block;
       vertical-align: top;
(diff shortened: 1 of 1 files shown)


albinekb:
Hmm I think there was a reason for the style that was there before, can't remember what 🤔 
@rauchg what was the emoji-issue you fixed with the previous settings?

rauchg:
@douvaughn unfortunately this breaks monospacing. The solution that CLI program authors use is to add additional spaces after the emoji to compensate. This would make such programs show lots and lots of empty space.

AndrewSB:
@rauchg adding extra spaces doesn't fix the inconsistency between cursor position and text, it just further offsets it to the right.

@douvaughn did you end up fixing this another way?

I'm having this is a problem too

Issue #1588, opened 2017-02-27
Hyper should fallback to default config when not specified in `~/.hyper.js`


sindresorhus:
My `~./hyper.js` is:

```js
'use strict';
module.exports = {
	config: {
		fontSize: 16,
		fontFamily: 'Menlo'
	},
	plugins: [
		'hyper-snazzy'
	]
};
```

Since I deleted default config like the `css` option, that property is now `undefined` in `decorateConfig` for plugins. Hyper should provide default values in the core for all config properties.

For now, I'll have to specify an empty `css: ''` option so plugins/themes using `config.css` doesn't break.

JessDF:
Hello, I was wondering - Is anyone is working on this bug at the moment?

ppot:
Could be integrated into #1509

okonek:
I created PR

chabou:
Fixed by #2126 
Thank you so much @okonek ❤️

PR #1667, opened 2017-03-15
Fallback default config when not specified in `~/.hyper.js`

Erazihel:
As mentioned by @sindresorhus in #1588, there is no fallback for properties removed from the `~/.hyper.js` file. This results in plugins using those properties to break. To avoid that, the default configuration is used as a fallback for these removed properties.

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -123,7 +123,7 @@ exports.init = function () {
     try {
       console.log('attempting to write default config to', path);
       const defaultConfigContent = readFileSync(resolve(__dirname, 'config-default.js'));
-      exec(defaultConfigContent);
+      cfg = exec(defaultConfigContent);
 
       writeFileSync(
         path,
--- a/app/config.js
+++ b/app/config.js
@@ -7,7 +7,7 @@ const {dialog} = require('electron');
 const isDev = require('electron-is-dev');
 const gaze = require('gaze');
 const Config = require('electron-config');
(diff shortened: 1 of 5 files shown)


app/config.js
```diff
@@ -7,6 +7,7 @@ const {dialog} = require('electron');
 const isDev = require('electron-is-dev');
 const gaze = require('gaze');
 const Config = require('electron-config');
+const deepAssign = require('deep-assign'); // eslint-disable-line import/no-extraneous-dependencies
```
matheuss: Please remove that rule 😮

package.json
```diff
@@ -103,6 +103,7 @@
   "dependencies": {
     "aphrodite-simple": "0.4.1",
     "color": "0.11.4",
+    "deep-assign": "2.0.0",
```
matheuss: This should be in `app/package.json` 👌

app/config.js
Erazihel: Sorry about that, I didn't understand why I had this error occurring. The fact that is was in the wrong `package.json` explains it... :sweat_smile:

Erazihel:
Hey, any news for this PR ? :)

MartyGentillon:
This is fixed in https://github.com/zeit/hyper/pull/1509

albinekb:
Fixed in #1876

PR #1670, opened 2017-03-16
add unicode text as css class to unicode-node

berkin:
It is frustrating to fix the gap between powerline chars. I tried to fix it with patching the font over and over again, tried to fix the font with a font editor. No luck, the gap is still there. It is still seems like below.

image

Then i come with idea to make this chars with css. I have added a classname to the unicode wrapper and styled it with css like below.
```css

		x-row .unicode-node-,
		x-row .unicode-node- {
			text-indent: -9999px;
		}
		x-row .unicode-node-:after,
		x-row .unicode-node-:after {
			content: ' ';
			text-indent: 0;
			border-top: 10px solid transparent;
			border-bottom: 11px solid transparent;
			width: 0;
			height: 0;
		}
		x-row .unicode-node-:after {
			border-left: 7px solid;
			float: left;
		}
		x-row .unicode-node-:after {
			border-right: 7px solid;
			float: right;
		}
```
Works like a charm! 
image

### DIFF
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -165,7 +165,7 @@ const oldCreateContainer = hterm.TextAttributes.prototype.createContainer;
 hterm.TextAttributes.prototype.createContainer = function (text) {
   const container = oldCreateContainer.call(this, text);
   if (container.style && runes(text).length === 1 && containsNonLatinCodepoints(text)) {
-    container.className += ' unicode-node';
+    container.className += `unicode-node-${text} unicode-node`;
   }
   return container;
 };


albinekb:
Neat solution! Won't this however break the mono-space nature of these elements? 🤔  as mentioned  in here: https://github.com/zeit/hyper/pull/1656#issuecomment-286469053

berkin:
I have only added an extra class to unicode wrapper span, i have not changed the width or sth else. The css rules are from my .hyper.js file.

albinekb:
see #1536

albinekb:
Ah sorry @berkin, I think this is OK to merge, maybe we should add a config setting for it (off by default) because it will be unused by the majority of users, thus just adding extra stuff to the DOM

berkin:
@albinekb oke sounds good, i could add it to config setting like `addUnicodeCssClasses: false`

chabou:
@berkin can you test if #1536 fix it or not ?

berkin:
@chabou i can test it but my approach is different. When we add unicode class, these nodes would be easily customised with css.

albinekb:
I'm ok with this if you make it an option, default it to off. `addUnicodeCssClasses` sounds good

Stanzilla:
@berkin is this still need with the latest `canary` version?

berkin:
this pr became obsolete, it was based on hterm. in hterm, unicode characters were wrapped by a span, i was just adding another css class to it. so it needs to be implemented with xterm @Stanzilla

Stanzilla:
@berkin yup, figured. just weeding out a few old PRs.

PR #1680, opened 2017-03-17
Add onDecorated property to decorated components

chabou:
Plugins could be more powerful if we give them a way to retrieve a reference to decorated components.

The only reference that a plugin could have is `hterm`by passing a `onTerminal` property.
It could use `ref`property in decorators but if it is not the first in plugin list, it will receive previous plugin HOC.

With this PR, every plugin's `decorate<Component>` functions (decorateTerm, decorateTerms, decorateHeader, decorateTab...) could retrieve a reference to <Component> instance.

Maybe it is a little dangerous/anti-pattern to expose all instances to plugins, but this is the same logic than `onTerminal`property.

If you are ok with this feature, I'll add website documentation.

### DIFF
--- a/lib/utils/plugins.js
+++ b/lib/utils/plugins.js
@@ -370,9 +370,27 @@ export const middleware = store => next => action => {
   nextMiddleware(middlewares)(action);
 };
 
+// expose decorated component instance to the higher-order components
+function exposeDecorated(Component) {
+  return class extends React.Component {
+    constructor(props, context) {
+      super(props, context);
+      this.onRef = this.onRef.bind(this);
+    }
+    onRef(decorated) {
+      if (this.props.onDecorated) {
+        this.props.onDecorated(decorated);
+      }
+    }
+    render() {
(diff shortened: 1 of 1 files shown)


chabou:
@rauchg @matheuss @ppot could you give me your opinion about this feature ?

rauchg:
Wouldn't there be a way for getting the instance yourself when you decorate though?

chabou:
If you are the only (or the first) plugin, there is a way by passing : `ref`property when rendering Term because it is a first level child. But if you are the second plugin, you are not decorating Term itself but the first plugin decorator. If this first plugin doesn't keep a reference to Term instance (using `ref`property), I think there is no way to find it (but maybe I'm wrong).

chabou:
This reason leads to add `onTerminal()` capability to plugin decorator.
This PR generalize this capability (and could replace it) to all component that could be decorated.

chabou:
@rauchg I really think that this PR is a good idea 😊

rauchg:
I completely agree. Thanks for your explanation. We need to document this on the website, however.

Issue #1529, opened 2017-02-16
Weird behaviour when opening config on Windows


Stanzilla:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 8.1
- **Hyper.app version**: 1.2.1
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/Stanzilla/e776f86ed2b919fdfd7a2fc3d9d6de49
- **Relevant information from devtools** _(CMD+ALT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: 
- **The issue is reproducible in vanilla Hyper.app**: 

## Issue

![hyper_2017-02-16_19-10-00](https://cloud.githubusercontent.com/assets/75278/23034363/def87c58-f47b-11e6-9f88-532fc6253bef.png)

When opening the config, a new tab is opened and gets a lot of echos, is this the correct behaviour? From the text it is looking like it's failing a few times but notepad is actually opened.

Should opening the config even spawn a new tab at all?

stefanivic:
@Stanzilla you are opening the config using the Bash, that's why it only echoes the text as `start notepad` is not a valid bash command. Your best option is to manually go to `c:\\Users\\yourUser\\.hyper.js` and edit it.

jonathandelgado:
Looks like #1510 reverted the previous windows fix for this.

stefanivic:
@jonathandelgado it wasn't really a fix, it just opened the directory instead of directly opening the editor. I am trying to find a better fix as we speak.

Stanzilla:
@stefanivic it does that with cmd as well
![hyper_2017-02-20_00-47-17](https://cloud.githubusercontent.com/assets/75278/23107986/3e47521a-f706-11e6-9f6e-dd95def964c6.png)

Stanzilla:
Can the process of trying to open it be done in background though? Also the echo, currently looks like a debug print left there by accident, if you ask me :D

stefanivic:
@Stanzilla We got this merged : https://github.com/zeit/hyper/commit/36f96abb1d28ba85ebe678df2e87268b66aaf018 So now, it will check which shell you are runing and open with coresponding option. The `echo` is there just to notify you what is going on, but i will look into it, how we can make it nicer...

[...]

PR #1694, opened 2017-03-24
Fix open Preferences when using Powershell

kvakulo:
A more general approach to opening the preferences in Notepad that works from both cmd and Powershell on Windows.

I've tested it locally on Windows 10 when using cmd and Powershell, and I think it should be ready to merge 🤞 😄

### DIFF
--- a/lib/actions/ui.js
+++ b/lib/actions/ui.js
@@ -219,7 +219,7 @@ function getEditCommand(shell, isWin) {
   } else if (isWin) {
     return [
       ' echo Attempting to open .hyper.js with notepad',
-      ' start notepad "%userprofile%\\.hyper.js"'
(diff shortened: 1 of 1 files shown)


ppot:
What about Bash?

kvakulo:
That should already work, I guess. If you take a look in the ui.js file just above my patch there's a special case for Bash on Windows :)

MartyGentillon:
I think this solution is better: https://github.com/zeit/hyper/pull/1736

Related issue: https://github.com/zeit/hyper/issues/1529

chabou:
agreed with @MartyGentillon 
Discussed in #1529: adding `opn` could be even better than #1736.

@kvakulo sorry to not merge your PR 😢

PR #1698, opened 2017-03-24
check for null/undefined session in addSessionData

### LINKED ISSUE #1628: After some time after start rendering tab contents doesn't work properly

- **OS version and name**: macOS Sierra 10.12.3
- **Hyper.app version**: 1.3.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: vanilla config
- **Relevant information from devtools** _(CMD+ALT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: 

## Issue
I'm not sure how to reproduce it, but everytime I start Hyper, it works fine, but after some time (seconds/minutes) it stops rendering content of tabs. I can open/close new tabs and when I do something it executes normally (for example I echoed text to file and checked it in different terminal), but I can't see anything, only cursor

screen shot 2017-03-10 at 11 15 09

When I close all tabs and open new one, it again works fine. It didn't happen until I upgraded to 1.3.0

zsol:
Work around incorrect event ordering between `SESSION_PTY_EXIT` and `SESSION_ADD_DATA`. See #1628 for more context.
This is ready to merge. Fixes #1628.

### DIFF
--- a/lib/actions/sessions.js
+++ b/lib/actions/sessions.js
@@ -54,7 +54,7 @@ export function addSessionData(uid, data) {
       data,
       effect() {
         const session = getState().sessions.sessions[uid];
-        if (session !== null && session !== undefined) {
+        if (session) {
           const enterKey = data.indexOf('\n') > 0;
           const url = enterKey ? isUrl(session.shell, data) : null;
           if (url) {
--- a/lib/actions/sessions.js
+++ b/lib/actions/sessions.js
@@ -53,22 +53,23 @@ export function addSessionData(uid, data) {
       type: SESSION_ADD_DATA,
       data,
       effect() {
-        const {shell} = getState().sessions.sessions[uid];
-        const enterKey = data.indexOf('\n') > 0;
-        const url = enterKey ? isUrl(shell, data) : null;
-        if (url) {
-          dispatch({
-            type: SESSION_URL_SET,
-            uid,
-            url
-          });
-        } else {
-          dispatch({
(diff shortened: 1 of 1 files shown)


lib/actions/sessions.js
```diff
@@ -53,22 +53,23 @@ export function addSessionData(uid, data) {
...
-            url
-          });
-        } else {
-          dispatch({
-            type: SESSION_PTY_DATA,
-            uid,
-            data
-          });
+        const session = getState().sessions.sessions[uid];
+        if (session !== null && session !== undefined) {
```
ppot: if (session) should do the trick

zsol:
friendly ping? :)

andrroy:
@ppot For what its worth, I have tested this PR locally. Ref discussion in #1628, I am unable to run the current version of Hyper. When testing this fix, I no longer have that issue.

andrroy:
Great work btw, @zsol!

lib/actions/sessions.js
```diff
@@ -53,22 +53,23 @@ export function addSessionData(uid, data) {
...
+              uid,
+              url
+            });
+          }
         }
+        dispatch({
+          type: SESSION_PTY_DATA,
+          uid,
+          data
+        });
```
MartyGentillon: Why has this been moved outside of the else?  Will it happen twice if it isn't a URL?

zsol:
No it won't, notice the return in the then double if. I think it's cleaner
this way.

PR #1705, opened 2017-03-26
Fix partial TermGroup decoration

chabou:
First TermGroup of each window is decorated [here](https://github.com/zeit/hyper/blob/master/lib/components/terms.js#L6)

But when a TermGroup add a sub-TermGroup in a SplitPane, it is not decorated.

This PR fix this.

### DIFF
--- a/lib/components/term-group.js
+++ b/lib/components/term-group.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import {connect} from 'react-redux';
 import Component from '../component';
-import {decorate, getTermProps} from '../utils/plugins';
+import {decorate, getTermProps, getTermGroupProps} from '../utils/plugins';
 import {resizeTermGroup} from '../actions/term-groups';
 import Term_ from './term';
 import SplitPane_ from './split-pane';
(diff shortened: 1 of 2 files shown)


chabou:
Warning!
This PR could break plugins which assume that `decorateTermGroup` and `getTermGroupProps` are called only for each RootTermGroup.
It could be better to not merge this PR and warn somewhere in docs than only main TermGroup is decorated. I think that @ppot rework will fix this partial decoration.

chabou:
Maybe hyper-pane is the only plugin which assume that :rofl: 
In this case, I can fix it and we could merge this PR

rauchg:
@chabou, this is the behavior I'd expect. So let's take our chances!

chabou:
@rauchg I've scripted a check for all plugins listed in https://github.com/bnb/awesome-hyper

Only 8 plugins use `getTermGroupProps`. 7 use this hook only to pass down props without any action. So they'll work without problem. The only one which use some logic and will not work is mine: `hyper-pane`. I'll fix it very soon.

No plugin use `decorateTermGroup`.

We can conclude that this PR will not break anything 🎉

PR #1713, opened 2017-03-28
Simplify and correct newlines replacement logic on Windows

kvakulo:
New lines in the `.hyper.js` config file are not replaced correctly on Windows.

There's an error in the replacement logic: `x => x.indexOf('\r') < 0 ? x : '${x}\r'`. The test should actually be `x.indexOf('\r') >= 0`.

However, I've simplyfied the replacement logic to make it more clear what happens.

Should be ready to merge!

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -88,7 +88,7 @@ function exec(str) {
 // to text formatted with DOS line endings. We do this because the default
 // text editor on Windows (notepad) doesn't Deal with LF files. Still. In 2017.
 function crlfify(str) {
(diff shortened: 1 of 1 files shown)


albinekb:
Ah sweet! I saw this error when testing hyper on windows, tested this and can confirm it's working 🎉  ping @stefanivic

albinekb:
Tested by @stefanivic too, works  👌

albinekb:
Thanks @kvakulo 🎉

PR #1727, opened 2017-04-02
Get font color 'glow' functionally

### LINKED ISSUE #536: Improvement idea: activity "glow"
If a tab has an activity marker (becomes green due to output), it'd be cool to make it "glow" (ie: become brighter), when subsequent data events occurs.

mfix22:
This is complete and ready to merge, although I have an open question about specific implementation open within the issue it closes (#536).

I think this is the simplest way to create the glowing affect, and can be more finely tuned without updating the function simply by changing the number reducer (`ui`) counts up to (currently `5`). 

Let me know if you see any updates! I will edit this PR to list other PRs as I see them.

Closes #536

### DIFF
--- a/lib/components/tab.js
+++ b/lib/components/tab.js
@@ -55,7 +55,7 @@ export default class Tab extends Component {
         isFirst && 'first',
         isActive && 'active',
         isFirst && isActive && 'firstActive',
-        hasActivity && 'hasActivity'
+        hasActivity > 0 && 'hasActivity'
       )}
       >
       { this.props.customChildrenBefore }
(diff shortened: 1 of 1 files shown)


albinekb:
I think this is 🆗  to merge 👍

lib/components/tab.js
```diff
@@ -124,12 +124,7 @@ export default class Tab extends Component {
         }
       },
 
-      hasActivity: {
-        color: '#50E3C2',
-        ':hover': {
-          color: '#50E3C2'
-        }
-      },
+      hasActivity: getHeaderColor(this.props.hasActivity),
```
chabou: I tested your branch and this line produces an error when I open a tab:
capture d ecran 2017-05-20 a 00 10 47

`this.props` seems to be `undefined` in this `styles` function. It seems to be a empty/non-instance `Tab` component.
I don't know what I'm doing wrong.

albinekb: 🤔  Will have a look

albinekb: @mfix22 can you fix this?

chabou:
Sorry, I have to close this PR.
There is still a bug and this is not exactly what we want as explained in https://github.com/zeit/hyper/issues/536.

But thank you so much for your help @mfix22 ❤️

PR #1736, opened 2017-04-08
Fix open Preferences on Windows

gordon2012:
Able to merge.

This change causes Edit -> Preferences to open ~/.hyper.js with the system's default application. I tested it to work with cmd, git bash, ubuntu bash, and Powershell. Someone may want to check it on their machine in case of differences in my config.

### DIFF
--- a/lib/actions/ui.js
+++ b/lib/actions/ui.js
@@ -217,9 +217,15 @@ function getEditCommand(shell, isWin) {
       ' nano .hyper.js'
     ];
   } else if (isWin) {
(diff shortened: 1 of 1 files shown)


MartyGentillon:
You might also want to consider the discussion related to the https://github.com/sindresorhus/opn library in https://github.com/zeit/hyper/issues/1529

albinekb:
This has been fixed in https://github.com/zeit/hyper/pull/1784
Hyper will now open with your OS default application for .js

Issue #1738, opened 2017-04-09
Electron v1.6.x compatibility


chabou:
I'm on MacOS

[Master](https://github.com/zeit/hyper/commit/ec18537d9618ab0a9333ed3114028eb1b4ddf0b7) ~is currently~ was broken because of electron upgrade in [this commit](https://github.com/zeit/hyper/commit/b725969a3927694fe4c175b0b1ed78602c4c6fb4)

HTR:
* Checkout master and resintall dependencies (`rm -Rf node_modules && yarn`)
* Launch Hyper (`npm run dev` and `npm run app`)
* Open DevTools
* Hit `ctrl` key

An exception occurs:
![Closure Exception](https://cloud.githubusercontent.com/assets/4137761/24840240/ec100228-1d69-11e7-9de7-5f94955f475e.png)

It sounds crazy because code (and `const cache` declaration) is ok. I have checked with electron v1.4.16 and everything is allright!

I studied this issue in depth and I found a closure problem with `isAccelerator` function.
With v1.4.16:
![capture d ecran 2017-04-09 a 10 24 06](https://cloud.githubusercontent.com/assets/4137761/24840277/a301884e-1d6a-11e7-9a9f-838ff3141540.png)

But with 1.6.6 a closure (with cache variable) is missing:
![capture d ecran 2017-04-09 a 20 48 38](https://cloud.githubusercontent.com/assets/4137761/24840285/b6595e8a-1d6a-11e7-826d-d4f7528554f6.png)

I don't know yet if it is a V8 issue (I strongly doubt that).
But I found an [Electron related issue](https://github.com/electron/electron/issues/8854): webpack use some `eval`

If this issue is confirmed, due to its nature, we can suppose there is another issues with this same scope cause. I would recommend to stick to stable release channel of electron (v1.4.16).

MartyGentillon:
On windows, it appears that this causes ctrl character presses to be suppressed, making it impossible to escape programs with ctrl+c or such.

iam4x:
I can't reproduce anymore with Electron 1.6.6 which is stable.

MartyGentillon:
This is certainly not fixed in electron 1.6.6.  On windows 10 creators, after the upgrade, ctrl is still suppressed.

iam4x:
@MartyGentillon I was talking about the original issue posted for MacOS, might be something different for Windows10.

MartyGentillon:
Somehow I doubt that, as the error message is exactly the same.  Also, the
[...]

PR #1739, opened 2017-04-09
Downgrade Electron to 1.4.16

chabou:
Master is broken since Electron upgrade (1.6.x). See #1738.

This PR set Electron to the last stable v1.4.16


MartyGentillon:
This change fixes the issue on windows where all ctrl-character key presses aren't sent to underlying shell.

PR #1743, opened 2017-04-12
readme es-Es

leovafme:

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -88,7 +88,7 @@ function exec(str) {
 // to text formatted with DOS line endings. We do this because the default
 // text editor on Windows (notepad) doesn't Deal with LF files. Still. In 2017.
 function crlfify(str) {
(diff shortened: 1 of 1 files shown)


albinekb:
Cool!

- Do we really want this?
-  How will we manage to update it to keep it inline with the "real" readme?

Maybe it's better to tell the person to use google translate? :D

chabou:
Agree.
Maybe a link to separated self-hosted website (matained by another team) and with a disclamer (maybe not up-to-date) ?

leovafme:
Go to google overload is more tedious, you can add a link in the documentation that leads to the translations and the documentation is just the start since to hyper terminal we would add multi language :)

albinekb:
We won't be able to support this, the english documentation is already outdated. You could make a separate repo at your github and we could link it in the readme as unofficial-docs.

Thanks anyways!

Issue #1699, opened 2017-03-24
Hyper is reading `.hyper.js` 40 times a second, constantly.


stereokai:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 10 14393.953
- **Hyper.app version**: 1.3.1
- **[Link for .hyper.js](https://gist.github.com/stereokai/b377492613ad7f10073ce8049d04c8fb)**
- **Relevant information from devtools** _(CMD+ALT+I on Mac OS, CTRL+SHIFT+I elsewhere)_:
```
bundle.js:1 (re)loading renderer plugins
index.html:35 total init time 739.385
bundle.js:47 ignore pty exit: session removed by user
```
- **The issue is reproducible in vanilla Hyper.app**: YES

## Issue
I used ProcessMonitor and recorded Hyper reading the `.hyper.js` settings file on a loop. Why is it doing that? Seems wasteful and unnecessary. See the Gif I attached. The recording is filtered to Hyper.js events only. Check out the counter at the bottom: 28,000 Hyper.js events in appx. 3 minutes! Seeing that this is the same loop of 4 file system events, the math is 40 reads per second! What's the point?

![2017-03-24_21-45-50](https://cloud.githubusercontent.com/assets/2685357/24313175/246b9690-10dc-11e7-8bb8-762d736f0201.gif)

regisbsb:
It is dog poo slow on my pc as well. Dell XPS 15 16GB RAM i7 8 core M2 SSD DISK

ppot:
Closed in #1772

PR #1772, opened 2017-04-28
#1699 - add configurable interval to .hyper.js config file watch.

paulbouwer:
While looking at another issue related to `gaze` scanning the home folder on Windows, I fixed the following: 
- #1699 - Hyper is reading .hyper.js 40 times a second, constantly.

You can update the following in the `.hyper.js` config file to change the interval:

```
    // number of milliseconds to wait between checking if config file has been updated
    configFileWatchInterval: 5000
```

I have verified with ProcessMonitor on Windows, that the watch() read interval has been altered to the configured value.

This PR should be ready to merge.

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -39,7 +39,8 @@ const watchers = [];
 let cfg = {};
 
 function watch() {
-  gaze(path, function (err) {
+  // watch for changes on config every 2s
+  gaze(path, {interval: 2000}, function (err) {
     if (err) {
       throw err;
     }


rauchg:
@paulbouwer sounds like we should correct the default. Once every 2 second sounds reasonable?

rauchg:
Oops you did that exactly. I think maybe we should lower it to 2?

ppot:
This should not be in config. But strictly typed.

rauchg:
I think maybe @ppot would be right. If we have a reasonable default, do we need customization?

paulbouwer:
@rauchg @ppot Should I simply set the interval to 2s and remove from the config file? Happy to do this - 2s is a better default than what we currently have.

Or is there merit in setting the default to 2s and leaving it in config file for those people who really want to minimise disk access further?

ppot:
@paulbouwer Yes you definitely remove it from the config file. For the case of people wanting to change it's value. This can be modified later on if requests move along this way.

paulbouwer:
@ppot I've removed it from config and added 2s interval to **gaze** call.

```
function watch() {
  // watch for changes on config every 2s
  gaze(path, {interval: 2000}, function (err) {
```

It should be good to go.

ppot:
@paulbouwer As I said: Use a `const scan = 2000;` on the upper

stereokai:
I am the one who originally reported #1699 (measured Hyper to be accessing `.hyper.js` *40 times a second*).

It's great to see you guys addressing this issue. I really appreciate that! If I may, I would like to add that looking at the bigger picture, even 2 seconds is highly excessive. Who changes their settings so often?

Personally, I wouldn't even watch the settings file, but require users to manually refresh. Don't gasp yet. A terminal emulator's settings file is not a development environment, where live-reloading justifies the additional I/O burden.

ppot:
@stereokai Yes, reducing it to each min would be a better idea. Else user can manually refresh.

stereokai:
@paulbouwer @ppot is it possible to make this setting opt-out via config like you suggested at the [top of this page](https://github.com/zeit/hyper/pull/1772#issue-225013615)?

arvindch:
Is there any update on this? I would like to be able to disable auto-reload, as well.

PR #1775, opened 2017-04-28
Added windows junk to .gitignore

MartyGentillon:


matheuss:
This should be on your global gitignore 👌

Issue #1774, opened 2017-04-28
Compilation fails on Win8.1(+?) and Visual Studio because of old node-gyp


Stanzilla:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 8.1
- **Hyper.app version**: 1.3.1
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: 
- **Relevant information from devtools** _(CMD+ALT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: 
- **The issue is reproducible in vanilla Hyper.app**: 

## Issue

So I tried to compile `master` on my system and it was always failing with ` MSBUILD : error MSB4025: The project file could not be loaded. Root element is missing. ` which lead me to https://github.com/nodejs/node-gyp/issues/1144 which lead me to check the node-gyp version used in Hyper and find out that 3.4.0 is not compatible with Visual Studio 2017.

I'm not sure how to proceed though, just adding a dependency to node-gyp@3.6.0 obviously fixes it but is certainly wrong. The correct fix would be to bump `electron-rebuild` but the last release `1.5.7` only includes node-gyp 3.4.0 (https://github.com/electron/electron-rebuild/blob/v1.5.7/package.json#L38), but `master` has 3.6.0 (https://github.com/electron/electron-rebuild/blob/master/package.json#L37) which would make this work. Did I miss something or is this the only way to fix it?

So @paulcbetts in order to get node-gyp working with Visual Studio 2017 i had to use 3.6.0 will there be a new release of `electron-rebuild` soon and can Hyper adopt that new version?

paulcbetts:
Set your GYP_MSVS_VERSION=2015 in the meantime, if you have VS2015 installed. We can also probably do the bump in electron-rebuild

Stanzilla:
I don't, 2017 only atm. :/ Anyway, waiting for the bump then :)

PR #1783, opened 2017-05-03
Use the opn library to open the config file

Stanzilla:
Discussion is/was here https://github.com/zeit/hyper/issues/1529

Had to temporarily depend on a newer node-gyp because it would not compile otherwise (https://github.com/zeit/hyper/issues/1774)

### DIFF
--- a/lib/actions/ui.js
+++ b/lib/actions/ui.js
@@ -1,5 +1,3 @@
-import {EOL} from 'os';
-
 import {remote} from 'electron';
 import * as shellEscape from 'php-escape-shell';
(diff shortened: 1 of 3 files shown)


chabou:
`opn` doesn't work for me.
childprocess exits with code `1`
I'm trying to investigate why

chabou:
We can use Electron API instead of `opn`
Thank you for your help @Stanzilla
Sorry to close this in favor of #1784

Stanzilla:
All good, hope it gets merged soon :)

PR #1784, opened 2017-05-03
Open Hyper config in the native desktop's default manner

### LINKED ISSUE #1529: Weird behaviour when opening config on Windows

chabou:
This PR fixes #1529

It uses Electron API to open config file with OS default editor.
Bonus: Dev config file is opened if used.

We should test if no default app is set to open a `.js` file on MacOS/Windows/Linux

### DIFF
--- a/app/menu.js
+++ b/app/menu.js
@@ -3,6 +3,7 @@ const path = require('path');
 const {app, shell, dialog} = require('electron');
 
 const {accelerators} = require('./accelerators');
+const {getConfigDir} = require('./config');
 
 const isMac = process.platform === 'darwin';
 const appName = app.getName();
@@ -25,12 +26,9 @@ module.exports = ({createWindow, updatePlugins}) => {
       {
         label: 'Preferences...',
         accelerator: accelerators.preferences,
-        click(item, focusedWindow) {
-          if (focusedWindow) {
-            focusedWindow.rpc.emit('preferences');
-          } else {
-            createWindow(win => win.rpc.emit('preferences'));
-          }
+        click() {
+          const configFile = path.resolve(getConfigDir(), '.hyper.js');
+          shell.openItem(configFile);
         }
       },
       {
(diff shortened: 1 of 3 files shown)


ppot:
Tested and it open in atom since it's my default for Mac I think! Good.

Stanzilla:
Now we've come full circle :D I think default without config option > config option though. You would have to keep a fallback to default anyway in case a user uninstalls his preferred editor. Also can't think of an app that lets you specify an another app to edit its config.

chabou:
Maybe, it is not a good idea to have a `.js` suffix. `.hyper.config` could be better handled by OS.
If you want a specific app to edit it, you can open with this app directly. We don't need to specify it in config.

Our default behavior should be the OS one imo.

MartyGentillon:
@ppot were you hoping that it might open it with a text editor in terminal mode?  Because that doesn't seem likely if we use the electron API.

But it does give me an idea for a plugin.  It might be nice to have a plugin that allows you to choose from a configurable list of commands and have Hyper type them directly into the console for you (sort of like shell aliases, but with fuzzy matching command finding goodness).  Then you could use that plugin to open the config when you want to do something different.

MartyGentillon:
of course, you might be able to do something similar with peco....

ppot:
The first step is renaming `.hyper.js` to `hyper.conf`

albinekb:
Why change to `.conf` then my editor won't know it's JS @ppot

ppot:
@albinekb Look better :crossed_fingers:

toddself:
For some of us, we have set no default editor since our default editor is a terminal based application.

With my default windows setup, this tries to open ~/.hyper.js with Windows Script Host which is a really bad idea.

Stanzilla:
@toddself What would you expect it to do in this case?

Razzeee:
This is what happens then, btw
image

toddself:
@stanzilla it should show me the "you need to choose a program to open this file with". 

Ideally what I would expect is that a menu option called "settings" opens up in an editor built into the program. My initial thought on this was that something was broken with hyper and it was supposed to be loading a module and showing me a dialog.

Issue #541, opened 2016-08-01
Search feature is missing


rwngallego:
I really love hyperterm, do you maybe have plans to add a Search functionality? It would be really nice to have highlighting of the search occurrences also.
## 

HyperTerm 0.7.1
Electron 1.2.8
darwin x64 14.5.0

darol100:
cmd + f should open an windows where the user can search.

damienh:
tried cmd + f but nothing happens.

shrift:
This feature is probably a make or break feature for me. Until I have search, I'll probably stick with OSX's Terminal. :(

matheuss:
Bumped to `Priority: High` since it's a feature available in almost any app that displays some text 😛

shrift:
Thanks!

morgz:
+1

ronvenna:
+1

yongzhihuang:
+9001

aleclarson:
Can we lock this thread? I only want to ping'd when the search feature is added. 😜

matheuss:
zeit.chat is available for further discussion 😄

albinekb:
@timneutkens is working on this 👼

chabou:
There is a PR: https://github.com/zeit/hyper/pull/1793
And there is a plugin too: https://www.npmjs.com/package/hyper-search

Issue #140, opened 2016-07-16
Feature request: ctrl-f Find in Scrollback


cjb:
Ideally with incremental search and fancy highlighting and all that jazz.

Thanks for writing hyperterm! <3

chibicode:
I wonder if this can be done from a plugin somehow. One solution is to use `tmux`, but I've moved away from it...

atdrago:
Agree, this is a much needed feature. Very helpful when looking through diffs, and would bring Hyperterm closer to feature parity with Terminal.

emaillenin:
I am not able to get even the regular "Find" to work. Nothing happens with "Cmd + F".

seanyesmunt:
Any update on this? Like @emaillenin `Cmd + F` doesn't work for me.

dimitrieh:
Would love to have this, one of the most used features in iterm2

bobber205:
Not having this makes this almost unusable for me.

aleclarson:
This is the only reason I don't use hyper over the macOS terminal.

Any chance there will be work done on this soon? 😀

saketmehta:
If this is done, bye bye iTerm. But for now, I'm gonna stick with iTerm.

ephetic:
Gonna jump on the bandwagon here.  I tried hyper right after it came out, but switched back to iTerm after a couple weeks because of the lack of scrollback search.  Just tried it again for a week and I still can't live without it.  I could learn tmux, but then I probably could live without iTerm too.

stephencookdev:
Also in the boat of "really want to use Hyper, but literally cannot use a terminal without Ctrl+F"

I've raised a PR https://github.com/zeit/hyper/pull/1793 to add a very basic incremental search feature. Would appreciate reviews :)

jaanauati:
i just published this plugin, it implements  basic text search controls (input, buttons, menu 
 and shortcuts). Hope it helps u all:
http://npmjs.com/package/hyper-search

seanyesmunt:
@jaanauati Awesome! I'll be sure to try it out!

jcrben:
> I tried to work around this by selecting all text using Command+A and pasting it to a text editor to search, but Command+A doesn't seem to select text in scrollback.

@chibicode - Command+A seems to select text in the scrollback for me - using 2.0.0 - for those needing a workaround

jackHedaya:
[...]

PR #1793, opened 2017-05-07
Add basic search functionality, and a small UI bound to mod+f.

stephencookdev:
The search functionality supports single-line, case-sensitive, directional search - with single highlighted results at a time.

This should address https://github.com/zeit/hyper/issues/140 and https://github.com/zeit/hyper/issues/541

This is ready to merge, in theory - but I would be surprised if there are no comments surrounding:

- my use of hterm (never used the library before, I've rewritten some private methods to give us the functionality we need, but hopefully I'm missing something more elegant)
- my styling (I was unsure how to incorporate themes into this, so just tried to aim for a UI that would fit any theme)

I've also made this as minimal a solution as possible, definitely after this I'd really want to expand the feature to support case-non-sensitive / regex options, and also to support "(n results found)" and multi-result highlighting.

For now, though - here's what it looks like. Please check out the branch and try to break it :)

results

errors

screenshot from 2017-05-07 10-50-48

screenshot from 2017-05-07 10-50-59

screenshot from 2017-05-07 10-52-29

screenshot from 2017-05-07 10-52-38

Features:

- Mod+F to open find
- Mod+F/Esc to close find
- Click "Find" or push "Enter" to find (cycling through 1 at a time) matching search
- Push Shift+Enter to cycle through in reverse
- Searching is per terminal, so only the focused tab/pane/window is actually searched


ppot:
When you are searching. You need to select every result at the same time. and using `enter` key to cycle between them. `cursor select the last one` then move to bottom and go back to previous find from each enter key.

ppot:
For the hterm lib part. Move you're code to utils/search.js 

export functionality and set them to hterm after. Something like:

`hterm.Terminal.prototype.find  = search.find(searchTerm, options = {})`

stephencookdev:
@ppot so currently yes, it's single-select, but pushing enter cycles through the results. Pushing shift+enter cycles through the results in reverse.

And panes are a good point, I should have mentioned. The UI looks exactly the same, and just only the active panes are searched (the same is true of tabs and windows). The search bar is positioned in the top right of the window, not the top right of the pane. I guess this could be confusing, but in my mind it's the window that has a search bar, not the pane.

I'll update the PR description with that^ info.

As for moving code to `utils/search.js` - what's the reasoning there? It seems unnecessary to declare a utility function, and then bind it to the `Terminal` prototype anyway. Any time anyone wanted to use the utility function `find(terminal, str)` they could also just do `terminal.find(str)`. So the API is exactly the same on the `Terminal` prototype, just with more duplication

ppot:
For moving those function to utils/search.js is to reduce the complexity of hterm.js file by moving specific parts to their file. But it's okay if it get to big it can be moved afterwards.

app/accelerators.js
```diff
@@ -24,6 +24,7 @@ const applicationMenu = { // app/menu.js
   paste: 'V',
   selectAll: 'A',
   clear: 'K',
+  find: 'F',
```
ppot: you have to `find: 'F'`

stephencookdev:
Okay @ppot - I've moved that logic out to utils/ in d6c44cec67e258e54da86fc7284f9e09b29a5021 :)

stephencookdev:
Any more comments/reviews on this...?

ppot:
Been busy sorry, @albinekb ?

lib/utils/search.js
```diff
@@ -0,0 +1,55 @@
...
+    terminal.scrollPort_.scrollTo(curRow, 5);
+    terminal.scrollPort_.scheduleRedraw();
+
+    // Make sure the redraw has happened before attempting to highlight the result
+    setTimeout(() => terminal.selectRow(curRow, foundIndex, foundIndex + searchTerm.length), 0);
+
+    return true;
+  }
+  return false;
+}
```
MartyGentillon: 50 line functions are a clear code smell.  This needs refactoring.

lib/utils/search.js
```diff
@@ -0,0 +1,55 @@
...
+  let focusRowNode = sel.focusNode;
+  while (focusRowNode && !('rowIndex' in focusRowNode)) {
+    focusRowNode = focusRowNode.parentNode;
+  }
+
+  // get anchor row
+  let anchorRowNode = sel.anchorNode;
+  while (anchorRowNode && !('rowIndex' in anchorRowNode)) {
+    anchorRowNode = anchorRowNode.parentNode;
+  }
```
MartyGentillon: Duplicate code

lib/components/searchbox.js
```diff
@@ -0,0 +1,190 @@
...
+    e.preventDefault();
+
+    // If a submit button was selected, return focus back to the input box
+    this.searchBox.focus();
+
+    const term = this.props.terms && this.props.terms.getActiveTerm();
+    if (term) {
+      const searchTerm = this.state.searchTerm;
+      const options = {
+        dir: this.state.dir
```
MartyGentillon: dir is a poor name, do you mean directory or direction?  (I am going to guess directory every time, especially when working with a console application).  Avoid abbreviations, it's more important that your code be easily readable than easily typable.

lib/utils/search.js
```diff
@@ -0,0 +1,55 @@
...
+  const selNodeOffset = (
+    startRowNode &&
+    startRowNode.textContent.indexOf((dir === 1 ? sel.focusNode : sel.anchorNode).textContent)
+  ) || 0;
+  let curRowOffset = selNodeOffset + (dir === 1 ? sel.focusOffset : sel.anchorOffset - 1);
+  const rowCount = terminal.getRowCount();
+
+  let foundIndex;
+  let curRow;
+  for (let i = 0; i <= rowCount; i++) {
```
MartyGentillon: Extract this for loop to it's own function.

lib/utils/search.js
```diff
@@ -0,0 +1,55 @@
...
+  if (foundIndex !== -1) {
+    terminal.scrollPort_.scrollTo(curRow, 5);
+    terminal.scrollPort_.scheduleRedraw();
+
+    // Make sure the redraw has happened before attempting to highlight the result
+    setTimeout(() => terminal.selectRow(curRow, foundIndex, foundIndex + searchTerm.length), 0);
+
+    return true;
+  }
+  return false;
```
MartyGentillon: The basic find function should probably just return the detailed results of the search.  That will allow it to be composed into other useful functions.  Leave the scrolling and highlighting to another function.

lib/components/searchbox.js
```diff
@@ -0,0 +1,190 @@
...
+      };
+      const searchResult = term.find(searchTerm, options);
+      this.setState({searchError: !searchResult});
```
MartyGentillon: How are you handling wraparound for the search?

lib/components/searchbox.js
```diff
@@ -0,0 +1,190 @@
...
+      noResultsText: {
+        color: '#000',
+        fontSize: '80%',
+        width: '90px',
+        marginBottom: '2px'
+      },
+
+      searchInput: {
+        fontSize: '16px',
+        padding: '10px',
```
MartyGentillon: This should scale with the user interface.

lib/components/searchbox.js
```diff
@@ -0,0 +1,190 @@
...
+      },
+
+      searchButton: {
+        display: 'block',
+        lineHeight: '100%',
+        margin: '10px 0',
+        padding: '0 15px',
+        background: 'transparent',
+        color: '#555',
+        fontSize: '16px',
```
MartyGentillon: This should scale with the user interface.

lib/utils/selection.js
```diff
@@ -50,4 +52,72 @@ exports.all = function (terminal) {
...
+      row = row.nextSibling;
+    } else {
+      return -1;
+    }
+  }
+  return getNodeAndOffsetWithinRow(row, position);
+}
+
+/*
+ * Copy-pasted from hterm's hterm.Screen.prototype.getNodeAndOffsetWithinRow_
```
MartyGentillon: I have to wonder if, rather than copying and pasting the code (duplicate code is a great way to make sure that a bug doesn't really get fixed), could you just use the function declared there?

something like
```
var GetNodeAndOffsetWithOverflow = hterm.Screen.prototype.getNodeAndOffsetWithOverflow_
```

lib/components/searchbox.js
```diff
@@ -0,0 +1,190 @@
...
+    </form>);
+  }
+
+  styles() {
+    return {
+      searchForm: {
+        position: 'absolute',
+        top: '50px',
+        right: '50px',
+        width: '400px',
```
MartyGentillon: This should probably scale with the user interface.

lib/components/searchbox.js
stephencookdev: Do you mean terminal wraparound (not line wrap-around?)

Currently the `find` method will always find the next item, and wraparound implicitly with no feedback to the user.

Ideally there should be a notification of this wraparound having happened to the user, but I want to address this in a separate PR

lib/components/searchbox.js
stephencookdev: Yeah, you're right. Do you know of any examples in the codebase where logic like this happens? If not off the top of your head, no worries - I'll dig around

lib/utils/selection.js
stephencookdev: I could, and I originally had something similar when developing

But I really don't like the idea of adding an explicit dependency to private methods in hterm. The maintainers of hterm would be well within their rights to remove/rename the private method, and mark it as only a minor change

lib/components/searchbox.js
MartyGentillon: I would think that using em would cause it to happen magically.

lib/components/searchbox.js
MartyGentillon: Cool, glad to hear you are thinking about it.

lib/utils/selection.js
MartyGentillon: Yep, this is a situation where there isn't really any good answer....  On one hand, you would sort of like to pick up their bug fixes, on the other hand, it is always best to avoid relying on internals.  I suppose this is probably the better compromise, though I still stubbornly wonder if there isn't a way to do both.  (Perhaps we should file a feature request on them?)

MartyGentillon: Here is a thought for a different compromise, use the function declared there and write a test to make sure that it is there.  Now, if they remove/rename the function, we will know about it, while if they fix a bug, we will get it.

oalhait:
Any updates on this/idea on when it'll get pushed to users?

stephencookdev:
I should have time to finish addressing comments + resolving merge conflicts tonight

lib/utils/selection.js
stephencookdev: I get what you're saying, and I can see the benefit - it just feels really weird to me to do this.

I would much rather go the route of taking the (fairly small) duplication hit for now, and lobby hterm to make those methods slightly more public

ppot:
Before fixing thing. You should rebase with master since lots of thing changed.

stephencookdev:
@ppot yeah, rebased - everything looks like it's working, after a bit of a fix up

@MartyGentillon should have addressed all of your comments.

UTs still to come, I didn't realise quite how bare the UTs were! :(

lib/utils/selection.js
MartyGentillon: Either or, for the most part, this is about considering options.  Whichever you implement will work fine.

lib/utils/search.js
```diff
@@ -0,0 +1,58 @@
...
+    return null;
+  }
+
+  return {
+    index: foundIndex,
+    row: foundRow
+  };
+}
+
+// find a searchTerm, highlighting it (and scrolling to it if currently off screen)
```
MartyGentillon: This comment needs to be updated now (you moved the scrolling away from this function.)

stephencookdev:
Okay @MartyGentillon I addressed that last comment

I looked again at adding UTs, but I really think it's gonna make more sense to address this in a separate PR, specifically looking at building up the UT framework here

It looks like the current ava set-up isn't even set up to handle the ES6+ syntax used in the project :/ A lot of the decisions

MartyGentillon:
Looks good to me.

chabou:
It works really great !
Awesome work @stephencookdev!

stephencookdev:
@chabou awesome, thank you :)

Is there anything else we need to do to get this merged?

chabou:
Just for the record, a plugin exists: https://www.npmjs.com/package/hyper-search

ItamarGronich:
Hey, guys? what's the preventing this from merging? i don't want to install a plugin that will later conflict with the build when this gets merged.

jq-87:
The previously mentioned plugin seems to have issues - it does not work for me and other people are also saying the same thing.

They have an open issue:
https://github.com/jaanauati/hyper-search/issues/12

So hopefully this feature gets merged in soon 🙏

stephencookdev:
Fixed the merge conflicts

Just a note that this PR has been open for 5 months now, though. I'm not sure who the best people to poke are to look at this, but if anyone knows - please poke away... :)

stshan:
Any updates on this? It seems kinda weird that such an important feature PR would be ignored for so long.

pratik0809:
+1 would like to see this feature merged in

david-a:
+1 for this one! please merge !

stshan:
ping

chabou:
We do not use `hterm` and DOM anymore, but `xterm` and `canvas`.

This PR will not work as it with canary, `hyper-search` plugin neither.

We maybe should use xterm search addon. https://github.com/xtermjs/xterm.js/tree/master/src/addons/search

@stephencookdev I am very sorry to close this PR 😞  And thank you so much for your help 🙏

rightaway:
So there's really no way to search at all in canary, using a plugin or not?

albinekb:
Not yet @rightaway, but as chabou stated it is possible to do this a lot easier now with xterm & xterm search addon.

RichardBronosky:
"Possible" in January. But now it's September.

albinekb:
Yes, PR welcome 👍  @RichardBronosky We are all busy.

karlhorky:
Looks like #3075 ended up fixing this and also #541 .

PR #1796, opened 2017-05-07
re-implement `clear`, make it not buggy, fix CMD+K

rauchg:
This makes everything better

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -177,14 +177,7 @@ export default class Term extends Component {
   }
 
   clear() {
-    this.term.clearPreserveCursorRow();
-
-    // If cursor is still not at the top, a command is probably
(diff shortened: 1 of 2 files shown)


ppot:
Sure?

albinekb:
😂  awesome

dcalhoun:
@rauchg were there specific issues/bugs that this resolved? I'm curious as to what this changed. Thanks.

dcalhoun:
@rauchg to follow up on my previous comment, I ask specifically because I still see a lot of issues with clear in regards to using [Jest](https://facebook.github.io/jest/). When using Jest's `--watch` mode, the runner always produces odd output (e.g. the history of previous runs is accessible if you scroll upwards).

Issue #1808, opened 2017-05-10
Cannot clear terminal with Command+K since latest version (1.3.2.1744)


saadq:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: macOS Sierra Version 10.12.2
- **Hyper.app version**: Version 1.3.2 (1.3.2.1744)
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: [Link](https://gist.github.com/anonymous/179755edd536fdc88eb7137b3abce058)
- **Relevant information from devtools** No errors in console.
- **The issue is reproducible in vanilla Hyper.app**: The only plugin I have is a theme.

## Issue

Ever since the latest update, it seems I can't clear the terminal with <kbd>command</kbd>+<kbd>K</kbd>. This issue only happens when I have done `npm start` to start webpack for one of my apps. Currently, whenever I press <kbd>command</kbd>+<kbd>K</kbd>, it makes `^L` appear at the bottom of the terminal.

Here is a screenshot of the issue: (You can see the `^L` come everytime I try to clear at the bottom):

![screen shot 2017-05-10 at 4 43 27 am](https://cloud.githubusercontent.com/assets/5678694/25890373/3a6adcde-353b-11e7-852a-601b9e6516ef.png)

EDIT: Just to provide some more info, if I exit out of the process with <kbd>control</kbd>+<kbd>C</kbd>, and then run the clear with <kbd>command</kbd>+<kbd>K</kbd> it works fine.

shezarkhani:
Same issue when running `tail -f`. Also, since the update CMD+K does not completely clear the buffer. I know this behavior is expected, but sometimes it is useful to actually clear the terminal completely (for example long running tails of logs, or running a `cat` of a huge file). Maybe this behavior should be a configuration.

bmrobin:
[looking at the PR](https://github.com/zeit/hyper/pull/1796/files), it seems that they did account for this before and are simply disabling it by default now.

pedroneves:
It's happening on windows 8.1 as well

fsegouin:
Same issue when running any command that would require a redraw on the last available line, it just goes up and up and up instead of clearing the previous line.

jesseditson:
[...]

PR #1811, opened 2017-05-10
wipe contents before \k

ppot:
Wipe contents before `\k`.

-> Still input `^L` don't know why. ( `^L` = result of commands clear ) 

FIX: 
#1808

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -177,6 +177,7 @@ export default class Term extends Component {
   }
 
   clear() {
+    this.term.wipeContents();
     this.term.onVTKeystroke('\f');
   }
 


stephan281094:
Thanks for the PR. Just tested this and it works, but it completely clears the screen, including the prompt, which shouldn't be cleared.

I think the `wipeContents` function should only be called when there's no prompt active. Likewise, the `onVTKeystroke('\f');` function should only be called when there **is** a prompt active. Not sure how to implement this though.

ppot:
@stephan281094 Can you show me a picture of what you mean. On my side it looked fine.

stephan281094:
Current behavior (with `wipeContents`):
with-fix
If you look closely, you can see the prompt disappear for a short time. This doesn't happen in the current version of Hyper.

curz46:
Perhaps this issue can be fixed by simply sending `\r` after clearing? That should trigger a new prompt to be rendered.

rauchg:
Are you using vi mode by any chance? You have to configure `.inputrc` if so

stephan281094:
No, I'm not using vi mode.

krainboltgreene:
What can I do to help this PR get in and be released?

stephan281094:
In my opinion this PR can be merged as it fixes clearing the screen when using `tail`/ `webpack`/ whatever.

PR #1826, opened 2017-05-14
Show plugin (name and version) loaded

chabou:
Currently, there is no way to know which plugin is loaded and its version.
I begun to add a `console.log` to my plugins but I realized it can be a useful information and should be a Hyper feature.

This PR add these informations:
* in shell in dev mode (from electron):
capture d ecran 2017-05-14 a 22 12 29
* in devtools (from renderer)
capture d ecran 2017-05-14 a 22 12 49
* in About dialog
capture d ecran 2017-05-14 a 22 07 17
capture d ecran 2017-05-14 a 22 07 43

To show these information in about dialog, I needed to use a custom Dialog on MacOS like on Windows/Linux. The only downside is that I can't show build number like currently:
capture d ecran 2017-05-14 a 22 15 37

But now, MacOS About dialog is consistant with other OS.

@rauchg: I don't know if you want to add Zeit Copyright (but it is not currently shown on Windows/Linux).


app/index.js
```diff
@@ -395,7 +395,8 @@ app.on('ready', () => installDevExtensions(isDev).then(() => {
       createWindow,
       updatePlugins: () => {
         plugins.updatePlugins({force: true});
-      }
+      },
+      getPlugins: plugins.getModules
```
albinekb: Why do you rename it from `getModules` to `getPlugins` 🤔  I think it's better to have the same name so it's easier to read the code :)

app/menu.js
```diff
@@ -7,18 +7,35 @@ const {getConfigDir} = require('./config');
...
 const appName = app.getName();
+const appVersion = app.getVersion();
 
 // based on and inspired by
 // https://github.com/sindresorhus/anatine/blob/master/menu.js
 
-module.exports = ({createWindow, updatePlugins}) => {
+module.exports = ({createWindow, updatePlugins, getPlugins}) => {
+  const showAbout = () => {
+    const pluginList = getPlugins().length === 0 ?
```
albinekb: save the plugins in a variable
`const plugins = getPlugins()`
at the top of this function, that way you avoid calling it multiple times 👍

lib/utils/plugins.js
```diff
@@ -69,6 +69,7 @@ const clearModulesCache = () => {
 };
 
 const getPluginName = path => window.require('path').basename(path);
+const getPluginVersion = path => window.require(window.require('path').resolve(path, 'package.json')).version;
```
albinekb: This should probably be in a try/catch, just to be safe 😄  maybe if a plugin is not installed yet etc?

albinekb: Also I'd import `path` to a variable
```
const path = window.require('path')
```
not really necessary since require caches it but looks cleaner

app/index.js
chabou: Because everywhere in plugin file, modules are mentioned. And `plugins.getPlugins` seems weird to me.
But you're right, it's really better to not mention `module` in exports.

app/menu.js
chabou: oups... Totally agreed

lib/utils/plugins.js
chabou: You're right, a local plugin without`package.json` file can be used.

chabou: You mean like that https://github.com/zeit/hyper/blob/1.3.3/app/plugins.js#L129-L142
😰 
😄

app/index.js
chabou: `getLoadedPlugins` will be a better name and it will export only plugin name and version. Not the entire module.

chabou:
I was certainly tired to make a PR of such poor quality. 😰 
Thank you so much @albinekb for your insights.

app/index.js
albinekb: awesome @chabou  🌮

app/plugins.js
```diff
@@ -307,6 +311,13 @@ function requirePlugins() {
 
       // populate the name for internal errors here
       mod._name = basename(path);
+      try {
+        // eslint-disable-next-line import/no-dynamic-require
+        mod._version = require(resolve(path, 'package.json')).version;
```
albinekb: this will be `sync` and block other stuff, better to do something like:
```js
const getVersion = (path) => new Promise((resolve, reject) => {
  fs.readFile(resolve(path, 'package.json'), (error, result) => {
    if (error) return reject(error)
    try {
      const { version } = JSON.parse(result.toString())
      return resolve(version)
    } catch (error) {
      return reject(error)
    }
  })
})

```
or just use https://github.com/sindresorhus/load-json-file 😄

lib/utils/plugins.js
```diff
@@ -68,7 +68,19 @@ const clearModulesCache = () => {
...
 
-const getPluginName = path => window.require('path').basename(path);
+const pathModule = window.require('path');
+
+const getPluginName = path => pathModule.basename(path);
+
+const getPluginVersion = path => {
+  let version = null;
+  try {
+    version = window.require(pathModule.resolve(path, 'package.json')).version;
```
albinekb: same here: https://github.com/zeit/hyper/pull/1826/files#r117410691

app/plugins.js
albinekb: maybe not an issue since it's (maybe?) cached, and looking at your screenshot it doesn't even take 10ms

chabou: yes I think it is not an issue, I measure it in renderer with 3 plugins without requiring `package.json`: 8ms.
With requiring: 12ms.

I don't want to add asynchronous complexity here and in the renderer for such a little gain.

albinekb: Indeed

albinekb:
Great work @chabou ❤️

chabou:
rebased and Copyright added
capture d ecran 2017-06-08 a 00 17 32

Ready to merge

albinekb:
Is it OK'd by the @zeit team?

chabou:
It seems impossible to easily add link to this Electron MessageBox.

Issue #1734, opened 2017-04-07
RightClick Context Menu after uninstalling the app still contains the Hyper link


fuaxio:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**:  Windows 7 SP1
- **Hyper.app version**:  1.3.1

## Issue
RightClick Context Menu after uninstalling the app still contains the Hyper link.
Any way to remove it?
![wwss](https://cloud.githubusercontent.com/assets/13122476/24818232/30a24e50-1bfd-11e7-80b4-072c639021a7.png)

fuaxio:
nvm solved the issue.
But installer needs to be improved

btd:
This issue still actual with 1.3.3, just got it at Win 10

fuaxio:
@btd you need to tweak some reg values by searching for "hyper". Do it very cautiously as a wrong reg value deleted might mess your system.

Issue #1668, opened 2017-03-15
Windows 10 context menu Open Hyper here remains after uninstalling


laur1s:
extremely annoying, how can I fix that?
Windows 10 x64

albinekb:
Here's a guide to do that: http://www.thewindowsclub.com/remove-click-context-menu-items-editors

@laur1s 

We should make hyper clean that up on uninstall, I'm not too into the windows side of hyper/electron, but feels like it should be supported.

laur1s:
The uninstall leaves so many things, that should be definitely removed. I had to remove config files, plugins, etc., manually. Furthermore, context menu did not reset after reinstallation, and  I still can't open hyper from it. I guess I'll have to use regedit for fixing that

chabou:
Fixed by #1831

Issue #1651, opened 2017-03-13
Lost context menu option after update


aulneau:
- **OS version and name**: Windows 10 64-bit
- **Hyper.app version**: 1.3.1

Hey there!

Love the work you all do. Just wanted to give you a heads up that when I updated through the notification in Hyper, I lost my context menu functionality. (Open Hyper here no longer works)

![hyper-context](https://cloud.githubusercontent.com/assets/11803153/23867639/3156fc3a-07e2-11e7-8d37-95510bb724da.PNG)

Thanks!

-a

ChiefORZ:
Yah, defenenitely a bug. i always have to re-add the Registry Entry `REG_SZ="C:\Users\%username%\AppData\Local\hyper\app-1.3.1\Hyper.exe" "%V"` at `\HKEY_CLASSES_ROOT\Directory\Background\shell\`

chabou:
Fixed by #1831

Issue #1133, opened 2016-12-12
[Bug][Windows] Context menu is showing Hyper but trows an error when clicked.


stefanivic:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 10 Pro, Version 1607, OS Build 14393.479
- **Hyper.app version**: Hyper 1.0.0
- **The issue is reproducible in vanilla Hyper.app**: Yes

## Issue

Steps to reproduce : 
1. Clean install Hyper 1.0.0.
2. Right click on desktop.
3. Click on Hyper.

Images off the issue : 
![](https://i.gyazo.com/0b4ff19bc68a7f4769787457f91f9f71.png)
![](https://i.gyazo.com/188e640d7dcae281b3d4e5a4fc2f5654.png)

klauscfhq:
I am also on Windows 10 Pro - Version **1607** - OS Build **14393.479** - and the issue is not present 🤔
Maybe try running the installer as administrator, if you have not already.
Also, the right click prompt to run Hyper, is kinda different for me;
![untitled](https://cloud.githubusercontent.com/assets/12670537/21120383/b7d79ea4-c0cf-11e6-9d87-7c89e91403ed.png)
Hope that helps!

berwyn:
Windows 10 Pro 1607 Build 14965.1001 here, and my "open here" option works correctly, but also matches klauscfhq's
![image](https://cloud.githubusercontent.com/assets/1403744/21120894/cc60d758-c086-11e6-85c8-369cc55979dc.png)

stefanivic:
Yeah, looks like something happened during the first install. I tried again as administrator but the issue was still there. Had to remove hyper from the context menu first and reinstall as administrator.  

@CodeTheory you can close the issue.

bayesrule:
same bug here, Win 7.
Wondering how to remove hyper from the context menu.... @stefanivic @CodeTheory

lednhatkhanh:
Same bug here, trying to remove it and rerun with administrator.

bayesrule:
Also, tried to uninstall Hyper from Win 7. The uninstallation is not clean at all...

lednhatkhanh:
[...]

PR #1831, opened 2017-05-15
[Windows]Refactor system context menu

Specro:
Refactored the system context menu code. Now it's less complex and easier to debug in the future. This should fix all these #1734 #1668 #1651 #1133 bugs as far as my testing goes. Any more testing from other Win users always appreciated 😄 Otherwise it's ready for merge.

Open Hyper here

### DIFF
--- a/app/system-context-menu.js
+++ b/app/system-context-menu.js
@@ -8,42 +8,70 @@ const regParts = [
     {name: 'Icon', value: `${appPath}`}
 ];
 
-function isRegistered(callback) {
-  new Registry({
-    hive: 'HKCU',
-    key: `${regKey}\\${regParts[0].key}`
-  }).get(regParts[0].name, (err, entry) => {
-    callback(!err && entry && entry.value === regParts[0].value);
+function addValues(hyperKey, commandKey, callback) {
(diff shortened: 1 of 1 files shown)


ppot:
As far as I am concern. I can't test on windows. Could you make a gif of the result (context) or pictures here.

Specro:
It looks the same as before. I added the picture nevertheless.

app/system-context-menu.js
```diff
@@ -8,42 +8,70 @@ const regParts = [
...
+          addValues(hyperKey, commandKey, callback);
         });
       });
-      Promise.all(regPromises).then(() => callback());
     }
   });
 };
 
 exports.remove = function (callback) {
-  isRegistered(registered => {
```
albinekb: Why remove this check? 🤔

Specro: It's not really needed since the destroy function returns an error if it can't find the Hyper key or something else goes wrong. Helps to debug if we get these menu bugs again. Also, less registry requests - faster (un)installation.

albinekb: Cool

chabou:
Tested on Win10. Works really great! No more problem with install/uninstall.
It's a good thing that registry key is rewritten at each Hyper update.

Issue #628, opened 2016-08-17
Line at the bottom of the terminal is not aligned properly


tleunen:
The last line of my terminal is sometimes "cut", like we don't see the bottom of some characters like (`q` or `p`).
<img width="128" alt="screen shot 2016-08-17 at 11 02 04 am" src="https://cloud.githubusercontent.com/assets/1972567/17741546/9af09e42-646a-11e6-8298-63ab8d998e75.png">

It's not always the case in all my terminal windows, maybe it's related when the terminal has tabs. Will try to find better repro.

jamox:
That's even worse than what I'm seeing (in #373)

jairtrejo:
Hello! Just wanted to report the same problem, it happens to me with powerline for tmux (pretty bad)
<img width="813" alt="screenshot 2016-08-30 10 41 55" src="https://cloud.githubusercontent.com/assets/210225/18116527/587d0526-6efd-11e6-996c-6b72d2bf8df4.png">
or with just tmux (not as bad, but still misaligned):
<img width="773" alt="screenshot 2016-08-30 22 00 35" src="https://cloud.githubusercontent.com/assets/210225/18116534/630ffe6c-6efd-11e6-9c29-5fa585774253.png">

fingermark:
Can confirm that this happens with "Times New Roman" as well.  The only fix I found was to adjust the line-height, as suggested by @davegomez in my "duplicate" ticket.

```
termCSS: 'x-row {line-height: 1}',
```

jamox:
Awesome, I can confirm that it fixed it for me!

tleunen:
Tested with the recent 0.8.1 version and still have the issue.
<img width="220" alt="screen shot 2016-10-08 at 4 29 40 pm" src="https://cloud.githubusercontent.com/assets/1972567/19215746/72257446-8d74-11e6-991c-367a7eaf7e7b.png">

```
fontSize: 12,
fontFamily: 'Menlo, "DejaVu Sans Mono", "Lucida Console", monospace',
```

And I also confirm that changing the line-height fixes the issue.

davegomez:
I can't reproduce the issue @tleunen. Have you tried disabling your extensions in case you are using one or maybe tweaking you configuration (using the default one), etc?

I have tried different fonts (even the ones you mention) and  themes but nothing.

tleunen:
It's not an issue that occurs 100% of the time :/ Maybe 3 times of 5 I'd say.

I believe I'm using the default config:
[...]

PR #1841, opened 2017-05-18
Set line-height same as font size

mihaliak:
Set `x-row` `line-height` same as `font-size` to fix issues with some fonts on last line of terminal.

This PR should fix https://github.com/zeit/hyper/issues/628

UPDATE: removed whole css as chabou said line-height is same as font size as default

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -253,9 +253,6 @@ export default class Term extends Component {
       .cursor-node[focus="false"] {
          border-width: 1px !important;
       }
-      x-row {
-        line-height: ${this.props.fontSize}px;
-      }
       ${hyperCaret}
       ${scrollBarCss}
       ${selectCss}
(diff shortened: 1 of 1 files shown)


mihaliak:
@chabou just removed whole css block for x-row, this should be fine

ppot:
@chabou @rauchg This was added by @dotcypress for emoji!

mojoaxel:
@chabou Removing the whole line doesn't fix #628 for me.
Instead I use : 
```css
x-row {
    line-height: 1em;
}
```

chabou:
You're right, we need to specify `1em` to have emoji vertical aligned.
@mihaliak  can you update your PR with @mojoaxel recommendations ?

I'll merge it. If it leads to some emoji regression we'll find another way to correct this regression. #628 really hurts

albinekb:
I think this is 👌  we can fix regressions.

PR #1844, opened 2017-05-19
CMD+K: Clear Differently if Command Running

casesandberg:
Ready to be merged after tested by others! Fixes https://github.com/zeit/hyper/issues/1808. 

Fixed using @stephan281094's comment on https://github.com/zeit/hyper/pull/1811:

> I think the wipeContents function should only be called when there's no prompt active. Likewise, the onVTKeystroke('\f'); function should only be called when there is a prompt active.

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -177,7 +177,20 @@ export default class Term extends Component {
   }
 
   clear() {
-    this.term.onVTKeystroke('\f');
+    // Command Running
+    if (this.term.getCursorColumn() === 0) {
+      // Clear Screen and Scrollback Buffer
+      this.term.wipeContents();
+    } else {
+      // Clear Scrollback Buffer
+      this.term.scrollbackRows_.length = 0;
+      // Clear Screen but Leave Prompt
+      this.term.onVTKeystroke('\f');
+      // Reset Cursor Position
+      this.term.syncCursorPosition_();
+      // Fix Scoll Position Bug
+      this.term.scrollPort_.redraw_();
+    }
   }
 
   moveWordLeft() {


stephan281094:
Just tested this as well and it works great! Thanks @casesandberg.

casesandberg:
I don't have any sort of VM to test either 😞

albinekb:
Got a windows machine at home i can test with tomorrow

lib/components/term.js
```diff
@@ -177,7 +177,20 @@ export default class Term extends Component {
   }
 
   clear() {
-    this.term.onVTKeystroke('\f');
+    // Command Running
+    if (this.term.getCursorColumn() === 0) {
```
chabou: Seems really hacky.
I haven't test it but I think if your command is currently printing a progress bar, it could really break things.

casesandberg:
Ah, yes you are right thats a big issue. Do you know of any other ways to
detect if a command is running?

lib/components/term.js
casesandberg: Ah, yes you are right thats a big issue. Do you know of any other ways to
detect if a command is running?

albinekb: hmm yeah, I've thought about this, while this seems better than it was before, it's not perfect. I'll have a look 🔍

casesandberg:
Any thoughts on merging this but keeping the issue open? Unless anyone can come up with a better way of detecting if a command is running. This gets us most of the way there.

Stanzilla:
@casesandberg Hey there, sorry for the long silence! Did you test the case chabou had suspected will have issues? Are you willing to work on that?

casesandberg:
Just checked and this was fixed at some point and is working in 2.0.0

PR #1848, opened 2017-05-20
Upgrade electron to 1.6.8

albinekb:
Looked at https://github.com/zeit/hyper/issues/1738 etc, turns out the eval problem was easy to work around (it was because of `cheap-eval-source-map`), with this change source-maps still work 👌 🙌 

Latest electron uses node 7.4, so we still need   the `--harmony-async-await` (right? 🤔 ) for plugin compatibility.

Tested on macOS and seems to work as before.
need to check with windows...

This should give some nice memory optimizations etc. 💯


webpack.config.js
```diff
@@ -7,7 +7,7 @@ const nodeEnv = process.env.NODE_ENV || 'development';
 const isProd = nodeEnv === 'production';
 
 module.exports = {
-  devtool: isProd ? 'hidden-source-map' : 'cheap-eval-source-map',
+  devtool: isProd ? 'hidden-source-map' : 'inline-source-map',
```
chabou: Maybe better to use `cheap-module-source-map` regarding to [issue in webpack](https://github.com/webpack/webpack/issues/2145#issuecomment-294361203) that you linked in [this Electron issue](https://github.com/electron/electron/issues/8854)

albinekb: Will do

albinekb:
Done @chabou 🎉

chabou:
You're right! I can't reproduce #1738 anymore with this source-map format change.
It is not intellectually satisfying to accept this "regression" about `eval` but it is worth it! 👍

rauchg:
This is awesome work!

albinekb:
@chabou  🤦‍♂️  thanks for noticing 😂 
❤️  @rauchg

albinekb:
Let's merge when the tests are done 😎

Issue #1151, opened 2016-12-13
Windows Installer not Creating Shortcuts


matankb:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 10
- **Hyper.app version**: 1.0.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: gist.github.com/matankb/6da6a994c6a8886fab751440b0741ab7
- **Relevent information from devtools**: N/A (no errors/warnings)
- **The issue is reproducible in vanilla Hyper.app**: Is Vanilla

## Issue

The windows installer didn't create a start menu or desktop shortcut when ran. I also can't find anything in `C:/Program Files`, making it impossible to re-launch Hyper (asides from reinstalling). However, the uninstall part of the Control Panel does list Hyper as one of the installed apps. `hyper_plugins` and `.hyper.js` also exist.  Is this intended? If so, where are the `.exe`s kept?

Thank you!

stefanivic:
@matankb Hyper is installed in C:\\Users\\{userName}\\AppData\\Local\\hyper. Try to find it there.

matankb:
@stefanivic Found it, thank you!

ljfio:
Can I just say that this issue happens every time that I install Hyper on Windows from the installer?
To fix this I think that you need to use `%LocalAppData%\hyper\Hyper.exe` instead of using the `%UserProfile%` environment variable when creating the shortcuts

Stanzilla:
Hrm, the electron-builder version that Hyper uses is pretty old, `13.6.0` maybe try with the latest? `17.9.0`

PR #1851, opened 2017-05-21
Update electron-builder to 17.9.0

Stanzilla:
Maybe fixes https://github.com/zeit/hyper/issues/1151?

I had to change a few fields in `package.json` because the format changed in `electron-builder`, would be nice if someone on macOS could test if it still pulls the correct stuff from Info.plist, otherwise functionality is the same.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,16 +73,29 @@
   "build": {
     "appId": "co.zeit.hyper",
     "asar": true,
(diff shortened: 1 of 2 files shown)


Stanzilla:
rebased

rauchg:
cc @matheuss

PR #1853, opened 2017-05-21
Enable and use asar

Stanzilla:
Followup to https://github.com/zeit/hyper/pull/386, I guess.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
   },
   "build": {
     "appId": "co.zeit.hyper",
-    "asar": false,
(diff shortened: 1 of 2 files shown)


albinekb:
Can someone just test on windows before mergre? 🙏

Stanzilla:
I did it on Windows, not sure how to properly test though, "works" good enough? 😄

albinekb:
Yeah that's probably enough @Stanzilla, I believe it didn't work at all before but it was fixed in the asar packer.

PR #1867, opened 2017-05-25
Refactor menu internals

ppot:
## Keymaps sub implementation

Include menus refactor under `/menus`


app/index.js
```diff
@@ -390,13 +390,10 @@ app.on('ready', () => installDevExtensions(isDev).then(() => {
...
 
-  const setupMenu = () => {
-    const tpl = plugins.decorateMenu(createMenu({
-      createWindow,
-      updatePlugins: () => {
-        plugins.updatePlugins({force: true});
-      }
-    }));
+  const makeMenu = () => {
+    const menu = new AppMenu(createWindow, () => {
```
chabou: This menu should be decotated by plugins

app/config.js
```diff
@@ -17,12 +16,12 @@ const winCfg = new Config({
   }
 });
 
-let configDir = homedir();
-if (isDev) {
+let configDir = _paths.homeDir;
+if (_paths.isDev) {
```
chabou: Maybe it could be better to abstract this "config path if we are in dev mode" in `config/paths`. Strange to move `isDev` in it and not this dev path mechanism.
isDev is an environnement notion, not a path one. maybe better to separate it from `paths`

app/menus/menu.js
```diff
@@ -0,0 +1,76 @@
...
+    this.pluginsMenu = pluginsMenu(accelerators, updatePlugins);
+    this.windowMenu = windowMenu(accelerators);
+    this.helpMenu = helpMenu(os, app, shell);
+
+    const configFile = path.resolve(config.getConfigDir(), _paths.conf);
+
+    if (process.platform === 'darwin') {
+      this.darwinMenu = darwinMenu(accelerators, configFile);
+    } else {
+      this.editMenu.submenu.push(
```
chabou: It could be better to add this submenu directly in `editMenu`

app/menus/menu.js
```diff
@@ -0,0 +1,76 @@
...
+        {
+          label: 'Preferences...',
+          accelerator: accelerators.preferences,
+          click() {
+            shell.openItem(configFile);
+          }
+        }
+      );
+
+      this.helpMenu.submenu.push(
```
chabou: It could be better to add this submenu directly in `helpMenu`

app/menus/menu.js
```diff
@@ -0,0 +1,76 @@
...
+// menus
+const viewMenu = require('./menus/view');
+const shellMenu = require('./menus/shell');
+const editMenu = require('./menus/edit');
+const pluginsMenu = require('./menus/plugins');
+const windowMenu = require('./menus/window');
+const helpMenu = require('./menus/help');
+const darwinMenu = require('./menus/darwin');
+
+module.exports = class Menu {
```
chabou: I don't understand why you made a class here. Do you plan to extend it?

app/menus/menu.js
```diff
@@ -0,0 +1,76 @@
...
+  constructor(createWindow, updatePlugins) {
+    this.viewMenu = viewMenu(accelerators);
+    this.shellMenu = shellMenu(process.platform === 'darwin', accelerators, createWindow);
+    this.editMenu = editMenu(accelerators);
+    this.pluginsMenu = pluginsMenu(accelerators, updatePlugins);
+    this.windowMenu = windowMenu(accelerators);
+    this.helpMenu = helpMenu(os, app, shell);
+
+    const configFile = path.resolve(config.getConfigDir(), _paths.conf);
```
chabou: Maybe `config/paths` should know this absolutePath

app/index.js
chabou: maybe when you call `make()`

app/config.js
ppot: @chabou The DEV mechanism is part of the Global Keymaps. But for this I need to rework the whole config which is another part. See #1509 with `config/paths.js`

app/menus/menu.js
ppot: As my previous comment : see previous comment

app/menus/menu.js
```diff
@@ -0,0 +1,39 @@
...
+  this.pluginsMenu = pluginsMenu(accelerators, updatePlugins);
+  this.windowMenu = windowMenu(accelerators);
+  this.helpMenu = helpMenu(_paths.icon);
+
+  if (process.platform === 'darwin') {
+    this.darwinMenu = darwinMenu(accelerators, _paths.confPath);
+  }
+
+  const menu = [].concat(
+    this.shellMenu,
```
chabou: We should avoid using `this`. I think it is ok to concat directly like:
```javascript
const menu = [].concat(
  shellMenu(accelerators, createWindow),
  editMenu(accelerators, _paths.confPath),
  ...
)
```

app/menus/menu.js
```diff
@@ -0,0 +1,39 @@
+const _paths = require('../config/paths');
+const {accelerators} = require('../accelerators');
```
chabou: Maybe submenus could require paths and accelerators directly when needed. I don't know. What is the advantage to do this here ?

app/menus/menu.js
ppot: Yeah. I kept this since it was in a `class` before.

app/menus/menu.js
ppot: Using this will only be a temp fix waiting keymaps commands.

Issue #1854, opened 2017-05-21
Windows Script Host


balmungsama:
I go to Edit > Preferences, and I get a Windows Script Host error

Script: C:\Users\john\.hyper.js
Line: 5
Char: 1
Error: "module" is undefined
Code: 800A1391
Source: Microsoft JScript runtime error

-

Hyper 1.3.3
Electron 1.4.16
win32 x64 10.0.15063

albinekb:
Duplicate of https://github.com/zeit/hyper/issues/1812 thanks for reporting

Issue #1810, opened 2017-05-10
Hyper config 'module' is undefined


Meadowcottage:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 10-64 Bit (Creators Update)
- **Hyper.app version**: 1.3.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: [.hyper.js](https://gist.github.com/Meadowcottage/558b4e184ee98e0fee223b7678abcba0)
- **The issue is reproducible in vanilla Hyper.app**: Yes. Issue only current in version 1.3.2+.

## Issue
Control+Comma (Shortcut to open .hyper.js file) in version 1.3.2 results in the following error.
![](https://i.imgur.com/PfI0qOa.png)
Tested in prior version, 1.3.1, and the issue isn't current. Guessing an issues from the 1.3.2 update.

DavidDeSloovere:
I got the exact same error. Fresh installation.

Hyper 1.3.2
Electron 1.4.16
win32 x64 10.0.16184

Pyrepenol:
Same here. win 10 64-bit, Hyper 1.3.2

andreynazarov3:
same here. win 10 64-bit

denniseffing:
Same here, Win 10 1703 64-Bit, Hyper 1.3.3

Szomy:
Try to set a default editor for .js files.
Windows tries to run javascript as a script.

albinekb:
Solved in #1872

Bandit:
I'm still getting this problem after downloading the .exe from the website. Has the installer not been updated perhaps?

ppot:
Yes, it will be in the next release! Not released yet

Positive07:
Note that it will open the file with Notepad (not your default editor, whatever it may be)

AmirinNS:
I just change the default application for .js file. For example, I am using Sublime for editing .js file instead of using Microsoft Script. It should work by this way 
![snap35](https://user-images.githubusercontent.com/23518901/28743783-e38efd9e-7484-11e7-89de-b9da273a093e.PNG)

isseiler:
This is still an issue:

[...]

PR #1872, opened 2017-05-26
🎉 Check for WScript.exe

### LINKED ISSUE #1812: Error on trying to edit preferences
When I go to Menú --> Edit --> Preferences, I'm getting the following message:

Error: 'module' is not defined.

Please see the attached image for more info.
Thanks.
image

-

Hyper 1.3.2
Electron 1.4.16
win32 x64 6.1.7601

albinekb:
This is my proposed solution for handling #1812

We could add a key in the config for setting the editor, but I think that's a bad idea. Better to teach the users to set the correct app for .js files :) maybe put an alert/dialog with a link on how to do it.

At least this will let people open their config without errors, and we can think about how we can make this better in the future

Fixes #1812 https://github.com/zeit/hyper/issues/1810 https://github.com/zeit/hyper/pull/1784#issuecomment-301396453 https://github.com/zeit/hyper/issues/1854


app/menus/menus/edit.js
```diff
@@ -52,7 +51,7 @@ module.exports = function () {
         label: 'Preferences...',
         accelerator: accelerators.preferences,
         click() {
-          shell.openItem(confPath);
+          openConfig();
```
chabou: Can you use the same `onpenConfig()` function here: https://github.com/zeit/hyper/blob/3c1f359198a71c6946f326bfda3e53f612ade23e/app/menus/menus/darwin.js#L21 ?

chabou:
I really love this solution! ❤️

app/menus/menus/edit.js
albinekb: yep fixed @chabou 👌

app/config/open.js
```diff
@@ -0,0 +1,33 @@
+const {shell} = require('electron');
+const {confPath} = require('./paths');
+
+module.exports = () => Promise.resolve(shell.openItem(confPath));
```
tnguyen14: Why are there 2 `module.exports` in this file?

albinekb: since we want to have a different module on windows compared to others

app/config/open.js
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  // Windows opens .js files with  WScript.exe by default
+  // If the user hasn't set up an editor for .js files, we fallback to notepad.
+  const canOpenNative = () => new Promise((resolve, reject) => {
+    exec('ftype JSFile', (error, stdout) => {
+      if (error) {
+        reject(error);
+      } else if (stdout && stdout.includes('WScript.exe')) {
+        reject(new Error('WScript is the default editor for .js files'));
```
tnguyen14: If `WScript.exe` exists, this will throw an error, which will then open the conf file with notepad? I thought the idea is to use `WScript.exe` if it exists?

ppot: @tnguyen14 No, `WScript.exe` run the file instead of editing it.

albinekb: It will check if your default editor is set to WScript.exe, if it is , it will use notepad.exe, since WScript.exe isn't an editor

tnguyen14: Ah. I see. I understand now. Thank you for the clarification. I was confused by this comment: `Windows opens .js files with  WScript.exe by default`

Stanzilla:
Does not work on Windows, even though a default program other than WScript is set for .js, it opens notepad.

dave-irvine:
I've got a bit of a problem here in that you can manually change filetype associations like this:

    ftype JSFile=notepad.exe %1 %*

but then hyper won't start any more. So without a configuration option to change what editor opens the configuration file, I'm stuck with notepad.

Any suggestions?

Issue #1506, opened 2017-02-14
A space before the cursor


tideman:
<img width="267" alt="screenshot" src="https://cloud.githubusercontent.com/assets/2640038/22918541/11bcbe80-f28b-11e6-82b5-91ff0cb8d02d.png">
 I am on the latest Hyper.app version
 I have searched the issues of this repo and believe that this is not a duplicate

OS version and name: macOS Sierra 10.12.2
Hyper.app version: 1.1.0

When my command prompt PS1 is loaded in Hyper it places one space before the cursor which leaves me confused when editing texts.

I tried editing the functions I use in my .bashrc to see if the space came from there but I didn't find it. 

In my regular bashshell the extra space is not there.

This is the code I use to generate my PS1

`
parse_git_branch() {
   git branch 2> /dev/null | sed -e '/^[^*]/d' -e 's/* \(.*\)/ (\1)/'
}

parse_latest_commit() {
   git log -n 1 --pretty=format:"%H"
}

ME_EMOJI=(🐶 🐱 🐭 🐹 🐰 🐸 🐯 🐨 🐻 🐷 🐮 🐵 🐼 🐙 🐧🐦🐤🍒🍈🍓😬😄😃😀☺️);
RANDOM_ME_EMOJI=${ME_EMOJI[$((RANDOM%14))]};

export PS1="$RANDOM_ME_EMOJI  \u \W \[\033[32m\]\$(parse_git_branch)\[\033[00m\] => "
`

markkra:
I thought it was maybe BLOCK vs BEAM or one of the plugins. But no luck fixing it. Seems like an "off by one" error someplace.

andrei-zhidkov:
I am experiencing the same issue when using tmux. Without tmux looks fine.
macOS Sierra 10.12.3
zsh 5.3.1 + tmux 2.3
Hyper 1.1.0

mattdarveniza:
Have you tried removing the emoji from your prompt? I've noticed this happens only with certain emoji. Might be something to do with how many unicode characters a certain emoji actually are? i.e. some only take one, others take two, which might be confusing hyper about how much space it should occupy.

seagoj:
This actually happens to me as well, but only in tmux. Outside of a tmux session, everything aligns correctly.

flybayer:
Broken here too on OSX, but only in tmux.

jakl:
Broken for me too.
[...]

PR #1873, opened 2017-05-26
Fix unicode display, third attempt :\

tibotiber:
This is a follow up PR to #1536. It reverts the revert of the merge, and fixes the "includes of undefined" issue that appeared after running `dist`. Since I cannot reproduce the issue locally, it is a blind fix, please check on your side before merging ;).
Cc @albinekb @rauchg

### DIFF
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -378,7 +378,7 @@ hterm.Screen.prototype.insertString = function (str) {
     }
     reverseOffset = 0;
   }
-  this.wrapUnicode(cursorNode);
+  cursorNode = this.wrapUnicode(cursorNode);
   if (this.textAttributes.matchesContainer(cursorNode)) {
     if (reverseOffset === 0) {
       cursorNode.textContent = cursorNodeText + str;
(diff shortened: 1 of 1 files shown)


tibotiber:
Added a quick fix for #1506.

Before:
before

After:
after

albinekb:
@tibotiber
>Uncaught DOMException: Failed to execute 'appendChild' on 'Node': This node type does not support this method.

tibotiber:
@albinekb Can you try again? Sorry, it's kinda unpractical as I'm not able to test myself.

I changed `wrapUnicode` to replace text nodes by span elements if it needs to append children.

albinekb:
Will do, but I wonder why this isn't happening to you 😱

albinekb:
Inserting an emoji: image

but when i press space after the emoji:
image

🤔

tibotiber:
@albinekb Thanks for your help with testing. I finally found how to reproduce the error. I had to remove `hyperlinks` from my plugins 😓. Tried your locales and envs settings but that didn't change anything. Glad I manage to reproduce anyway, it makes things much easier for debugging 😄. 

I believe I fixed the issue now. Since the previous commit, `wrapUnicode` may reassign the node it was given and return it, so the calling site needs to update the node with the returned value. This was done properly but I missed that the node was also needed to be reassigned to `this.cursorNode_`. Hence the `cursorNode_` was left out of updates. It is fixed now.

Hope we're good soon (c'mon, I can do this 😛).

albinekb:
It's getting there! kapture 2017-05-28 at 23 06 13

Now i can type emojis, but seems like the width is not correct? When i type something after an emoji (another emoji, space, character etc) it breaks 🤔

schmittyjd:
Will this pr also fix issues rendering powerline fonts?

tibotiber:
@schmittyjd yes powerline fonts rendering will be fixed

@albinekb 
- the width of emojis is correct, it is in line with how iTerm and Terminal handle it, see [this comment](https://github.com/zeit/hyper/pull/1536#issuecomment-280816134) from @rauchg 
- for the emoji inputs, there are still some issues yes. I ignored them since it is also present in master. See examples below:

zsh before:
zsh-before

zsh after:
zsh-after

bash before:
bash-before

bash after:
bash-after

What do you think?

tibotiber:
@albinekb friendly ping: any opinion about the issues for unicode input? Seems kind of independent from this PR to me, although the behaviour changes from bad to bad 🤔. Could it be addressed in a subsequent PR?

PS: VS Code moving forward with this too apparently: https://twitter.com/tyriar/status/870346230911483908 and they are on xterm

tibotiber:
ping: what is the status with this? are we proceeding or dropping it for the upcoming switch to xterm?

AndrewSB:
+1

would love an update on this too!

LitoMore:
xterm +1

Stanzilla:
@tibotiber Wanna close this? Or is anything in this PR still needed on `canary` after the switch to xterm?

albinekb:
@tibotiber thanks for attempting to fix this 🙏 As we have moved away from the problematic `hterm` lib, I'm closing this PR here. Again, thanks for your efforts and hope you (I did!) learned something by doing this.

chabou:
@tibotiber Really sorry and thank you so much for your help 🙏

tibotiber:
@albinekb @chabou thanks for the kind message and no worries, what's important is that we've got hyper in a better shape now. At the very least i've learnt why we're moving to `xterm`, `hterm` was indeed problematic 😅. The work on these few PRs has been interesting anyway :). Cheers and keep the great work on hyper and more.

Issue #1121, opened 2016-12-12
Command ctrl + C does not work


fabdelgado:
The ctrl + c command to interrupt operations does not work.

markpanado:
I tested mine with the new update and it's working.

mqudsi:
It doesn't work for me, either. Ctrl+c is completely ignored running on Windows 10 build 1607 running cmd.exe (not bash.exe). In bash.exe, ctrl+c works, but the up arrow key does not work.

willharrison:
ctrl + c works for me.  Having issues with ctrl+ r, up arrow, down arrow, and ctrl + backspace.

warpdesign:
Same problem here.

Note that the moment I press 'shift' or 'ctrl', the  text I have selected gets unselected, so keeping shift+ctrl down and pressing c to paste does nothing as there's nothing left selected.

shern91:
i also facing the save problem in copy text to clipboard, no matter i click Ctrl+C or Shift+Ctrl+C, once i click any of key those text will deselected. 

P/S: Windows 10 64 bit user.

stereokai:
All the issues described so far are happening on my machine as well.

mqudsi:
I just want to point out that ctrl+c works in a bash.exe shell, even though it does not work in the standard cmd.exe shell.

stereokai:
@mqudsi just to remove any shadow of doubt, it doesn't perform a copy, but cancels the current command (as it should)

ppot:
@warpdesign Fix in #1218

kvj:
The problem isn't only Ctrl+C alone, but any hotkeys bound to menu items under Linux and Windows. They're intercepted by Electron and not delivered to main window, I guess. For example, all my keybindings for tmux start from Ctrl+A and don't work in Hyper under Linux (instead, Select All is executed), but work well in macOS (Select All is Cmd+A there)

liuyang1204:
@kvj +1

mqudsi:
@liuyang1204 there's a +1 button.

@kvj I'm not so sure about that. If that were the case, ctrl+c would not suddenly function if a bash.exe session is started under Windows.

annnoo:
Having quite the same problem in #1199 . I bound the "Close Session" shortcut to another key, but Ctrl+W is not working at all in vim.

jamiewade:
Pressing `Ctrl + C` closes the window for me, but I want to use this keyboard command to stop processes running, for example Gulp.

[...]

PR #1876, opened 2017-05-26
Implements Commands Key mapping

ppot:
-> keymaps files
-> config structure change (file deconstruction)
-> plugins extends keymaps
-> update hterm to use `commands-registry`


app/config.js
```diff
@@ -1,82 +1,31 @@
...
-  }
-});
-
-const path = _paths.confPath;
-const pathLegacy = _paths.pathLegacy;
+const win = require('./config/windows');
 
 const watchers = [];
-
+const scanInterval = 2000;
```
albinekb: I would:
```js
const getWatchConfig = () => {
  if (process.platform === 'win32') {
    // watch for changes on config every 2s on windows
    // https://github.com/zeit/hyper/pull/1772
    return { interval: 2000 }
  }

  return {}
}

const watchConfig = getWatchConfig()

...

gaze(confPath, watchConfig, function (err) {
```

easier to read?

albinekb:
One thing that would be 💯^10 is to be consistent with variable names, I'm thinking of the words
`cfg`, `conf` and `config` that are mixed, it's easier to read and understand the code if you use the same name for these, chose one and stick to it 👌

app/config.js
ppot: Yeah  agree on that!

app/config.js
```diff
@@ -1,82 +1,31 @@
...
       throw err;
     }
     this.on('changed', () => {
-      try {
-        if (exec(readFileSync(path, 'utf8'))) {
-          notify('Hyper configuration reloaded!');
-          watchers.forEach(fn => fn());
-        }
-      } catch (err) {
-        dialog.showMessageBox({
```
albinekb: This PR changes from displaying a dialog to just a notification, correct?

app/config.js
```diff
@@ -135,14 +54,20 @@ exports.getPlugins = function () {
...
-  recordState(win) {
-    winCfg.set('windowPosition', win.getPosition());
-    winCfg.set('windowSize', win.getSize());
+exports.getKeymaps = function () {
+  return cfg.keymaps;
+};
+
+exports.extendKeymaps = function (keymaps) {
+  if (keymaps) {
+    cfg.keymaps = keymaps;
```
albinekb: Hmm.. I don't like mutating stuff like this. But I think we do this in other places? Maybe that's a job for another PR.

app/config/import.js
```diff
@@ -0,0 +1,41 @@
+const {writeFileSync, readFileSync} = require('fs');
+const {defaultConfig, confPath} = require('./paths');
+const _init = require('./init');
+const _keys = require('./keymaps');
```
albinekb: nit: rename this to `_keymaps`, easier to understand the code

app/config.js
ppot: yeah we do this alot...

app/config/init.js
```diff
@@ -0,0 +1,43 @@
...
+const _extract = function (script) {
+  const module = {};
+  script.runInNewContext({module});
+  if (!module.exports) {
+    throw new Error('Error reading configuration: `module.exports` not set');
+  }
+  return module.exports;
+};
+
+const _syntaxValidation = function (cfg) {
```
albinekb:
```js
const _syntaxValidation = function (cfg) {
    try {
      return new vm.Script(cfg, { filename: '.hyper.js', displayErrors: true, lineOffset: 0 })
    } catch (error) {
      notify(`Error loading config: ${error.name}, see DevTools for more info`)
      console.error('Error loading config:', error)
      return
    }
}
```
tested:
```js
_syntaxValidation(`
const c = 'a'
const b = 'd'
const c = '2'
`)
```

notify: *Error loading config: SyntaxError, see DevTools for more info*
console:
```js
 Error loading config:
 .hyper.js:4
const c = '2'
          ^^^
SyntaxError: Identifier 'c' has already been declared

File: app/config.js (hunk 3 above)
Comment (albinekb) [inline app/config.js]: I'm OK with this for now, but we should clean it all up

File: app/config/init.js (hunk 5 above)
Comment (ppot) [inline app/config/init.js]: Are you commenting as an error or a test?

File: app/config/paths.js [hunk 6]
```diff
@@ -15,10 +14,14 @@ let confDir = homeDir;
 const devDir = resolve(__dirname, '../..');
 const devConfig = resolve(devDir, conf);
 const defaultConfig = resolve(__dirname, defaultConf);
-const pathLegacy = resolve(homeDir, legacyConf);
 
 const icon = resolve(__dirname, '../static/icon.png');
 
+const keymapPath = resolve(__dirname, '../keymaps');
+const darwinKeys = resolve(keymapPath, 'darwin.json');
```
albinekb: Should't these be `path.join`?
http://tips.tutorialhorizon.com/2017/05/01/path-join-vs-path-resolve-in-node-js/

don't know how resolve vs join is performance/resource-wise, 
it's probably so tiny that it's non-existent 😄  

🤔

app/config/keymaps.js
```diff
@@ -0,0 +1,57 @@
...
+const _setCommandsForKeys = function (commands) {
+  for (const command in commands) {
+    if (command) {
+      keys[commands[command]] = command;
+    }
+  }
+};
+
+const _import = function (customsKeys) {
+  const path = () => {
```
albinekb: Why not move this to `app/config/paths.js` and just expose it as `defaultPlatformKeyPath`?

Is there a need to get other platforms' keys?

app/keymaps/darwin.json
```diff
@@ -0,0 +1,34 @@
+{
```
albinekb: run these trough http://jsonprettyprint.com or similar 😄

app/menus/menu.js
```diff
@@ -8,18 +10,19 @@ const helpMenu = require('./menus/help');
...
 module.exports = (createWindow, updatePlugins) => {
+  const commands = getKeymaps().commands;
   const menu = [].concat(
-    shellMenu(createWindow),
-    editMenu(),
-    viewMenu(),
-    pluginsMenu(updatePlugins),
-    windowMenu(),
-    helpMenu()
+    shellMenu(commands, createWindow),
```
albinekb: ES6 style:
```js
const menu = [
  ...shellMenu(commands, createWindow),
  ...editMenu(commands),
  ...viewMenu(commands),
  ...pluginsMenu(commands, updatePlugins),
  ...windowMenu(commands),
  ...helpMenu(commands)
]

Comment (albinekb) [inline app/menus/menu.js]: then remove this further down:
```
if (process.platform === 'darwin') {
  menu.unshift()...
```

and:
```js
const menu = [
  ...(process.platform === 'darwin' ? darwinMenu(commands) : []),
  ...shellMenu(commands, createWindow),
  ...editMenu(commands),
  ...viewMenu(commands),
  ...pluginsMenu(commands, updatePlugins),
  ...windowMenu(commands),
  ...helpMenu(commands)
]
```
💅

app/menus/menus/shell.js
```diff
@@ -29,7 +27,7 @@ module.exports = function (createWindow) {
       },
       {
         label: 'Split Vertically',
-        accelerator: accelerators.splitVertically,
+        accelerator: commands['pane:vertical'],
```
albinekb: maybe `pane:splitVertical` to make it clearer, same with horizontal

app/menus/menus/window.js
```diff
@@ -1,15 +1,14 @@
...
     role: 'window',
     submenu: [
       {
         role: 'minimize',
-        accelerator: accelerators.minimize
+        accelerator: commands['window:minimize']
       },
       {
-        role: 'zoom'
+        role: 'zoom',
```
albinekb: What does this actually do? 😂 never used it

app/menus/menus/window.js
```diff
@@ -71,7 +70,7 @@ module.exports = function () {
       },
       {
         role: 'togglefullscreen',
-        accelerators: accelerators.enterFullScreen
+        accelerators: commands['window:full']
```
albinekb: `window:toggleFullScreen`, easier to understand

app/config/init.js
albinekb: haha the comment was to update _syntaxValidation to the code i wrote in my comment,
the blocks under is to show what it does

app/config/paths.js
MartyGentillon: given that they contain `..` resolve is probably better.

app/menus/menus/window.js
MartyGentillon: Apparently nothing on Windows?

app/config/paths.js
albinekb: they do not @MartyGentillon `resolve(keymapPath, 'darwin.json')` `keymapPath` is already expanded

app/menus/menus/window.js
albinekb: Aha, makes the window fill the screen (on macOS) 😄

app/config/paths.js
MartyGentillon: True enough for this particular path, but that would be the reasoning nonetheless.

app/config/init.js
ppot: Done 👍

app/keymaps/darwin.json
ppot: Done 👍

app/config.js
ppot: I won't change it since using a dialog prevent hyper from opening.

app/menus/menus/window.js
MartyGentillon: Given that it does nothing on Windows (and KDE), should it even be in the menus on Windows (or Linux)?

MartyGentillon: Alternately, that could be a bug, It looks like that is supposed to be the same as "Maximize" in windows and linux, which don't appear anywhere in the menus on either (though they do appear in the window controls).

ppot:
@albinekb Next review?

TalAmuyal:
Will there be an option of completely disabling keybindings?

ppot:
@TalAmuyal What do you mean exactly? Removing all the key from the menu?

TalAmuyal:
I mean that as a user of `tmux` and `(neo)vim`, I have all the "on-demand" functionality in them (and the shell).
For example, some of my non-terminal keybindings:
- `<Ctrl>+T` - Open tab (aka window) -` tmux`
- `<Alt>+<Arrow>` - Move between tabs - `tmux`
- `<Ctrl>+R` - History search - `zsh`
- `<Ctrl>+W, H/J/K/L` - Switch pane - `Vim`

In short, every keybinding that the terminal provides might override an existing config.
So for me, the ultimate terminal has no keybindings, as everything is  pre-configured in a file.

ppot:
@TalAmuyal  If you are using MacOS they are prefixed with `cmd`. You can always override them in the `.hyper.js` at the `keymaps` object.
```
  keymaps: {
    // exemple
    'window:devtools': 'ctrl+shift+o'
  }
```

sandric:
I long waited for feature from early when I saw project, and I giving huge thx to @ppot for implementing it, so that I finally can try one more time to tackle my long lasting dream to use emacs from inside terminal. Having said that, is there any docs about how to make custom keybinding? I tried to tweak basic darwin kbs and they seems to work, but is there a way to create your own functions, not only predefined, like "editor:selectAll"?

ppot:
Thanks @sandric  for now this could be possible but by creating plugins with commands execution function. This will be made more easily once we switch to xterm!

Algram:
Does that mean we can use ctrl-r for reverse-i-search on linux now? If yes, how?

ppot:
@Algram On current master yes.

saadq:
This is awesome. Is there going to be a way to see a list of available commands, or maybe have a way to allow commands to be logged to the console when you do some action (similar to Sublime Text's `sublime.log_commands(True)`?

ppot:
mmm Make an issue for this as a suggestion ;) something like a command invocation actions

stereokai:
@ppot Does this fix [Command ctrl + C does not work #1121](https://github.com/zeit/hyper/issues/1121)?

computersarecool:
So,
Given the status of this pull request, how would one enable Readlines (emacs) style keybindings in Hyperterm?

sindrenm:
Hype! Can confirm that this is indeed working! :tada: :clap: 

Any idea when it'll get released? :grinning:

computersarecool:
To me it is unclear from the title "Implements Commands Key mapping" what this pull request will do but it closes a lot of open issues.

@ppot will this pull request make it so that keybindings in the Hyper terminal emulator will work as they should?

i.e. will [these](http://www.keyxl.com/aaaf192/83/Linux-Bash-Shell-keyboard-shortcuts.htm) work?

ELLIOTTCABLE:
Ditto @computersarecool, especially given that this is linked from the v1.4.0 release-notes — what, exactly, just got added? ^_^

ynnoj:
@computersarecool @ELLIOTTCABLE https://hyper.is/#keymaps

iamstarkov:
for @ELLIOTTCABLE and everyone else looking for available shortcuts

look 'em up here https://github.com/zeit/hyper/blob/master/app/keymaps/

PR #1878, opened 2017-05-27
Proxy ReactDOM for plugins

albinekb:
We do this for React, etc. But some react components require `react-dom`, to not have duplicate versions of ReactDOM, we proxy the one already in use (as we do for React)


albinekb:
thanks 💯

PR #1892, opened 2017-06-01
Add broadcast options

tzilist:
This PR allows user to broadcast commands across multiple terminal sessions. The options allowed are:

1. Broadcast across all sessions (including ones in a different tab)
2. Broadcast across all sessions in the current tab
3. Broadcast to only the current tab

This is very much a work in progress and might need some adjustments! I have currently tested it on a mac but need to verify that the key commands work on window/linux so if anyone could help me with that, that would be great!


albinekb:
https://www.npmjs.com/package/hyper-broadcast

tzilist:
Oh I wish I had seen that earlier. 😂 Gonna close this for now!

albinekb:
Thanks anyways! And I'm sure you learnt something from this 😉👍🏻 @tzilist

PR #1896, opened 2017-06-02
Fixed a bug with installing plugins

tschoffelen:
Due to the new `package-lock.json` file generated by NPM 5, new plugins will not be installed automatically anymore. This commit adds `--no-shrinkwrap` to the `npm install` command to make sure the contents of `package-lock.json` are ignored, causing all dependencies listed in `package.json` to be installed correctly.

The parameter `--no-shrinkwrap` is also available in earlier versions of NPM, so no backwards compatibility issues should arise from this change.

### DIFF
--- a/app/plugins.js
+++ b/app/plugins.js
@@ -237,8 +237,8 @@ function install(fn) {
     /* eslint-enable camelcase  */
     // Shell-specific installation commands
     const installCommands = {
-      fish: 'npm prune; and npm install --production',
-      posix: 'npm prune && npm install --production'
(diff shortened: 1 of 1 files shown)


Stanzilla:
Oh so that is why it fails, nice find!

joggienl:
Hi!
Is there a release planned with this fix in? Would be very nice to have it available soon!

Thanks!

Yhubee:
Sorry for the really noob question but where do I put this in? I'm using the Windows Installer.

PR #1897, opened 2017-06-02
Embedded npm

chabou:
With this PR, user won't need to have npm installed to have plugins.
Hyper will be protected against (npm@5) incompatibility too.

It works great but some tasks remain to be done :
* [x] Find a way to make a `npm prune`
* [x] Benchmark `Hyper`size and speed
* [ ] Test on Windows & Linux

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -22,6 +22,7 @@
     "git-describe": "3.0.2",
     "mkdirp": "0.5.1",
     "ms": "0.7.1",
+    "npm": "4.6.1",
     "node-fetch": "1.6.3",
     "node-pty": "0.6.6",
     "semver": "5.3.0",
--- a/app/plugins.js
+++ b/app/plugins.js
@@ -78,17 +78,10 @@ function updatePlugins({force = false} = {}) {
 
     if (err) {
       console.error(err.stack);
-      if (/not a recognized/.test(err.message) || /command not found/.test(err.message)) {
-        notify(
-          'Error updating plugins.',
-          'We could not find the `npm` command. Make sure it\'s in $PATH'
-        );
-      } else {
-        notify(
-          'Error updating plugins.',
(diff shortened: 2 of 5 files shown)


chabou:
Using `npm prune` programmatically is not trivial because there is no option to make it in another directory than cwd (like `prefix` option for `npm install` command).

chabou:
Maybe #381 duplicate.
Not the same approach.

rauchg:
I really want this for the next release! Great work @chabou

chabou:
Little benchmark. 

* Speed
Adding/removing 5 plugins when Hyper is started `['hyper-font-smoothing', 'hyper-simple-vibrancy', 'hyper-pane', 'hyper-broadcast', 'hyper-autoprofile']`
  * Master: adding 1693ms, removing 1362ms
  * This PR: adding 679ms, removing 473ms
* Exe size (on Windows)
  * Master: 64 396 288
  * This PR: 67 333 632

Speed improvements was not an objectif but it is pretty cool.
Size increase worth the system consistency

NPM v4.6.1 because v5.0.3 API is buggy. Callbacks are too many called.

albinekb:
https://github.com/zeit/hyper/pull/172#issuecomment-239582144

chabou:
Yes, using internal (public but non documented) API is bad. And I really understand @iarna disclaimer. 
But it will not be a problem to adapt our code to future changes. This PR is using a very little part of this API. It is simple and it works great.

However, I am totally ok if we prefer #381 approach.

albinekb:
Yeah I think this is OK now, but we'll need to change it sooner or later, so better to go with the "final" solution right away (run it as cli)

iarna:
You can get machine readable output from calling `npm prune --json` and `npm install --json`, albeit this is much improved with npm@5.

chabou:
Closed in favor of #381 which is ready to be merged.

PR #1900, opened 2017-06-02
🌹 Update travis node

albinekb:
Since electron 1.5 we're running node 7.4, so better to run the tests at that version? 🤔 

As soon as electron 1.7 is released it will be 7.9
https://github.com/electron/electron/pull/9116

Don't know if travis has 7.4, so let's see 😄


ppot:
Passed. So let's merge.

PR #1901, opened 2017-06-02
Make npm run clean platform agnostic

Stanzilla:
meh-ish solution since it installs rimraf globally, open to other suggestions

`npm cache clear` was removed because it is no longer needed with npm@5


package.json
```diff
@@ -10,10 +10,10 @@
     "test:unit": "ava test/unit",
     "test:unit:watch": "npm run test:unit -- --watch",
     "prepush": "npm test",
-    "postinstall": "install-app-deps && npm run rebuild-node-pty",
+    "postinstall": "install-app-deps && npm run rebuild-node-pty && npm install -g rimraf",
```
chabou: you don't need to install it globally. Add it as a dev dependency

Stanzilla: @chabou that won't work since it will try to delete itself in that step

Stanzilla:
The other idea would be to just make a `wclean` task that uses Windows tools.

package.json
albinekb: we can't merge this if it installs it globally 🌍

chabou:
Or maybe use: https://www.npmjs.com/package/rimraf-standalone

Stanzilla:
That would mean having rimraf-standlone.js in the repo, would you guys be fine with that?

package.json
Stanzilla: I know, that's why I have been asking for other options :)

chabou:
I don't know if I'm fine with that but it should be in a `./bin/` directory in any case 😛

Stanzilla:
Thoughts?

PR #1902, opened 2017-06-02
react-addons-pure-render-mixin is deprecated

Stanzilla:
... so use React.PureComponent instead

* https://www.npmjs.com/package/react-addons-pure-render-mixin
* https://facebook.github.io/react/docs/react-api.html#react.purecomponent

Let's try this again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,6 @@
     "mousetrap": "1.6.1",
     "ms": "0.7.2",
     "php-escape-shell": "1.0.0",
-    "react-addons-pure-render-mixin": "15.3.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -121,7 +121,6 @@
     "mousetrap": "1.6.1",
     "ms": "0.7.2",
     "php-escape-shell": "1.0.0",
-    "react-addons-pure-render-mixin": "15.3.2",
```
chabou: You need to commit your `yarn.lock` too

Stanzilla: Oh oops, done!

PR #1918, opened 2017-06-09
Check default app

albinekb:


albinekb:
Fix for https://github.com/zeit/hyper/pull/1872#issuecomment-306976819
ftype doesn't work as documented 😄

Stanzilla:
Tested and works well, 👍

Issue #1862, opened 2017-05-23
`undefined` in `plugins` array causes Hyper to refuse to open windows, with no visible errors whatsoever


ELLIOTTCABLE:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: macOS Sierra **10.12.5 Beta** (16F60a)
- **Hyper.app version**: Hyper ** 1.3.3.1754**
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/ELLIOTTCABLE/678ad79704a104c8f29efcfaddd09e07
- **The issue is reproducible in vanilla Hyper.app**: **Yes**

## Issue
Having accidentally commented out the first element of a comma-prefixed `plugins` array in `.hyper.js`,

```js
   plugins: [
  //  'hypercwd'
    , 'hyperterm-alternatescroll'
    , 'hyperterm-focus-reporting'
      // ...
   ],
```

… I discovered that Hyper shows no errors, but refuses to open any windows.

Probably not a common mistake, or a huge problem, but I thought it'd be worth reporting. 😜

ppot:
#1509

nitsakh:
Hello, I would like to take up this issue. I haven't contributed to Hyper before. Would the #hyper channel on slack be a good place to get help?

chabou:
Do not hesitate to ask on #hyper channel.
Yep, we must ignore `undefined` plugins. Maybe remove `undefined`entries in plugin array after config parsing.

albinekb:
😂 comma prefixed, nice 👌 

Yeah let's just do `.filter(Boolean)`

albinekb:
Fixed in #1927

ELLIOTTCABLE:
<'3

Thanks, @nitsakh! And welcome to open-source!

PR #1927, opened 2017-06-12
Addressing issue #1862 by filtering out Boolean values from the plugins list

nitsakh:
This adds the `.filter(Boolean)` to filter out the undefined values in the config plugins array.

### DIFF
--- a/app/config/init.js
+++ b/app/config/init.js
@@ -36,6 +36,8 @@ const _init = function (cfg) {
       notify('Error reading configuration: `config` key is missing');
       return _extractDefault(cfg.defaultCfg);
     }
+    // Ignore undefined values in plugin array Issue #1862
+    _cfg.plugins = _cfg.plugins.filter(Boolean);
     return _cfg;
   }
   return _extractDefault(cfg.defaultCfg);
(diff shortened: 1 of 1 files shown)


app/config/init.js
```diff
@@ -36,6 +36,8 @@ const _init = function (cfg) {
       notify('Error reading configuration: `config` key is missing');
       return _extractDefault(cfg.defaultCfg);
     }
+    // Ignore undefined values in plugin array Issue #1862
+    _cfg.plugins = _cfg.plugins.filter(Boolean);
```
chabou: Could `_cfg.plugins` be `undefined`? If yes we should check it:
```javascript
_cfg.plugins = (_cfg.plugins && _cfg.plugins.filter(Boolean)) || [];
```
`localPlugins` should be filtered too

nitsakh: Yes, that's right! Will add that. About `localPlugins` I was thinking that it is dev only, so thought it's not needed to run through that check each time unnecessary.

chabou: If `localPlugins`is empty, it will do nothing. It is pretty useless you're right, but it is better for consistency imo.

PR #1935, opened 2017-06-14
Escape parentheses for dropped files/folders

albinekb:
Maybe it's better to just wrap the path in `""`? But then we need to escape other stuff? 🤔 

Edit: yeah, using `'` is the best approach here:
http://wiki.bash-hackers.org/syntax/quoting#strong_quoting


rauchg:
Amazing. Thank you @albinekb

PR #1957, opened 2017-06-25
Add Travis dep for Linux

chabou:
`electron-builder` compatibility
See https://github.com/electron-userland/electron-builder/wiki/Multi-Platform-Build#to-build-app-in-32-bit-from-a-machine-with-64-bit

Fixes https://travis-ci.org/zeit/hyper/jobs/246404332

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,8 @@ matrix:
 addons:
   apt:
     packages:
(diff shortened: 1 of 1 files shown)


chabou:
Fix confirmed: https://travis-ci.org/zeit/hyper/jobs/246889698

Issue #1589, opened 2017-02-27
Crashes on open with invalid config


sambass1:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: <macOS Sierra 10.12.3>
- **Hyper.app version**: <Hyper 1.2.1>

When I open Hyper the terminal open and closed quickly. The app is still open however. When I click on preferences in the menu bar the below pop up box shows. I've tried uninstalling and reinstalling but same problem occurs.

**A JavaScript error occurred in the main process**
Uncaught Exception:
TypeError: Cannot read property 'write' of undefined
    at Server.rpc.on (/Applications/Hyper.app/Contents/Resources/app/index.js:269:24)
    at emitOne (events.js:96:13)
    at Server.emit (events.js:188:7)
    at Server.ipcListener (/Applications/Hyper.app/Contents/Resources/app/rpc.js:34:11)
    at emitTwo (events.js:106:13)
    at EventEmitter.emit (events.js:191:7)
    at WebContents.<anonymous> (/Applications/Hyper.app/Contents/Resources/electron.asar/browser/api/web-contents.js:219:13)
    at emitTwo (events.js:106:13)
    at WebContents.emit (events.js:191:7)

Anzumana:
How did you uninstall the app? You could try https://freemacsoft.net/appcleaner/
And then reinstall.

sambass1:
I used appcleaner to remove it and still the same issue. Also installed the older version of Hyper and the same problem.

I think it's an issue in the preferences but I can't edit those without using the terminal.

rohan-buchner:
I had the same issue. Open the `~/.hyper.js` file in a regular editor / other terminal. then make sure the file has no typos in in. In my case the plugins array had an extra comma in it.

sambass1:
Great, advice. Opened the  ~/.hyper.js file in a regular editor and changed what I needed. Working fine now. Thanks @rohan-buchner

albinekb:
I think we should handle that error and show a nice alert so next time the user knows that there's something wrong in the Preferences👍🙌

CaRDiaK:
I'm having this exact problem with Windows. I also see other issues listed. I made a backup of my config before I changed it to use powershell. I'm now assuming one of my args is wrong because it just throws an uncaught exception on startup. 

[...]

PR #1966, opened 2017-07-01
Added validation for config.shell 

ChrHan:
Added validation for `shell` part of config.

Also revert invalid shell (which for now just looks for valid file) to default.
This should fix issue #1589, and branch is ready to merge.

### DIFF
--- a/app/config/init.js
+++ b/app/config/init.js
@@ -1,3 +1,4 @@
+const fs = require('fs');
 const vm = require('vm');
 const notify = require('../notify');
 
@@ -14,6 +15,7 @@ const _syntaxValidation = function (cfg) {
   try {
     return new vm.Script(cfg, {filename: '.hyper.js', displayErrors: true});
   } catch (err) {
+    notify(`Configuration Error! Please fix ~/.hyper.js!`);
     notify(`Error loading config: ${err.name}, see DevTools for more info`);
     console.error('Error loading config:', err);
     return;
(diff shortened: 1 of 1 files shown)


app/config/init.js
```diff
@@ -14,6 +15,7 @@ const _syntaxValidation = function (cfg) {
   try {
     return new vm.Script(cfg, {filename: '.hyper.js', displayErrors: true});
   } catch (err) {
+    notify(`Configuration Error! Please fix ~/.hyper.js!`);
```
CodeTheory: There's no need for this since one line below we show a more in-depth notification.

ChrHan: Reverted on commit ce4bedd

app/config/init.js
```diff
@@ -39,6 +40,10 @@ const _init = function (cfg) {
     // Ignore undefined values in plugin and localPlugins array Issue #1862
     _cfg.plugins = (_cfg.plugins && _cfg.plugins.filter(Boolean)) || [];
     _cfg.localPlugins = (_cfg.localPlugins && _cfg.localPlugins.filter(Boolean)) || [];
+    if (!fs.existsSync(_cfg.config.shell)) {
```
ppot: test this before. No need to add fs there. The script will know it if `_cfg.config.shel` exist before the assignation.

ChrHan:
@Stanzilla not sure if I'd be able to pick this PR on the following weeks; you can close it if necessary

chabou:
As discussed this is not the right solution for this problem.

BTW thank you so much for your help and really sorry for answer delays 🙏

PR #1967, opened 2017-07-02
Update to Webpack 3. Enable ModuleConcatenation

jeremenichelli:
I build locally a version of hyper with Webpack 3 and scope hoisting in the build. I haven't spotted any bugs on Mac (though it would be good if someone on Linux and Windows could generate a build from this branch and test).

Overall, this version started faster than the current one.

I started the app several times from scratch, here are some results:

**without** scope hoisting:
- 291.74ms
- 282.86ms
- 279.19ms

**with** scope hoisting:
- 225.94ms
- 230.86ms
- 254.24ms

I'm testing in a MacBook Air and the difference can be noticeable. Of course this PR doesn't need to be merged, the change is easy and you can do it yourselves, but just as a reminder for the next release in case you want to consider it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -155,7 +155,7 @@
     "node-gyp": "3.6.2",
     "redux-logger": "3.0.6",
     "spectron": "3.6.4",
-    "webpack": "^3.0.0",
+    "webpack": "3.0.0",
     "xo": "0.17.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -155,7 +155,7 @@
     "node-gyp": "3.6.2",
     "redux-logger": "3.0.6",
     "spectron": "3.6.4",
-    "webpack": "2.6.1",
+    "webpack": "^3.0.0",
     "xo": "0.17.1"
   }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -155,7 +155,7 @@
     "node-gyp": "3.6.2",
     "redux-logger": "3.0.6",
     "spectron": "3.6.4",
-    "webpack": "2.6.1",
+    "webpack": "^3.0.0",
```
CodeTheory: Could you pin the dependency?

Stanzilla:
Is tree shaking enabled by default in @3 now? If not, might be worth doing that as well http://jakewiesler.com/tree-shaking-es6-modules-in-webpack-2/

jeremenichelli:
@Stanzilla seeing that we are not transpiling modules on babel, Webpack does tree shaking by default on _harmony modules_ and dead code elimination should get rid of unused functions already. Further reading: http://2ality.com/2015/12/webpack-tree-shaking.html

jbergstroem:
Perhaps rebase [and update deps] and hope that brings more reviewers to the PR?! 👍

ppot:
Thanks for the PR, I will close since it's integrated into #1958

Issue #1535, opened 2017-02-17
Bunch of Chinese characters together.


xiamu14:
- [x] I am on the latest Hyper.app version

- [x]  I have searched the issues of this repo and believe that this is not a duplicate

- window 10
- window powershell
- Hyper.app v1.2.1

**Issue**
[screenshot](http://static.ohf2e.com/issue.JPG)
Just look this, bunch of Chinese characters together

xiamu14:
添加选项

rauchg:
We are working on improving support for international characters and emoji!

xiamu14:
I will always wait until the software better.

HereChen:
My terminal is really in mess. 0_0

![tim 20170421104817](https://cloud.githubusercontent.com/assets/6745477/25260679/7d30a9a0-2680-11e7-822a-32875cd33b16.png)

MartyGentillon:
I noticed the issue HereChen is referring to where the terminal decides to wrap around the edge after certain characters are displayed (It does it for me after logging this very repository).  Reminds me a little of an old tube tv set with the horizontal alignment off. 

I am pretty sure that it is in the underlying hterm library.

Stanzilla:
https://github.com/zeit/hyper/issues/2322

PR #2000, opened 2017-07-13
fix Bunch of Chinese characters together (#1535)

shuishoudage:
fixed Chinese character together problem.
Maybe fixed unicode display problem as well (not sure)
befor:
screen shot 2017-07-13 at 3 53 52 pm
after:
screen shot 2017-07-13 at 3 54 10 pm

### DIFF
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -155,7 +155,7 @@ hterm.Terminal.prototype.setFontSize = function (px) {
     .unicode-node {
       display: inline-block;
       vertical-align: top;
-      width: ${this.scrollPort_.characterSize.width}px;
+      width: 1em;
     }
   `;
 };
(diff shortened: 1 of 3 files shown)


lib/hterm.js
```diff
@@ -155,7 +155,7 @@ hterm.Terminal.prototype.setFontSize = function (px) {
     .unicode-node {
       display: inline-block;
       vertical-align: top;
-      width: ${this.scrollPort_.characterSize.width}px;
+      width: 1em;
```
Tyriar: This looks kind of dangerous as unicode characters can be both wide and narrow. Also `1em` is not precise and would therefore not align with the grid.

shuishoudage: okey, got it. I will think another way to solve it. Thank you

timothyis:
Thanks for your contribution, but we no longer use hterm, so I'll close this. 
Feel free to contribute again if you find anything you'd like to do!

Issue #265, opened 2016-07-18
Add context menu


akashnimare:
Add menus like open new tab, copy, paste, close tab, open tab etc on right click in hyperterm.

sindresorhus:
Check out [`electron-context-menu`](https://github.com/sindresorhus/electron-context-menu) if you want a good starting point. Easily extendable.

akashnimare:
@sindresorhus @rauchg is it a good idea to make a plugin for this or it should be a default feature of hyperterm ?

sindresorhus:
I think it should be default, so plugins can be able to modify it.

akashnimare:
Sounds good.

mnpenner:
In particular, I'd like to be able to right-click hyperlinks and then copy them.

rickycodes:
Opened https://github.com/zeit/hyper/pull/2001

PR #2001, opened 2017-07-13
Add context menu

rickycodes:
closes: #265 

First stab at creating a right-click context menu...

right-click-copy

/cc @leo


rickycodes:
@CodeTheory @leo this should be good to go now ;)

rickycodes:
Looking at the issue, I saw https://github.com/zeit/hyper/issues/265#issuecomment-292229142 and decided to add two items to a "selectionMenu"

selectionMenu

/cc @CodeTheory

CodeTheory:
This is incredible! Can't wait to have this in! 🎉 
cc @matheuss @rauchg @chabou @ppot

rickycodes:
I ended up just using `.concat` to add the two new items since `splice` is problematic 😓

Stanzilla:
This does and will clash with
```
// if true, on right click selected text will be copied or pasted if no
    // selection is present (true by default on Windows)
    quickEdit: true,
``` 
so I guess having it enabled should disable the menu or the other way round

Stanzilla:
Same with
```
    // set to `true` (without backticks) if you're using a Linux setup that doesn't show native menus
    // default: `false` on Linux, `true` on Windows (ignored on macOS)
    showHamburgerMenu: '',
```

in the case of Windows or true the right click menu should probably adjust itself to not duplicate items with the hamburger menu

rickycodes:
> `cut` `undo` `select all` and `redo` do not work. nothing happens.

do these items behave as expected or differently when using the main menus? (I haven't really been able to get them to work in either case), they should perform the same as the other menus (since they're just duplicated from there)

re: windows settings, I will tweak what I have to make sure there's no conflicts (and test a bit in windows) to see what makes the most sense.

/cc @Stanzilla

Stanzilla:
Yeah, they don't work from the main menu either. Do they work on any OS?

lib/components/term.js
```diff
@@ -98,9 +107,43 @@ export default class Term extends Component {
...
+      },
+      {
+        label: 'Open Selection as URL',
+        click() {
+          openExternal(selectedText);
+        },
+        enabled: !isCollapsed
+      },
+      {
+        label: 'Search the Web for Selection',
```
CodeTheory: Is there anyway to open a link with the users default search engine? That would be awesome.
Default search engine probably would come from the default browser > default search engine

rickycodes: Could leave it as Google by default and make it configurable via `.hyper.js`?

rickycodes:
> Yeah, they don't work from the main menu either. Do they work on any OS?

this is a good question, I'm not sure under what circumstances they are designed to work... the main thing is that this is not a regression on my part ;)

lib/components/term.js
rickycodes: both terminal and iTerm2 have `Search with Google`

CodeTheory: Yeah, I'd like that. I know a few people who'd like not to use Google.

henrikdahl: With Mac OS services you should be able to search with default browser and engine but most likely unusable since Hyper also supports other operative systems ✌️

rickycodes: in https://github.com/zeit/hyper/pull/2001/commits/b3a9d307f1c55b38e7267acaa773460953dd8c1e I've made this configurable via exposing a new key/value (`searchEngineUrl`) from config so this can be customized via `.hyper.js`

defaults to google, but can be anything the user desires.

this would likely need to be address in `app/config/config-default.js` as well (similar to `bellSoundURL`)

/cc @henrikdahl @CodeTheory

rickycodes:
I nixed the url/google functionality (as I think that should be handled in a different PR, or possibly a plugin?)

@Stanzilla this should account for the items you mentioned `quickEdit` and `showHamburgerMenu`. let me know if this makes sense

/cc @CodeTheory

Stanzilla:
`showHamburgerMenu` is a bit problematic, the default is `''` which is sort of `auto`, that means it is actually `true` on Windows even if it is not set to `true` in the config. That is a problem because a Windows user can't set it to `false`when he would prefer using your context menu over the hamburger and also makes you checking for it kinda useless since it will never be `false` on Windows. 

My concern was to not duplicate menu entries when both hamburger and context are enabled but since you removed the URL/Google option it's 100% the same anyway, no?

rickycodes:
@Stanzilla what do you think makes the most sense in your opinion? should I just drop the `showHamburgerMenu` check? I'm not entirely sure what makes the most sense for this as I am not a windows user

Stanzilla:
I'd drop the check, yeah. Makes the most sense to me.

rickycodes:
@Stanzilla I dropped that `showHamburgerMenu`

lmk if this makes sense

ppot:
@rickycodes Are you binding the menu context with keymaps?

rickycodes:
@ppot I'm just building the context menu from what already exists (combining edit + shell `.submenu`'s here: https://github.com/zeit/hyper/pull/2001/files#diff-ac33cd5e16269e3f364363f9865b9ff5R24)

so yea, they have the same `role/accelerator` properties, if that what you mean?

app/menus/menu.js
```diff
@@ -17,6 +17,16 @@ const appVersion = app.getVersion();
 
 module.exports = (createWindow, updatePlugins, getLoadedPluginVersions) => {
   const commands = getKeymaps().commands;
+
+  const contextMenuTemplate = () => {
+    const seperator = {type: 'separator'};
+    return editMenu(commands).submenu
+    .concat(seperator, shellMenu(commands, createWindow).submenu);
```
ppot: you are using it once. no need for `const` there 
simply use {type: 'separator'}

ppot:
@rickycodes The problem that I see is that, since you expose the menu to plugins. They will update the base menu instead of the contextMenu. You should make a copy of the menu and then add your own items to it and those of plugins.

rickycodes:
@ppot I'm using `.concat`, so it actually _is_ a copy already:

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/concat
> The concat() method is used to merge two or more arrays. **This method does not change the existing arrays, but instead returns a new array.**

since the copy is all that's exposed in `menu.js` there's no way to alter the original `. submenu`s

rickycodes:
unless I am missing something?

chabou:
Sorry for the delay, I was on vacation.
Amazing job here!! 😍 

Reviewing it right now

app/menus/menu.js
```diff
@@ -17,6 +17,15 @@ const appVersion = app.getVersion();
...
 module.exports = (createWindow, updatePlugins, getLoadedPluginVersions) => {
   const commands = getKeymaps().commands;
+
+  const contextMenuTemplate = () => {
+    return editMenu(commands).submenu
+    .concat({type: 'separator'}, shellMenu(commands, createWindow).submenu);
+  };
+
+  // expose to plugins
+  app.contextMenuTemplate = contextMenuTemplate();
```
chabou: It could be better to have same way than menu, to decorate it from plugins:
Plugins export a `decorateContextMenu` function which is called when app start like [here](https://github.com/zeit/hyper/blob/de8e70f935c79bb25d996ec53bd87ca7a525922c/app/index.js#L176) 

Or maybe call a plugin exported function when context menu will show for a given term. With this function, plugin can (un)check some item according to its internal state for this term. Very useful for plugin like `hyper-broadcast` or for a plugin that will let user define term style directly from context menu 😍

lib/components/term.js
```diff
@@ -98,9 +106,24 @@ export default class Term extends Component {
...
     this.getScreenNode().addEventListener('mousedown', this.handleMouseDown);
+    this.getScreenNode().addEventListener('contextmenu', this.handleRightClick);
     terms[this.props.uid] = this;
   }
 
+  handleRightClick(e) {
+    e.preventDefault();
+    e.stopPropagation();
+
+    const {quickEdit} = config.getConfig();
```
chabou: You dont have to import/check `config`. You should check `this.props.quickEdit` like  [here](https://github.com/zeit/hyper/blob/5ec705000f73dbad5f70b9152708cbeaf4e3c086/lib/components/term.js#L283)

lib/components/term.js
```diff
@@ -98,9 +106,24 @@ export default class Term extends Component {
...
+    const show = !quickEdit;
+
+    const contextMenu = Menu.buildFromTemplate(contextMenuTemplate);
+    if (show) {
+      // throttle popup so right click has time to highlight text
+      setTimeout(() => contextMenu.popup(getCurrentWindow()), 100);
```
chabou: I am really not fan to call this here.
You are right, context menu should be contextual to right-clicked term.
But I would prefer to dispatch a redux action (and maybe reduce it to retain in state which term is concerned), make a rpc call and popup context menu from app.

`Term` is a component and should not have any knowledge of `electron`.

rickycodes: > Term is a component and should not have any knowledge of electron.

ya 😅 

i had wondered about this

I'll take a look at setting this up

chabou: Do not hesitate to ask question in this PR or directly on slack!
Really happy to help you on this awesome feature!

rickycodes: @chabou I've made some changes that add a redux action + state/reduce

it's currently toggling the contextMenu as an `effect` in the actions, which I don't think is quite right :thinking:

> make a rpc call and popup context menu from app

could you maybe elaborate on that, where _ideally_ would you like to see this/what are your thoughts on setting that up? (sorry this code base is still new to me) 😅

chabou: It could be better to make a `rpc.emit` in this `effect()`.
`contextMenu.popup` should occur in MainProcess not in the renderer.

rickycodes:
@chabou I've gone ahead and added the 'open context menu' rpc event
all menu related code has been removed from client now (`/lib`)

rickycodes:
@chabou have you had a chance to review these changes? I noticed some stuff got moved around in `master`, I can go ahead and adjust if this is looking good to you.

LMK

lib/reducers/ui.js
```diff
@@ -251,6 +253,7 @@ const reducer = (state = initial, action) => {
 
     case SESSION_SET_ACTIVE:
       state_ = state.merge({
+        contextMenuOpen: false,
```
chabou: I'm not sure it is a good idea to maintain context menu state in renderer redux store. If you want to keep this, we must find a better way to pass it to `false`. Because I think that if you click in another application window, contextMenu will close but you wont have a `SESSION_SET_ACTIVE` redux action.

app/index.js
```diff
@@ -290,6 +293,14 @@ app.on('ready', () => installDevExtensions(isDev).then(() => {
       Menu.getApplicationMenu().popup(Math.ceil(x), Math.ceil(y));
     });
 
+    rpc.on('open context menu', () => {
+      const commands = getKeymaps().commands;
+      const contextMenuTemplate = editMenu(commands).submenu
+        .concat({type: 'separator'}, shellMenu(commands, createWindow).submenu);
+      const contextMenu = Menu.buildFromTemplate(contextMenuTemplate);
```
chabou: Plugin can indirectly modify context menu because you reconstruct its template each time. Maybe enough and improvements about plugin decoration could be done in another PR.

app/index.js
```diff
@@ -290,6 +293,14 @@ app.on('ready', () => installDevExtensions(isDev).then(() => {
...
+      setTimeout(() => contextMenu.popup(), 100);
```
chabou: Why do you need this `setTimeout`? If you really need this, you should add a comment to explain why.

chabou:
I really want to merge this PR one day. So you can rebase with `master` and resolve conflict, it won't be a waste of time.

rickycodes:
@chabou I've gone ahead and synced things up and made the adjustments you suggested

> improvements about plugin decoration could be done in another PR

I think this is a good call, get this in now and worry about plugin decoration in another PR? and maybe even create a couple plugins based on some of the work I did previously: eg "Search the Web for Selection" and "Open Selection as URL" to start?

chabou:
Oh, I didn't understand that right-clicking over a word select it. Nice! Exactly like iTerm.

This is a big feature and I need to review it and test it before merge. It can take time but be sure that I want it in master ASAP.

Thank your for your help and your well illustrated comments ;)

vladikoff:
This is gonna be great!

chabou:
I got an exception when I right-clicked on a Hyper windows that haven't the focus.
hyper exception 2017-09-09 a 07 59 50

rickycodes:
@chabou are you able to provide some additional details? I can't seem to reproduce :(

maybe it should focus the pane on right click just before doing anything else? I'll take a quick look before Monday

chabou:
@rickycodes 
HTR:
* Open Hyper window.
* Click on another application window
* Right-click on hyper.
=> exception

rickycodes:
@chabou ah yes, i see it!

maybe we need to focus the application window beforehand? any thoughts?

chabou:
When I right-click unfocus App on Mac, context menu is opening but app keep unfocused (Menu still concerns previously focused app).
It would be great to have the same behaviour

rickycodes:
@chabou calling `window.focus()` beforehand seems to clear up the `exception` issue you caught: https://github.com/zeit/hyper/pull/2001/commits/01ba38ba4b9ead19a1a87cc0e80d31dcbb6f055e

rickycodes:
> It would be great to have the same behaviour
ah, I see so calling `window.focus()` is not an ideal solution... but I am not entirely sure how else to go about this :thinking:

chabou:
@rickycodes In [Menu.popup() documentation](https://github.com/electron/electron/blob/master/docs/api/menu.md#menupopupbrowserwindow-options) we can read:
```
menu.popup([browserWindow, options])
  * browserWindow BrowserWindow (optional) - Default is the focused window.
```
In our case (right-clicking on an unfocused Hyper window), none is focused. I think we need to tell  to Electron on which BrowserWindow the context menu should popup.
`Menu.buildFromTemplate(contextMenuTemplate).popup([window]);` should do the trick.

rickycodes:
@chabou I just tried that out, but it still seems to cause an exception :(

chabou:
Great! I will review/test it ASAP.

chabou:
There is something that needs to be improved.
A right-click on an unfocused pane gives focus but no `SESSION_SET_ACTIVE` redux action is triggered. And it leads to an inconsistent state:
* You are in Pane1
* Open a new Pane2
* Right-click on the unfocused Pane1 and paste something
* you can type some chars
* Hit split pane shortcut
* A new pane is opened but with Pane2 as origin

It is a good thing to change focus on right click, but we need to reflect this in Redux state.

rickycodes:
@chabou not too sure what the best approach to solve this is... I've added a `dispatch` call to `setActiveSession` in the `onRightClick` call just before the `dispatch` call to `handleRightClick`

let me know what your thoughts are

matheuss:
LGTM – just met ricky and he’s nice therefore the PR must be nice as well

Stanzilla:
Probably makes more sense to switch the PR to targeting canary now, no?

rickycodes:
@Stanzilla yea I was wondering about that, @chabou is this something I should do?

chabou:
Targeting canary would be awesome. But you'll need to merge canary to your branch before (fixing potential conflict and lint errors because we are using prettier now).

But I'm a little bit embarrassed to ask you that because we should have merged your PR before.
If you don't have time for this, I'll merge your PR in master and merge master back in Canary. I'll resolve potential conflict/lint error by myself.

rickycodes:
@chabou I just tried to rebase over canary myself and am noticing ALOT of changes 😢 

if you want to take a stab at porting changes that would be very nice

rickycodes:
@chabou this has now been rebased over `zeit:canary` and is now targeting it ;)

I had to make a couple adjustments to accommodate the new changes

lmk what you think

app/ui/window.js
```diff
@@ -15,8 +15,9 @@ const editMenu = require('../menus/menus/edit');
 const shellMenu = require('../menus/menus/shell');
 const {getKeymaps: commands} = require('../config');
 
-const contextMenuTemplate = editMenu(commands).submenu
-  .concat({type: 'separator'}, shellMenu(commands, app.createWindow).submenu);
+const contextMenuTemplate = createWindow => {
```
rickycodes: @chabou I've been toying around with this a bit
```
const contextMenuTemplate = createWindow => {
  const shell = shellMenu(commands, createWindow).submenu
  const edit = editMenu(commands).submenu
  return edit.concat({type: 'separator'}, shell);
};
```

app/ui/window.js
```diff
@@ -15,8 +15,9 @@ const editMenu = require('../menus/menus/edit');
...
+  return editMenu(commands).submenu.concat({type: 'separator'}, shellMenu(commands, createWindow).submenu);
+};
```
rickycodes: @chabou I've been toying around with this a bit
```js
const contextMenuTemplate = createWindow => {
  const shell = shellMenu(commands, createWindow).submenu
  const edit = editMenu(commands).submenu
  return edit.concat({type: 'separator'}, shell);
};
```

app/ui/window.js
```diff
@@ -11,6 +11,13 @@ const createRPC = require('../rpc');
 const notify = require('../notify');
 const fetchNotifications = require('../notifications');
 const Session = require('../session');
+const editMenu = require('../menus/menus/edit');
+const shellMenu = require('../menus/menus/shell');
+const {getKeymaps: commands} = require('../config');
+
+const contextMenuTemplate = createWindow => {
+  return editMenu(commands).submenu.concat({type: 'separator'}, shellMenu(commands, createWindow).submenu);
+};
```
rickycodes: maybe an alternative to this:
@chabou I've been toying around with this a bit
```js
const contextMenuTemplate = createWindow => {
  const shell = shellMenu(commands, createWindow).submenu
  const edit = editMenu(commands).submenu
  return edit.concat({type: 'separator'}, shell);
};
```

rickycodes:
@chabou have you had a chance to look at this? I've actually made some refinements, and have an example where the menu changes/updates based on selection... in this version I've made it so `cut` and `copy` are _only_ available if text is selected... text selection also expands a couple other options: `Open Selection as URL` and `Search the Web for Selection`:

hyper

you can compare the changes here: https://github.com/rickycodes/hyper/compare/rickycodes:issue-265...add-context-menu

lmk what you think!

lib/components/terms.js
```diff
@@ -64,6 +64,13 @@ export default class Terms extends Component {
     this.terms[uid] = term;
   }
 
+  componentDidMount() {
+    window.addEventListener('contextmenu', () => {
```
chabou: I didn't know/see there was this event.
In this case, please remove any mention of `right click` notion. We should only speak about context menu.

chabou:
I love the idea to pass selection in redux action.
Please make it happen :pray:

rickycodes:
@chabou I've made the requested changes

chabou:
I have a problem with selection with your PR.
When I right click to show contextMenu at a point "P", when I discard contextMenu, `xterm` is on SelectionMode starting on this point "P" (but I haven't any mouse button pressed).
See:
contextmenu

chabou:
Can you remove "Open Selection as URL" and "Search the webfor Selection"? I don't think it should be a core feature. But it is perfect to filter Copy/Cut in contextMenu depending on selection presence.

rickycodes:
@chabou I've removed the extended menu with the additional options. I noticed canary has the ability to click on URLs anyway, so maybe `Open Selection as URL` isn't so useful now

we can discuss creating plugins to accomplish those items once this lands/how to expand on this so plugin authors can modify the context menu

Stanzilla:
The CircleCI failure looks unrelated.

rickycodes:
@Stanzilla yea i noticed that, but it seems unrelated (the build was passing earlier and I haven't made any adjustments that would have changed that)

chabou:
I can't reproduce https://github.com/zeit/hyper/pull/2001#issuecomment-340436323 anymore 🤔 

Let's merge this. We'll fix bugs if needed: not a big deal.

chabou:
@rickycodes I can't  push to your branch. Did you allow maintainers to modify your branch?
I resolved conflicts (in GitHub interface) and Github doesn't see my last commit 😞.
Can you push a empty commit to trigger it ?

rauchg:
This feature is 🔥

chabou:
Thank you so much for your help and your awesome work 🙏 
Really sorry for the multiple delays along this PR.

PR #2005, opened 2017-07-16
Enabled adding font-weight in config file to change font weight

asthas:

### DIFF
--- a/lib/reducers/ui.js
+++ b/lib/reducers/ui.js
@@ -38,6 +38,7 @@ const initial = Immutable({
   cursorBlink: false,
   borderColor: '#333',
   fontSize: 12,
+  fontWeight: 500,
   padding: '12px 14px',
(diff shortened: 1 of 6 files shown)


lib/reducers/ui.js
```diff
@@ -38,6 +38,7 @@ const initial = Immutable({
   cursorBlink: false,
   borderColor: '#333',
   fontSize: 12,
+  fontWeight: 500,
```
ppot: why setting `500` when you're setting `600` in config-default? 

also what is the default look like?

asthas: @ppot I fixed that but the problem I am facing is that when I tried adding font-weight property ), it seems like the hterm package doesn't support font-weight property. Should I overwrite hterm's package set method? Any other way around it? Any suggestion would be appreciated.

chabou:
based on Hterm, this PR can't be merged.
Sorry but I have to close it.

Thank you for your help 🙏

PR #2053, opened 2017-08-06
Fix travis build by removing extra snap package

Stanzilla:
(maybe) Fix build by removing extra snapcraft package that is no longer available.


ppot:
We need to be able to merge other PR in order to advance the app.

PR #2057, opened 2017-08-08
`window` instead of `this`

ppot:


ppot:
@CodeTheory Can you just accept this so I can merge

PR #2062, opened 2017-08-09
update auto-updater

ppot:
This update the updater process for Hyper using a single entry point

### DIFF
--- a/app/auto-updater.js
+++ b/app/auto-updater.js
@@ -8,7 +8,7 @@ const {version} = require('./package');
 const {platform} = process;
 const FEED_URL = `https://releases.hyper.is/update/${platform}`;
 
-const start = () => {
+const initialize = () => {
   if (!isDev && process.platform !== 'linux') {
     autoUpdater.on('error', (err, msg) => {
       console.error('Error fetching updates', msg + ' (' + err.stack + ')');
@@ -47,5 +47,5 @@ const start = () => {
 };
 
 module.exports = {
-  start
+  initialize
 };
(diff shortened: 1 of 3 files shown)


chabou:
Good catch, it is better to check for update availability only once but notification should occur in all opened windows, not only the last focused one.

app/auto-updater.js
```diff
@@ -1,50 +1,45 @@
...
-
-function init() {
-  autoUpdater.on('error', (err, msg) => {
-    console.error('Error fetching updates', msg + ' (' + err.stack + ')');
-  });
-
-  autoUpdater.setFeedURL(`${FEED_URL}/${version}`);
-
-  setTimeout(() => {
-    autoUpdater.checkForUpdates();
```
chabou: It is better to keep this first call.
We can't wait 30mn after start to notify user that an update is available. If user uses Hyper 15mn per day, he'll never see updates.

app/auto-updater.js
```diff
@@ -1,50 +1,45 @@
...
   }
+};
 
-  const {rpc} = win;
-
-  const onupdate = (ev, releaseNotes, releaseName) => {
-    rpc.emit('update available', {releaseNotes, releaseName});
-  };
-
-  autoUpdater.on('update-downloaded', onupdate);
```
chabou: We need this callback.

app/auto-updater.js
```diff
@@ -1,50 +1,51 @@
...
-    autoUpdater.checkForUpdates();
-  }, ms('30m'));
-
-  isInit = true;
-}
 
-module.exports = function (win) {
-  if (!isInit) {
-    init();
+    autoUpdater.on('checking-for-update', () => {
```
chabou: What is this event?
Never triggered 😞

ppot:
Closing since can't work on this unless I have an Apple developper ID and can code sign.

PR #2070, opened 2017-08-10
Normalize shell name before selecting URL-matching regex

KuroTsuto:
Ready for merge - no work left. One potential snag is the additional `'path'` import; implementing the logic directly without relying on the import wouldn't be a problem, though.

This fix improves URL/webview support by normalizing shell identifiers prior to attempting to select a respective URL-matching pattern from `lib/utils/url-regex.js`'s (shell-identifier-keyed) exports. This mitigates an issue where setting `config.shell` to an absolute path, and/or a binary name including the file extension would disable the URL/filepath-triggered webview behavior.

### DIFF
--- a/lib/utils/url-command.js
+++ b/lib/utils/url-command.js
@@ -1,7 +1,8 @@
+import path from 'path';
 import * as regex from './url-regex';
 
 export default function isUrlCommand(shell, data) {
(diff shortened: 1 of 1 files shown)


ppot:
I will integrate it. However we plan on moving webviews to a specific component later on ;)

PR #2083, opened 2017-08-14
Use yarn instead of npm for scripts.

nesl247:
Since everything (README, etc.) mentions yarn instead of npm, this PR changes all the scripts to use it.


chabou:
Thank you for your help @nesl247 ❤️

PR #2088, opened 2017-08-15
Tabs and Panes

ppot:
- Include Tabs and Panes as object for windows.
- Needed for session restoration


chabou:
I think it is possible to have session restoration without this.

Can you detail what is currently blocking this feature in our TermGroup architecture ?

ppot:
Things changed. Moving on to a better approach.

PR #2108, opened 2017-08-21
Plugins revamp

ppot:
[WIP] As part of keymaps PR #1509, Include plugins code structure optimized.

- extentions file
- plugs path
- plugs folder creation inside config


ppot:
@albinekb Not at all, just code moving for now!

albinekb:
Allright @ppot beautiful 👌

PR #2117, opened 2017-08-25
Added yarn install to contribute instructions

isseiler:
The contribute instructions don't mention installing yarn ahead of time in case the builder doesn't already have it installed.


chabou:
Thank you for your help 🙏

PR #2125, opened 2017-08-30
Using options from default config when not specified in user config

okonek:
App uses options from default config when not specified in user config. That's issue #1588. Now if you won't specify an option in your config, it will be automatically added from default config.

### DIFF
--- a/app/config/init.js
+++ b/app/config/init.js
@@ -36,6 +36,10 @@ const _init = function (cfg) {
       notify('Error reading configuration: `config` key is missing');
       return _extractDefault(cfg.defaultCfg);
     }
+
+    // Use options from default config  when not specified in user config #1588
+    _cfg.config = Object.assign({}, _extractDefault(cfg.defaultCfg).config, _cfg.config);
+
     // Ignore undefined values in plugin and localPlugins array Issue #1862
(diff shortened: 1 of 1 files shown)


chabou:
Thank you for your help.
We'll review it ASAP.

app/config/init.js
```diff
@@ -36,6 +36,10 @@ const _init = function (cfg) {
       notify('Error reading configuration: `config` key is missing');
       return _extractDefault(cfg.defaultCfg);
     }
+
+    // Use options from default config  when not specified in user config #1588
+    _cfg.config = Object.assign({}, _extractDefault(cfg.defaultCfg).config, _cfg.config);
```
Florian-R: Just curious. Are we sure the `config` will always be flat or shall it be a deepMerge here instead of `Object.assign`?

okonek: Thank you for review. I'll change it.

okonek:
I'll create next PR in second

chabou:
FYI: Next time, you don't have to make another PR. You only need to commit in your branch. PR will be automatically updated.

okonek:
Ok sorry. This is on of my first contributions to open-source. But now there is new PR: #2126

albinekb:
No problem @okonek 👍  🎉

okonek:
Could you review it?

PR #2126, opened 2017-08-30
Using options from default config when not specified in user config using DeepMerge

okonek:
App uses options from default config when not specified in user config. That's issue #1588. Now if you won't specify an option in your config, it will be automatically added from default config.


okonek:
Anyone?

ppot:
@albinekb I don't know if it's  a good Idea to use `deepmerge`. If we go in the way to allow only custom flag. Since this is what it is. Then in this way we need to remove the whole install process of `.hyper.js` file. And only read and assign base on available flags. For this. I would say no on this part.

app/config/init.js
```diff
@@ -1,4 +1,5 @@
 const vm = require('vm');
+const deepmerge = require('deepmerge');
```
chabou: I prefer to use `'lodash/merge'` here. More widely used.

okonek:
I'll change tomorrow

chabou:
Oh I think I understood.
When I used `lodash/merge` notation, it didn't mean `lodash` or `merge`.
Add `lodash` dependency and cherry-pick `merge` method when requiring:
`const merge = require('lodash/merge');`

chabou:
And please use `yarn`. `npm@5` is ok but our project use `yarn`.

okonek:
Ok. :) I understand like lodash or merge

chabou:
Can you remove `package-lock.json` files and resolve conflicts?

app/package.json
```diff
@@ -19,6 +19,8 @@
     "electron-config": "1.0.0",
     "electron-is-dev": "0.3.0",
     "electron-squirrel-startup": "1.0.0",
+
```
chabou: Can you remove this extra blank line?

okonek: Ok :)

okonek:
Anything else?

chabou:
Can you remove your `package-lock.json` files please?
(Sorry for the delay)

okonek:
Done

chabou:

okonek:
Thank You too. I learned how is it to contribute to open-source project.

PR #2130, opened 2017-08-30
Ensure correct update channel is displayed in About window

leo:


app/menus/menu.js
```diff
@@ -19,7 +19,11 @@ module.exports = (createWindow, updatePlugins, getLoadedPluginVersions) => {
   const config = getConfig();
   const {commands} = getKeymaps();
 
-  const updateChannel = config.canaryUpdates ? 'canary' : 'stable';
+  let updateChannel = 'stable';
```
albinekb: I think this logic should live inside the config, so that it's populated correctly whereever we need this? But OK for now

leo: We have different logics in those two places at the moment, but yep - definitely! 👍

PR #2146, opened 2017-08-31
Updated dependencies to the latest version

leo:
There comes a time for any big project where you need to ensure that old versions of your dependencies don't get dragged along until infinity. IMO, we just reached that point for Hyper (the longer we wait, the more is piling up).

I'm carefully testing this locally to ensure that everything is working afterwards and I suggest you (as a reader) do that as well, if you have time.


app/index.js
```diff
@@ -55,8 +55,6 @@ const isDev = require('electron-is-dev');
 const AppMenu = require('./menus/menu');
 const config = require('./config');
 
-app.commandLine.appendSwitch('js-flags', '--harmony-async-await');
```
albinekb: With this removed, I assume electron runs node 8 now? then we should change travis to reflect the same node version (so that tests run the same node version)  (...for when we have tests 😄 )

leo: Yup. Done! e0bb8b7

Issue #2156, opened 2017-09-01
Close session conflicts with quit hotkey


MarkDaviesEsendex:
- [x ] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x ] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- OS: Windows 10
- Hyperjs version: 1.4.2
- hyper.js file: https://gist.github.com/MarkDaviesEsendex/f4d80724973952a4e5d27ba1c9b40713

## Issue
The CTRL+SHIFT+W hotkey is used for both close session and quit meaning that I have to close a tab with x button or middle mouse click

chabou:
oups 😊 
Thank you for your feedback

louislva:
Would CTRL+SHIFT+W for window and CTRL+ALT+W for session be suitable?

chabou:
`ctrl+alt`is not a suitable modifier. See https://github.com/zeit/hyper/issues/2179#issuecomment-327690367
We should find another shortcut

peterdanis:
I suggest:
"window:close": "ctrl+shift+q" (q as quit)
"pane:close": "ctrl+shift+w"

PR #2160, opened 2017-09-02
Changed "Close Session" shortcut so it doesn't have the same as "Quit"

louislva:
Fixed issue #2156

### DIFF
--- a/app/keymaps/linux.json
+++ b/app/keymaps/linux.json
@@ -18,7 +18,7 @@
   "pane:prev": "ctrl+pagedown",
   "pane:splitVertical": "ctrl+shift+d",
   "pane:splitHorizontal": "ctrl+shift+e",
-  "pane:close": "ctrl+shift+w",
+  "pane:close": "ctrl+alt+w",
   "editor:undo": "ctrl+shift+z",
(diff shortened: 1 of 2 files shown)


chabou:
@phfontess `ctrl+tab` and `ctrl+shift+tab` will certainly be used to cycle between tabs like Chrome

phfontess:
Oh I'm so sorry. I wanted to say only Ctrl + W, as it is used in version 1.3.3.

louislva:
Ctrl+W for closing panes would be problematic as Ctrl is used as a character in most terminals including hyper. Lots of command line apps probably depend on Ctrl + W for their own shortcuts.

chabou:
`ctrl+w` is used to delete word before cursor in bash/zsh
And agree with @louislva `ctrl+[a-z]` should be reserved to shell and command line programs.

chabou:
Thank you so much @louislva

louislva:
Thanks for letting me contribute! I've been wanting to get into OSS contributing for a while now, and the hyper community seems really friendly.

chabou:
You are really welcome and we are proud to have your first OSS contribution 🙏

Issue #2097, opened 2017-08-18
[Linux] desktop file is lacking a Categories= entry


probonopd:
The desktop file is lacking a `Categories=` entry in the `*.desktop` file.

According to the [menu spec](https://standards.freedesktop.org/menu-spec/latest/apa.html),

> By including one of the Main Categories in an application's desktop entry file, the application will be ensured that it will show up in a section of the application menu dedicated to this category. If multiple Main Categories are included in a single desktop entry file, the entry may appear more than once in the menu. 

Hence, please add at least one of the following in the `Categories=` key.

Main Category | Description | Notes
-- | -- | --
AudioVideo | Application for presenting, creating, or processing multimedia (audio/video) |  
Audio | An audio application | Desktop entry must include AudioVideo as well
Video | A video application | Desktop entry must include AudioVideo as well
Development | An application for development |  
Education | Educational software |  
Game | A game |  
Graphics | Application for viewing, creating, or processing graphics |  
Network | Network application such as a web browser |  
Office | An office type application |  
Science | Scientific software |  
Settings | Settings applications | Entries may appear in a separate menu or as part of a 		"Control Center"
System | System application, "System Tools" such as say a log viewer or network monitor |  
Utility | Small utility application, "Accessories"

In addition, you could specify one or more from the longer list of [Additional Categories](https://standards.freedesktop.org/menu-spec/latest/apas02.html).

Please test the result with `desktop-file-validate` and make sure it passes.

electron-builder offers native support for this since [v19.22.1](https://github.com/electron-userland/electron-builder/releases/tag/v19.22.1).

Reference: https://github.com/AppImage/AppImageHub/issues/2

probonopd:
This seems still to be the case in the most recent release.

chabou:
PR Welcome 🙏

h93xV2:
I would like to make a PR on this one.

chabou:
@h93xV2 It would be awesome 😍

probonopd:
If someone could be so nice to ping me once a new AppImage is available on the Releases page, I could add it to the AppImageHub central directory of available AppImages. Thanks.

chabou:
@probonopd Just released right now!

probonopd:
Thank you.

PR #2184, opened 2017-09-07
Add category option under linux key in root package.json

h93xV2:
In response to issue entry #2097, a specific category option has been defined under the linux key in the root package.json file. This allows Hyper to take advantage of features that have been added to the most recent version of electron-builder.

Note: I tried to make this commit before, but mistakenly added app/yarn.lock, in addition to the solution of the stated issue, to the branch. Thus, this is a corrected version of my earlier attempt.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "appId": "co.zeit.hyper",
     "extraResources": "./bin/yarn-standalone.js",
     "linux": {
+     "category": "TerminalEmulator",
       "target": [
         {
           "target": "deb",


h93xV2:
@albinekb I believe this is enough. Hyper fits perfectly in the `TerminalEmulator` category.

chabou:
Thank you so much @h93xV2 !
(FYI: you don't have to make a new PR. Only need to commit in your branch to update a PR)

h93xV2:
@chabou I will remember that in the future, thanks!

PR #2185, opened 2017-09-07
Add default keymaps reference to documentation

### LINKED ISSUE #2167: Add keymaps map to documentation
Hi there,

than you for amazing app.

After 1.4.x I was't able to type like half of characters so I had to downgrade to 1.3.3. It seems to be fixed in 1.4.3 but not I'm not able to use keymaps (new tab, close tab, ...). I noticed there is new keymap config property but there is no keymap map/table/docs nor any note default keymaps has changed.

Any chance to get one?

arojunior:
closes #2167 

This PR just add a table referencing the default keymaps for Windows, Linux and MacOS. It was not in the documentation.

### DIFF
--- a/website/index.html
+++ b/website/index.html
@@ -698,6 +698,17 @@
 
 };</code></pre>
 
+        <p>Default keymaps: </p>
+
+        <table>
+          <tbody>
+            <tr>
+              <td><a href="https://github.com/zeit/hyper/blob/master/app/keymaps/win32.json">Windows</a></td>
+              <td><a href="https://github.com/zeit/hyper/blob/master/app/keymaps/linux.json">Linux</a></td>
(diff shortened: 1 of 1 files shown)


website/index.html
```diff
@@ -698,6 +698,17 @@ <h2 id="keymaps"><a href="#keymaps">Keymaps</a></h2>
 
 };</code></pre>
 
+        <p>Default keymaps: </p>
+
+        <table class="keymaps">
```
chabou: Why table has this `keymaps` class? 🤔

arojunior: You're welcome.

I was following the same pattern of the other tables as api and config with that styles and etc, but at the end I decided just put the links, so now actually there's no reason to. I can update the commit with no class if you want to.

chabou: yes please

arojunior: done.

chabou:
🙏  Thank you so much for your help

PR #2186, opened 2017-09-07
Issue 2156 fix

h93xV2:
In response to issue #2156, the keymapping conflict has been removed from linux.json and win32.json in  the app/keymaps/ directory. Now the keymap for window close is ctrl+shift+q and the keymap for pane close is ctrl+shift+w.


h93xV2:
I realized that I have two unnecessary commits in this PR. I would like to remove them before this gets merged into the master branch, thanks.

PR #2202, opened 2017-09-13
Revert #2126

### LINKED ISSUE #2200: Hyper.js crashes on Windows with shell set to PowerShell
This is a new bug for me since upgrading to 1.4.4.  

With the shell set to Powershell, launching Hyper.js will exit immediately after launching.  It looks it's passing "--login" to "powershell.exe" even though I've cleared my shellArgs config setting, which causes PowerShell to error out. 

It looks like, in 1.4.4, if shellArgs === [], it defaults back to ['--login']. 

Doing "shellArgs: ['']" resolves the issue.

        -
        Hyper 1.4.4
        Electron 1.7.6
        win32 x64 10.0.15063

chabou:
Fix #2200

### DIFF
--- a/app/config/init.js
+++ b/app/config/init.js
@@ -1,5 +1,4 @@
 const vm = require('vm');
-const merge = require('lodash/merge');
 const notify = require('../notify');
 
 const _extract = function (script) {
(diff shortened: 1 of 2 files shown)


zimme:
I believe #2126 is causing the base16 theme not to work in 1.4.4. I unset all default color settings in my config and setting my theme with `base16: { scheme: "tomorrow }` instead. However since 1.4.4 I guess the defaults are loaded again as I've unset them.

PR #2209, opened 2017-09-14
An add-on that must be by default

MAPESO:
The add-on is ***hyperlinks***. It is quite useful, it should come by default when you run Hyper for the first time :)

@leo

### DIFF
--- a/app/config/config-default.js
+++ b/app/config/config-default.js
@@ -111,7 +111,9 @@ module.exports = {
   //   `hyperpower`
   //   `@company/project`
   //   `project#1.0.1`
-  plugins: [],
+  plugins: [
+    'hyperlinks'
+  ],
 
   // in development, you can create a directory under
   // `~/.hyper_plugins/local/` and include it here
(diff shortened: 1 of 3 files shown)


CodeTheory:
Hi there @MAPESO! Thanks so much for the contribution.

Unfortunately, this isn't within the scope of features we intended Hyper to have initially since this isn't the expected behaviour of a terminal (thus making it perfect as an optional plugin). _However_, we're currently updating to use xterm.js which comes with this feature built in, so even if we accepted this now, the plugin would be deprecated soon. 

A side note though:
The scope of this pull request is too large. Even if there only a few changes, the changes are broad. We wouldn't be able to merge a pull request like this since it does two very different things. 
1. Changes the features of Hyper with a default plugin
2. Updates webpack so the development build changes

Ideally in future, you could keep those changes separate and keep pull request scopes very small 🙌 

Take a look at some of the issues we have around if you want to contribute still 🎉 

Thanks again, and sorry we couldn't merge

PR #2210, opened 2017-09-15
Fix #2195: Keymap definition must not require ordered keys

### LINKED ISSUE #2195: Keymap definition must not require ordered keys
For now, keymap definition should be ordered to be matched:
https://github.com/zeit/hyper/blob/48f5b6e08fac1a768d721a3b31f19693ffe0d65a/lib/utils/keymaps.js#L5-L34

If a shortcut definition has an other order than `cmd` -> `ctrl` -> `shift` -> `alt` -> `other key`. It won't work.

We should normalize keymaps internal and normalize keys in the same way before comparing.
Maybe something like: `key.split('+').sort().join('+)`?

wilsontayar:
Fix for #2195, tested in my windows machine with combinations like 'alt+shift+o', 'shift+ctrl+p', etc.

This branch is ready to be merged.

I'm using `localeCompare()` in order to properly sort all the characters because the `sort()` function sorts using unicode, causing some non english keyboards to be affected by a strange bug while normalizing.

Example:
keymap: `alt+shift+p`
normalized: `alt+p+shift`
normalized on non english keyboard: `p+alt+shift`


albinekb:
I think this is something we should write tests for! 👍

wilsontayar:
Sure! 
I've created two tests in order to check the localized keymaps.
Let me know if you find anything else. :)

lib/utils/keymaps-normalize.js
```diff
@@ -0,0 +1,23 @@
...
+    if (userKeymaps[keymap]) {
+      normalized[normalize(keymap)] = userKeymaps[keymap].toLowerCase();
+    }
+  }
+
+  return normalized;
+}
+
+module.exports.returnCommand = (keys, commands) => {
+  return getNormalizedKeyMaps(commands)[normalize(keys)];
```
chabou: Maybe better to normalize keyMaps once and for all here: https://github.com/wilsontayar/hyper/blob/73d4d9e761a4a2823c1dd883c0ef96f1a4e0c8f4/app/config/keymaps.js#L10
`toLowerCase()` is already a kind of normalization.

chabou:
@wilsontayar I removed your localeCompare function and your tests still pass. I think that the strange bug that you pointed out occurs only when KeyboardEvent key (certainly using Unicode) is normalized. It seems hard to have a unit test about this.

lib/utils/keymaps-normalize.js
wilsontayar: Definitely better to normalize once. I'll look into it. :)

wilsontayar:
Thanks for the reviews @chabou and @albinekb! 👍 
Gonna take a look into this later today for sure. I'll keep you guys updated.

wilsontayar:
I've made some changes and now, as suggested by @chabou, the normalization occurs directly at the config. 
Also updated the tests to include uppercase characters. The `toLowerCase` is important in the normalization process as well, so I thought it was a nice addition.

@chabou could you please see if your tests are still passing when you remove `localeCompare`?
I did it after the changes and it seems everything is working as expected now, gif as proof below. ;)

localized_commands

app/config/keymaps.js
```diff
@@ -1,4 +1,5 @@
 const {readFileSync} = require('fs');
+const {normalize} = require('../../lib/utils/keymaps-normalize');
```
chabou: For now, no file located in`lib/` directory is required from `app/**`.

It could be better to put this `normalize()` function in `app/utils/`

lib/utils/keymaps.js
```diff
@@ -1,4 +1,5 @@
 import {remote} from 'electron';
+import {returnCommand} from './keymaps-normalize';
 
 const {getKeymaps} = remote.require('./config');
```
chabou: If we move `keymaps-normalize` into `app` directory, we can require only a `getCommand` function.
`getCommand(keys)`.
`getKeymaps()`will be called directly in this `getCommand` function, on app side.

wilsontayar: Nice idea, I'll look into it. :)

wilsontayar:
@chabou Thanks for nice reviews, I'm learning a lot here. 😄 
I've updated everything as you suggested. If you find anything else just let me know.

app/config/keymaps.js
```diff
@@ -35,8 +36,8 @@ const _extend = function (customsKeys) {
   if (customsKeys) {
     for (const command in customsKeys) {
       if (command) {
-        commands[command] = customsKeys[command];
-        keys[customsKeys[command]] = command;
+        commands[command] = normalize(customsKeys[command]);
+        keys[normalize(customsKeys[command])] = command;
```
chabou: Maybe better to have: `const normalizedKeys = normalize(customsKeys[command])` and use it in following assignations or use `keys[commands[command]] = command` directly.

But I think this file should be totally refactored. Several things should be improved. We'll work on that later.

chabou:
We have to revert your PR, because there is error in devtools
capture d ecran 2017-09-17 a 22 54 02

ppot:
@chabou The error might be since he didn't change the reference inside: https://github.com/zeit/hyper/blob/master/lib/utils/keymaps.js

PR #2217, opened 2017-09-16
Replace `chokidar` by `fs.watch`

### LINKED ISSUE #2152: Hyper crashing since update to 1.4.x

- **OS version and name**: macOS 10.12.6
- **Hyper.app version**: 1.4.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: default hyper config
- **Relevant information from devtools** _(CMD+ALT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: N/A

## Issue
When I run docker using `docker-compose up <app name>` hyper crashes.  It never happened until I updated to 1.4.0 and it happened still with 1.4.1 and still with 1.4.2.

chabou:
As asked by @rauchg, I replaced `chokidar` by `fs.watch` or `fs.watchFile`. 
It could fix #2152.


leo:
Isn't `fs.watch` much more unreliable than chokidar? Why should we use it?

chabou:
We are suspecting Hyper crashes (https://github.com/zeit/hyper/issues/2152) to be related to filesystem.
Crashes begun when we replaced `gaze` by `chokidar`. This is maybe related.
We are watching only 1 file and seems acceptable to use `fs.watch` directly.

app/config.js
```diff
@@ -17,7 +14,13 @@ const _watch = function () {
...
-  _watcher = chokidar.watch(cfgPath, watchCfg);
+  if (process.platform === 'win32') {
+    // watch for changes on config every 2s on windows
+    // https://github.com/zeit/hyper/pull/1772
+    _watcher = fs.watchFile(cfgPath, {interval: 2000});
+  } else {
+    _watcher = fs.watch(cfgPath);
+  }
 
   _watcher.on('change', () => {
```
rauchg: are we handling `error` ?

rauchg:
@leo all those fancy watchers have created problems or added dependency weight. We're not using it in a critical path, so in the event of oddities or missed events, we're ok with that. I'm even ok with manually polling for `fs.stat`, if it weren't that there are better optimized methods for it.

LukaHorvat:
This update broke hyper for me.

chabou:
Fixed by #2220

PR #2227, opened 2017-09-18
Fix for normalize keymap internals #2210

wilsontayar:

Related: #2210, #2225, #2195.

My last PR got reverted due to errors showing up in the developer tools.
By investigating, I found out that I had imported the wrong module (instead of `require('../config')` I used `require('../config/keymaps')`).

With that fixed, a circular dependency started showing up, so I had to isolate the functions in individual modules. For better organization, I created the `/app/utils/keymaps` folder.

I've also added tests for the new individual modules.

Sorry about the last PR! 😅


app/utils/keymaps/normalize.js
```diff
@@ -0,0 +1,7 @@
+module.exports = keybinding => {
```
albinekb: Would it make sense to add a comment describing why this is needed and what its meant to do? It's not super-clear from reading this code 👍  would make it great for new contributors.

wilsontayar: Nice idea! Thanks for the tip. :)

albinekb: Great 👌

wilsontayar:
@chabou here's the fix for the broken PR, do you mind taking a look to see if anything else is missing?
@albinekb I've added a comment on the normalize module as you suggested. :)

Thanks ppl! 👍

chabou:
Thank you so much for your valuable help! 🙏

app/config/keymaps.js
```diff
@@ -1,4 +1,5 @@
 const {readFileSync} = require('fs');
+const normalize = require('../utils/keymaps/normalize');
 const {defaultPlatformKeyPath} = require('./paths');
```
ppot: @chabou you can make a PR for this 😂  `defaultPlatformKeyPath` should maybe be : `defaultFormKeyPath`

Issue #2252, opened 2017-09-21
Window sizing issues with multiple tabs on 2.0.2


richsilv:
- [ ] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version | **I'm on v2.0.2**
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Ubuntu 17.04
- **Hyper.app version**: 2.0.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/richsilv/d462c71e9089f7c25aeb36cb4aaa49cc
- **Relevant information from devtools** _(CMD+ALT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: N/A
- **The issue is reproducible in vanilla Hyper.app**: Is vanilla

## Issue
I've observed two separate issues with tab sizing in version 2.0.2:

* Opening a new (second or more) tab does not correctly resize the embedded terminal to fill the available space, leaving it occupying only a portion in the top left of the screen.

![screenshot from 2017-09-21 12-21-54](https://user-images.githubusercontent.com/3180526/30693431-7e7c0042-9ec7-11e7-906a-dedf3f987c2b.png)

* Resizing one open tab and then moving to another leaves the newly-opened tab blank until the page is resized.

![screenshot from 2017-09-21 12-07-43](https://user-images.githubusercontent.com/3180526/30692980-8a28498e-9ec5-11e7-8bae-d3824587ffcd.png)

djhi:
I reproduced it on Ubuntu 16.04 too.

dcalhoun:
I've experienced this on macOS 10.12.6 as well.

Stanzilla:
Reproduced on Windows as well https://dl.dropboxusercontent.com/s/gmlpceuwd2i1byw/2017-11-03_22-34-18.webm

chabou:
@Stanzilla 🤔  your video doesn't show this same bug
And I think that these bugs have been fixed by https://github.com/zeit/hyper/pull/2273 (landed in v2.0.3)

Stanzilla:
Oh oops, true. Making a new one then.

richsilv:
Agreed, this is fixed by #2273 - I'll close.

PR #2253, opened 2017-09-21
fix terminal sizing issues

richsilv:
Fixes https://github.com/zeit/hyper/issues/2252

* Forcibly sizes new tab terminals to prevent a problem where the `cols` and `rows` props which are provided by Redux are correct for the current window size (having been set by the original tab terminal), but these figures are not reflected in the newly created terminal tab.  Currently, the check (cols/rows in props from Redux === required cols/rows) will prevent such a resize from happening.
* Does not attempt to resize tabs which are not active when the window is resized as their bounding rectangle will have no width or height.
* Instead, forcibly resizes tabs when they are made active.

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -81,17 +81,7 @@ export default class Term extends PureComponent {
   }
 
   onOpen() {
-    // we need to delay one frame so that aphrodite styles
-    // get applied and we can make an accurate measurement
-    // of the container width and height
-    requestAnimationFrame(() => {
-      // at this point it would make sense for character
-      // measurement to have taken place but it seems that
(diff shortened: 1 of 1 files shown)


Stanzilla:
Fixes https://github.com/zeit/hyper/issues/2234

chabou:
When reviewing your PR, I noticed that, in current canary, Term aren't unmounted when changing tab.

When changing tab, Term's `componentDidMount()` is not called and `onOpen()` is not called. I think this is the real cause of these issues

rauchg:
Thanks a lot @richsilv. Submitted a similar PR with some resizing simplifications (#2273)

Issue #2286, opened 2017-09-26
Config: SyntaxError not being shown in DevTools


kajka:
Hi there!

Yesterday I updated my iMac to new macOS 10.13 and here are some issues which I came across:

When I add a default window size to the configuration file, the entire configuration scatters and returns to the default settings (fonts, their size, theme, remove all plugins etc.). The same thing happens when I add settings for the [Enhanced Tabs Plugin](https://github.com/henrikdahl/hyper-tabs-enhanced).

I get an error message when loading settings:

![screen shot 2017-09-26 at 09 52 02](https://user-images.githubusercontent.com/2843796/30849239-fe30abca-a2a1-11e7-9808-c68d28b36b25.png)

The only entry in the DevTools console leads to:

<img width="912" alt="screen shot 2017-09-26 at 09 51 49" src="https://user-images.githubusercontent.com/2843796/30849275-12a0e912-a2a2-11e7-847a-59fd1c5bb54f.png">

Another thing is my theme: so under the Sierra everything worked as it should. Under the High Sierra however for some reason I lost rounded corners in the application:

<img width="723" alt="screen shot 2017-09-26 at 10 06 22" src="https://user-images.githubusercontent.com/2843796/30849384-78cea30a-a2a2-11e7-9f9d-f409c9708639.png">

I deleted the app, wipe all settings and reinstalled - this did not help. I removed everything again and installed Canary version - problem still occurs.

---
OS: macOS 10.13 High Sierra
Hyper version: 1.4.7/2.0.3
My config file: https://gist.github.com/kajka/d333f948c65177a6300a25eb3425e427

MAPESO:
@kajka Do you have an example of rounded edges?

kajka:
Sure! Here you go:

![diskutil--shadow](https://user-images.githubusercontent.com/2843796/30850718-b6bfe346-a2a6-11e7-955e-853f11184c1e.png)

MAPESO:
The **hyper-tabs-enhanced** plug-in produces an error that causes the minimize and close buttons to disappear.

I recommend that you stop using this plugin, it seems to cause a bug with **Hyper 2.0.3 (canary)**

![bug-hyper-tabs-enhanced](https://user-images.githubusercontent.com/16585386/30851688-f5ab463c-a26e-11e7-950f-1a077edfd22c.png)

kajka:
[...]

PR #2305, opened 2017-09-28
🐛 Fix missing comma in default config

albinekb:
Doh, this is a trailing comma, not needed... but let's add it because then you can just comment out the lines under without adding the comma?
____________________
In #2286 we started getting reports of a missing comma in the config, very weird.
So i looked in the default config, and saw the comma there.

Now @peauc reported the same thing but on macOS, then i thought it couldn't be the issue I was originally thinking #2286 was about (sync issue, user problem).

and after double checking master config again i remembered that i forgot to check canary, well here's the issue, finally fixed 🎉  

Don't know when this was introduced but let's not blame anyone, this should have been caught by the linter...! 😱

### DIFF
--- a/app/config/config-default.js
+++ b/app/config/config-default.js
@@ -97,7 +97,7 @@ module.exports = {
     bell: 'SOUND',
 
     // if true, selected text will automatically be copied to the clipboard
-    copyOnSelect: false
+    copyOnSelect: false,
 
     // if true, on right click selected text will be copied or pasted if no
     // selection is present (true by default on Windows)


oswaldoacauan:
Is this actually fixed on the canary branch?

https://github.com/zeit/hyper/blob/canary/app/config/config-default.js#L101

albinekb:
yes @oswaldoacauan, the copyOnSelect prop is inside an object, and is the last one, so there should be no trailing comma (we could add it, but this is valid, hence why the linter didn't "catch" it)

oswaldoacauan:
I lost a couple minutes trying to figure out why it was saying my config was invalid since I just copy-pasted some plugin config under the `config` key.

A trailing comma here will prevent people of making the same mistake :)

albinekb:
I think so too @oswaldoacauan, but we should update the linter to require trailing commas in the config file while adding the comma there too,  so we don't forget it in the future when we add more stuff to the config.

There's no harm in adding trailing comma, it will just make it easier for everyone.

I can't work on this right now, but should be able to do it soon-ish, otherwise, PR welcome 👍

PR #2312, opened 2017-09-30
Fix plugin requiring to authorize @ in config

### LINKED ISSUE #2311: Specify plugin version in config
If you specify `<plugin>@next` in config, it works great for `.hyper_plugins/package.json`: `next` is use and not `last`. But Hyper try to access `.hyper_plugins/node_modules/<plugin>@next/` and not 
 `.hyper_plugins/node_modules/<plugin>/`

chabou:
Fix #2311

### DIFF
--- a/app/plugins.js
+++ b/app/plugins.js
@@ -199,7 +199,7 @@ exports.subscribe = fn => {
 function getPaths() {
   return {
     plugins: plugins.plugins.map(name => {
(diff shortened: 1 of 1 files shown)


rauchg:
So are we supporting both `@` and `#` for versions? If so, we should adjust the config template / website to just use `@` if it's more standard.

Issue #1266, opened 2016-12-18
Hamburger menu not visible on Linux


binjospookie:
Ubuntu 16.04

<img src="http://bit.ly/2gOPiZL" alt="hamburger's bug" />
<img src="http://bit.ly/2gOQige" alt="no HTML for menu" />

binjospookie:
<img src="http://bit.ly/2gOQige" alt="no HTML for menu" />

ppot:
Hi there. The hamburger menu is hidden by default on linux. You can enable it in your .hyper.js file here: https://gist.github.com/juangl/74346eb89777775d4f841149ef16b20f#file-hyper-js-L33 by setting showHamburgerMenu to true.

brychanrobot:
In Mint 18.1 Cinnamon, setting showHamburgerMenu to true doesn't seem to change anything.

stefanivic:
@matheuss @brychanrobot this happened to me on Arch also.

connorgolden:
Same in Elementary 0.4

bederman:
Same on Ubuntu GNOME 16.10

Simon-L:
Same for me on Linux Mint 18, no matter what value `showHamburgerMenu` is set to.
It might have to do with the window being borderless and the window decorations being applied differently based on the desktop environement.

In my case, I cannot access any of the menus of Hyper. It might be related to the shortcuts issue I'm having, as https://github.com/zeit/hyper/pull/1058 didn't fix it.

matheuss:
Could you guys make sure that you're using `'showHamburgerMenu': true` instead of `'showHamburgerMenu: 'true'?

connorgolden:
Ahh, tried messing with the quotes and what ended up working is: `showHamburgerMenu: true` (no quotes at all) That seemed to work for me. Can others confirm?

mkolo:
Removing quotes worked for me too. 
Elementary OS 0.4 Loki

Jop-V:
Instead of "without backticks", it should say "without backticks and without quotes".

chabou:
A lot of users are confused.
@Jop-V your proposal can help. PR is welcome 😄

toastflaming:
^

jcrben:
Why is this false by default? It's extremely poor UX.

PR #2315, opened 2017-10-01
Improve documentation of showHamburguerMenu option

fernahh:
**Problem**
A lot of users are confused with documentation of `showHamburgerMenu` option.

**Solution**
Improve documentation.

resolves https://github.com/zeit/hyper/issues/1266


chabou:
Thank you for your help

Issue #75, opened 2016-07-07
Add `line-height` configuration parameter


rauchg:

jasonLaster:
So, this is the closest I got to figuring out how hterm manages lineheight

https://chromium.googlesource.com/apps/libapps/+/HEAD/hterm/js/hterm_scrollport.js#676

It does not look like line-height is part of hterm's built in prefs

https://chromium.googlesource.com/apps/libapps/+/HEAD/hterm/js/hterm_preference_manager.js#105

rpunkfu:
Huuray, managed to do it! :tada: Quite hacky way, but still... Will PR later today :D 

![screen](http://i.imgur.com/I2jUXoO.jpg)

jasonLaster:
Love it. Happy to review

rpunkfu:
Working on it, still fighting with cursor node haha

spoike:
Another reason for having a lineHeight config, fonts with powerline symbols do not get proper height by default. The arrow symbol in the screenshow below is higher than the line-height.

![screen shot 2016-08-04 at 13 14 13](https://cloud.githubusercontent.com/assets/9253/17399938/632de408-5a45-11e6-8a9b-2c4a4f82a3c7.png)

[Input Mono](http://input.fontbureau.com/) was used for the screenshot above.

spoike:
Current workaround is to add the following to the `.hyperterm.js` configuration for `termCSS`:

AliveDD:
config: {
    fontSize: 11,
    lineHeight: 33,
    fontFamily: 'Hack',

... lineHeight doesnt work :(

pinn3:
@AliveDD Uh yeah, that's why this ticket exists. Parameters inside the config property are not CSS settings only

kartsims:
@spoike your workaround doesn't work for me, did you mean adding this to `.hyperterm.js` as such :

```javascript
    // custom css to embed in the main window
    css: `
      x-row > span {
        line-height: 2em;
        padding: 0.2em 0 0.02em 0;
      }
    `,
```

Thanks for your help, I am getting to know HyperTerm and loving it

spoike:
@kartsims you're almost there. it's for `termCss` instead of `css`.

colepeters:
@spoike @kartsims Is this the exact block you’re adding, including the backticks instead of default single quotes?

```js
    termCSS: `
      x-row > span {
        line-height: 2em;
        padding: 0.2em 0 0.02em 0;
      }
[...]

PR #2323, opened 2017-10-03
line-height customization fix

Cygnusfear:
Tweaked CSS  `.terminal .xterm-rows` to align with bottom instead of top. Now a margin can be set in .hyper.js such as:

`termCSS: '.terminal .xterm-rows > div {margin: 3px 0}',`

Previously that would mess up the terminal prompt. Yay for better line-height! Should fix https://github.com/zeit/hyper/issues/75 for now

### DIFF
--- a/lib/components/style-sheet.js
+++ b/lib/components/style-sheet.js
@@ -148,7 +148,7 @@ export default class StyleSheet extends React.PureComponent {
           .terminal .xterm-rows {
             position: absolute;
             left: 0;
-            top: 0;
+            bottom: 0;
           }
 
           .terminal .xterm-rows > div {


Cygnusfear:
Notice this isn't a complete fix, the prompt isn't positioned correctly on startup.

Cygnusfear:
in fact, with no additional changes to the code (ignore this pull request) this seems to be working for me:
`termCSS: '.xterm {line-height: 17px} .xterm-rows > div {display: block; line-height: 15px}'`

Tyriar:
@Cygnusfear do you have a link to the issue this is fixing?

Tyriar:
xterm.js allows configuration of the line-height via CSS, I believe calling `term.refresh(0, term.rows - 1)` afterwards will do the refresh and fix everything.

Note that this is becoming an option in v3 https://github.com/sourcelair/xterm.js/blob/66688c19a6fb336fb68fb085b98e121caf9df018/src/Terminal.ts#L82

JesseLeite:
With all respect, @Tyriar what does line-height coming to xterm v3 have to do with hyper?

Tyriar:
@JesseLeite it's my understanding that Hyper will be upgrading to v3 when it's out in the next couple of weeks.

PR #2330, opened 2017-10-04
Add new keymaps to jump between tabs

chabou:
I didn't want that users have to change multiple keymaps for this jump feature. It could have been the case if we used action definition like this:
```
'tab:jump:1': 'cmd+1'
'tab:jump:2': 'cmd+2'
...
'tab:jump:last': 'cmd+9'
```

So I decided to introduce a naming convention for this action:
``` 
'tab:jump:prefix': 'cmd"
```
It will automatically generate the 9 actions mentioned above.

This PR adds 9 menu items too:
capture d ecran 2017-10-04 a 23 39 10

### DIFF
--- a/app/menus/menus/window.js
+++ b/app/menus/menus/window.js
@@ -1,4 +1,21 @@
 module.exports = commands => {
+  // Generating tab:jump array
+  const tabJump = [];
+  for (let i = 1; i <= 9; i++) {
+    // 9 is a special number because it means 'last'
+    const label = i === 9 ? 'Last' : `${i}`;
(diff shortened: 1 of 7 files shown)


app/menus/menus/window.js
```diff
@@ -1,4 +1,21 @@
 module.exports = commands => {
+  // Generating tab:jump array
+  const tabJump = [];
+  for (let i = 1; i <= 9; i++) {
+    // 9 is a special number because it means 'last'
```
albinekb: Why not use `-1` for that?

PR #2331, opened 2017-10-04
Fix keyboard event handling

chabou:
This PR fixes issues with some combinations.

With french keyboard (but maybe US keyboard too), it was impossible to use `ctrl+alt+[a-z]` shortcut because keydown event has a modified `key` property.

For example, `ctrl+alt+e` has `ê` as `key` value and not `e`. Our key comparison was failing.

Another issue with foreign keyboard layouts is with digit. If user defines a `cmd+1` shortcut, he must hit `cmd+shift+1` or `cmd+numpad1`.

This PR is deeply inspired by [Mousetrap](https://github.com/ccampbell/mousetrap) which works really great!   

Bonus: new keys are now compatible for keymaps definition:
```
'backspace',
'tab',
'enter',
'capslock',
'esc',
'pageup',
'pagedown',
'end',
'home',
'ins',
'del'
```

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -138,6 +138,9 @@ export default class Term extends PureComponent {
   }
 
   keyboardHandler(e) {
+    if (e.type !== 'keydown') {
+      return true;
+    }
     // test key from keymaps before moving forward with actions
     const key = returnKey(e);
     if (key) {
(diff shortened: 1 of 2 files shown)


lib/components/term.js
```diff
@@ -138,6 +138,9 @@ export default class Term extends PureComponent {
   }
 
   keyboardHandler(e) {
+    if (e.type !== 'keydown') {
```
albinekb: this looks scary 😄  Is this `keyboardHandler` only for `keydown` ?

PR #2332, opened 2017-10-05
Add activity glow on tab title

chabou:
Each time that a tab have new activity, a glow effect is played:
hyper-activity_glow

Glow effect is throttled to 1 per second max

### DIFF
--- a/lib/components/tab.js
+++ b/lib/components/tab.js
@@ -1,21 +1,33 @@
 import React from 'react';
-import {PureComponent} from '../base-components';
+import {Component} from '../base-components';
 
-export default class Tab extends PureComponent {
+export default class Tab extends Component {
   constructor() {
     super();
 
     this.handleHover = this.handleHover.bind(this);
     this.handleBlur = this.handleBlur.bind(this);
     this.handleClick = this.handleClick.bind(this);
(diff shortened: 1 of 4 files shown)


chabou:
Fixes #536

chabou:
here a new try with smaller and slower glow animation: 
hyper-activity_glow

chabou:
thank you for your feedback @CodeTheory

rauchg:
One way to do a less-obtrusive glow might be to just animate `opacity`. Bring it down to 0.8, then quickly back up to 1.

chabou:
Here is glow effect with opacity:

hyper-activity_glow

Stanzilla:
I personally think the glow looks better and is "more hyper" but the opacity version is fine as well.

albinekb:
Maybe use the "facebook shimmer" effect? https://codepen.io/redouglas/pen/gobsm 🙉

jbergstroem:
Talk about distracting 😢 -- ever run a big rsync in another tab? That tab would be blinking constantly. I would prefer if this wasn't default, but easily themeable -- its just too opinionated.

chabou:
@timothyis What should we do?

Stanzilla:
Should be an option, imo.

PR #2337, opened 2017-10-09
[WIP] Hyper CLI

### LINKED ISSUE #2288: Hyper CLI
As started in https://github.com/zeit/hyper/pull/282, we will in 2.0 have a way to manage hyper via the CLI.

chabou:
Will fix #2288

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -63,6 +63,7 @@ const config = require('./config');
 config.setup();
 
 const plugins = require('./plugins');
+const {addSymlink, addBinToUserPath} = require('./utils/cli-install');
 
 const Window = require('./ui/window');
 
@@ -97,6 +98,13 @@ if (isDev) {
 } else {
   //eslint-disable-next-line no-console
   console.log('running in prod mode');
+  if (process.platform === 'win32') {
+    //eslint-disable-next-line no-console
+    addBinToUserPath().catch(err => console.error('Failed to add Hyper CLI path to user PATH', err));
+  } else {
+    //eslint-disable-next-line no-console
+    addSymlink().catch(err => console.error('Failed to symlink Hyper CLI', err));
+  }
 }
 
 const url = 'file://' + resolve(isDev ? __dirname : app.getAppPath(), 'index.html');
--- a/app/package.json
+++ b/app/package.json
@@ -27,7 +27,6 @@
     "queue": "4.4.0",
     "semver": "5.4.1",
     "shell-env": "0.3.0",
-    "util.promisify": "false1.0.0",
     "uuid": "3.1.0",
     "winreg": "1.2.4"
   }
--- a/app/package.json
+++ b/app/package.json
@@ -27,6 +27,7 @@
     "queue": "4.4.0",
     "semver": "5.4.1",
     "shell-env": "0.3.0",
+    "util.promisify": "false1.0.0",
     "uuid": "3.1.0",
     "winreg": "1.2.4"
   }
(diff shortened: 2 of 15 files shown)


chabou:
* Hyper CLI javascript file is bootstrapped by scripts (located in `build/(mac|win|linux)/`)
* Right platform script(s) are copied at build time to `bin/`
* At Hyper startup time (only in production mode), platform script is symlinked to `/usr/local/bin` (or `/usr/bin/` on Linux)

For now, only macOS has been tested and it works great:
* Build Hyper `yarn dist -- -m` 
* Launch `./dist/mac/Hyper.app`
* Type `hyper <args>`
* <args> are printed

app/utils/cli-install.js
```diff
@@ -0,0 +1,50 @@
...
+
+const lstat = pify(fs.lstat);
+const readlink = pify(fs.readlink);
+const unlink = pify(fs.unlink);
+const symlink = pify(fs.symlink);
+
+const target = process.platform === 'darwin' ? '/usr/local/bin/hyper' : '/usr/bin/hyper';
+const source = cliScriptPath;
+
+const checkInstall = () => {
```
albinekb: I would write all these promise-chain-functions without the return statement (implicit return) saves a few lines and also makes the code more readable, but if it's like this in the readme, keep it. 👍  (very very minor thing 😂 )

app/index.js
```diff
@@ -97,6 +98,13 @@ if (isDev) {
 } else {
   //eslint-disable-next-line no-console
   console.log('running in prod mode');
+  if (process.platform === 'win32') {
+    //eslint-disable-next-line no-console
+    addBinToUserPath().catch(err => console.error('Failed to add Hyper CLI path to user PATH', err));
```
albinekb: Should this be shown to the user (notification) or not?

app/index.js
```diff
@@ -97,6 +98,13 @@ if (isDev) {
...
+  } else {
+    //eslint-disable-next-line no-console
+    addSymlink().catch(err => console.error('Failed to symlink Hyper CLI', err));
```
albinekb: Same with this one, shown to the user or just a log?

app/package.json
```diff
@@ -27,6 +27,7 @@
     "queue": "4.4.0",
     "semver": "5.4.1",
     "shell-env": "0.3.0",
+    "util.promisify": "false1.0.0",
```
albinekb: this is built in: https://nodejs.org/api/util.html#util_util_promisify_original

app/utils/cli-install.js
chabou: I'll change that

app/index.js
chabou: Not enough place to print error in notification.
On macOS, error could be read in Console.app. But I don't know for other platform

app/package.json
chabou: I forgot to remove it from this file.
I replaced `util.promisify` by `pify`.
Promisify built-in is only available from node v8.x and we are using v7.9.0 (with Electron v1.7.8).

app/index.js
albinekb: Let's keep it as it is and see if it's needed later on 👍  Let's not over engineer it

app/package.json
albinekb: 👍  No probs

app/utils/cli-install.js
chabou: I really prefer a first explicit return in these functions

chabou:
I'm closing this PR.
As @albinekb asked, I'm creating a feature branch with multiple PR to make review easier.

Issue #2318, opened 2017-10-02
Right clicking the mouse in Hyper v2.+ doesn't paste the clipboard contents 🙃 


spences10:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 10 Pro v1703
- **Hyper.app version**: @latest
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://github.com/spences10/dotfiles/blob/master/.hyper.js
- **Relevant information from devtools** _(CMD+ALT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: N/A
- **The issue is reproducible in vanilla Hyper.app**: yes

## Issue

Right clicking the mouse in Hyper v2.+ doesn't paste the clipboard contents 🙃 

I have downgraded to Hyper v1.+ now

Thanks for all the awesome work 👍

bcorcoran:
I know it would take some getting use to, but you should be able to press CTRL+SHIFT+C/V to  copy/paste in Hyper on windows. CTRL+C is reserved for interrupt and [CTRL+V](https://en.wikipedia.org/wiki/Control-V) is not related to pasting at all.

spences10:
@bcorcoran but, but, it works in v1.4.8 why no workee in v2.+ 🙃

chabou:
Fixed buy #2392

Issue #2260, opened 2017-09-21
[v2] copyOnSelect and quickEdit not working with v2


mojoaxel:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

* **OS version and name**: ubuntu linux 16.04 64x
* **Hyper.app version**: canary branch ([tag 2.0.2](0502bbfc79d5421e0a6758ea671b7bb19d76c1cb)) - both custom build and release.
* **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/mojoaxel/a55766b1b6efc1f1802d1e43dab96919

## Issue
* [ ] **copyOnSelect** is not working with the current v2.0.2. Selecting a text is not copying it to the clipboard.
* [ ] **quickEdit** is not working with the current v2.0.2. It looks like right-click events are not handled at all.

nacimgoura:
Same error.

albinekb:
@nacimgoura, please use the reactions that github has introduced some years ago 👌 
![image](https://user-images.githubusercontent.com/5027156/30742801-2c6fe8ca-9f9b-11e7-8e16-ee4873484af1.png)

luukverhoeven:
Same error here

albinekb:
@luukverhoeven, please use the reactions that github has introduced some years ago 👌 
![image](https://user-images.githubusercontent.com/5027156/30742801-2c6fe8ca-9f9b-11e7-8e16-ee4873484af1.png)

nacimgoura:
Why was the issue closed?
I still have the problem on version 2.0.4 with ubuntu 16.04.

PR #2348, opened 2017-10-11
Added quickEdit support for canary

AndyBitz:
resolves issue #2318, #2260

Pastes the clipboard contents to the terminal on right click.

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -1,7 +1,7 @@
 /* global Blob,URL,requestAnimationFrame */
 import React from 'react';
 import Terminal from 'xterm';
-import {clipboard} from 'electron'
+import {clipboard} from 'electron';
 import {PureComponent} from '../base-components';
 import terms from '../terms';
 import returnKey from '../utils/keymaps';
@@ -80,7 +80,7 @@ export default class Term extends PureComponent {
 
     if (process.platform === 'win32') {
       // xterm does not support the mousedown event
-      window.addEventListener('mousedown', this.mouseHandler)
+      window.addEventListener('mousedown', this.mouseHandler);
     }
 
     window.addEventListener('resize', this.onWindowResize, {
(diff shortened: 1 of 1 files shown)


lib/components/term.js
```diff
@@ -24,6 +25,7 @@ export default class Term extends PureComponent {
     this.onWindowResize = this.onWindowResize.bind(this);
     this.onTermRef = this.onTermRef.bind(this);
     this.onTermWrapperRef = this.onTermWrapperRef.bind(this);
+    this.mouseHandler = this.mouseHandler.bind(this);
```
albinekb: This should only be bound on windows 🙌

AndyBitz: Listen and bind only on windows 🤙

lib/components/term.js
```diff
@@ -195,6 +221,10 @@ export default class Term extends PureComponent {
     // to do in case of splitting, see `componentDidMount`
     this.term._events = {};
 
+    if (process.platform === 'win32') {
+      window.removeEventListener('mousedown', this.mouseHandler);
```
albinekb: If you check for `this.mouseHandler` instead of platform, this code is future proof if someone in the future wanted to make this an option on all platforms 👍 


```js
if (this.mouseHandler) {
  window.removeEventListener('mousedown', this.mouseHandler)
}
```
🙌

AndyBitz: But wouldn't this always resolve to true?
Even if `this.mouseHandler` is unbound it's still a function of the class. 🤔

albinekb: Oh 🤦‍♂️ you're totally right, there's some more adjustments needed, but that's . probably overkill.

But should this really be default? Should it not be an opt-in setting? 🤔   (im not using windows so i don't know)

AndyBitz: I went through #2001 and #1247 and it seems like that the `quickEdit` option takes care of that.
Since it's default on windows I thought so. 🙆‍♂️ 
Maybe the `process.platform === 'win32'` should be replace with `this.props.quickEdit`?
Then it should work as it did in before v2.

AndyBitz: > `on right click selected text will be copied...`

I'll try to add this as well to make it complete. 🚀

albinekb:
Is this what all windows users expect when enabling `quickEdit`? Code looks 👍 now at least, but I can't say more than that. I don't use this function and I don't use windows :)

AndyBitz:
Yes, at least most of them, that's why it's default :)

PR #2364, opened 2017-10-16
Pasting a copied file or dir provides full path to resource

### LINKED ISSUE #1285: Cmd+v file path into hyper shell
- **OS version and name**: MacOS 10.12.2

## Request
Mac terminal allows the user to cmd+c a directory or file and cmd+v the path to said directory or file into a terminal shell. I use this feature extensively to quickly move between directories that I'm too lazy to get to via command line. It would be helpful if Hyper.app had this feature.

derrickpelletier:
closes #1285 

**ready for merge, tested on windows, linux, and mac**

~not ready for merge because: 
I've tested on linux, and mac, *but not windows* (posted in slack but didn't get a reply), so if someone can test that, that would be great. There is windows specific parsing in there that I haven't tested at all.~

Also, wasn't 100% sure if this is the appropriate spot for this code, but it seemed like it to me (first contrib to `hyper`). Kept clipboard processing in a util file to not clutter the `Term` component. If suggestions for a better place for this, let me know.

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -116,6 +116,8 @@ export default class Term extends PureComponent {
     this.fitResize();
   }
 
+  // intercepting paste event for any necessary processing of
+  // clipboard data, if result is falsy, paste event continues
   onWindowPaste(e) {
     if (!this.props.isTermActive) return;
 
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -117,8 +117,10 @@ export default class Term extends PureComponent {
   }
 
   onWindowPaste(e) {
+    if (!this.props.isTermActive) return;
+
     const processed = processClipboard();
-    if (processed && this.props.isTermActive) {
+    if (processed) {
       e.preventDefault();
       e.stopPropagation();
       this.term.send(processed);
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -5,6 +5,7 @@ import {PureComponent} from '../base-components';
 import terms from '../terms';
 import returnKey from '../utils/keymaps';
 import CommandRegistry from '../command-registry';
+import processClipboard from '../utils/paste';
 
 // map old hterm constants to xterm.js
 const CURSOR_STYLES = {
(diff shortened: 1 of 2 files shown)


derrickpelletier:
Alright, managed to get this tested on windows. Is ready for review and merge.

lib/components/term.js
```diff
@@ -110,6 +116,17 @@ export default class Term extends PureComponent {
     this.fitResize();
   }
 
+  onWindowPaste(e) {
+    if (!this.props.isTermActive) return;
+
+    const processed = processClipboard();
```
albinekb: I'd like a comment to why this code is here to intercept the event, so it's easy to know why this is needed without following the code all the way

lib/utils/paste.js
```diff
@@ -0,0 +1,21 @@
+import {clipboard} from 'electron';
+
+const getPath = platform => {
+  switch (platform) {
+    case 'darwin': {
+      const filepath = clipboard.read('public.file-url');
```
albinekb: does this return an empty string if you have something else than a file URL in the clipboard? 🤔  
that is, will this code only intercept pasted file urls and nothing else? (not messing with pasted commands for example)

derrickpelletier: Yeah, in my tests it returned an empty string. So an empty string would return out of the `processClipboard` method, where term would evaluate to false (no processing happened) and continue with paste. 

I could only see it messing with some non-file paste if some application writes to those specific formats during a copy?

lib/components/term.js
derrickpelletier: Yep will do

albinekb: Perfect 👍

derrickpelletier:
Excellent!

jq-87:
Any idea when this will make it to stable?

chabou:
Very soon. I'm pretty confident that our `2.0.0-canary.14` was our last 2.0.0 canary :tada:

PR #2367, opened 2017-10-17
Replace nodejs version 8.4.0 with 8 in .travis.yml

PeterDaveHello:
It's better to specify only 8 in Travis CI so that we don't need to manually update its version especially when there is any security update.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ language: node_js
 matrix:
   include:
     - os: linux
-      node_js: 8.4.0
+      node_js: 8
       env: CC=clang CXX=clang++ npm_config_clang=1
       compiler: clang
 


albinekb:
The reason why we have locked it is because we run almost all code with "electron's node". t
That is, the node that's running in electron. And we lock the travis version to the current version shipped with the version of electron we run.

Better would be if we could fix our tests to run via electron-shipped-node instead of travis node.

example would be:

we had electron 1.x which shipped with node 7.0, but if we then had node 7 in travis, and travis selected node 7.4+, async/await would be available natively within the tests, making them pass.

but in when the code is run by hyper, it will use electron-node (7.0 in this case) and crash

thoughts about this @PeterDaveHello ? 👍

albinekb:
Closing this, I think the solution is to write somthing to automatically manage node version based on electron-node version, see my comments in #2374

PeterDaveHello:
hmmm, cool

PR #2342, opened 2017-10-10
[Hyper-cli] Add cli cmd

chabou:
Replace #2337 

DO NOT MERGE IT. This feature was created to make review easier.
This PR is already merged in #2375 branch (the only one that should be merged).

* Hyper CLI javascript file is bootstrapped by scripts (located in `build/(mac|win|linux)/`)
* Right platform script(s) are copied at build time to `resources/bin/`
* When installing Hyper on Linux, platform script is symlinked to `/usr/local/bin`.
* At Hyper startup time (only in production mode), platform script is symlinked to `/usr/local/bin` on Mac, or `resources/bin` is added to user PATH env var on Windows.

PR #2375, opened 2017-10-18
[Hyper-cli] embed hpm

chabou:
This PR has PR #2342 already merged into it.

Next steps:
* [X] Prevent `Hyper CLI` to create a new instance of `Hyper` if it is already launched. It should act the same way that launching Hyper twice by icon.
  * [X] macOS (new tab)
  * [X] Windows (new window)
  * [X] Linux (new window)
* [X] Test it on Linux
* [X] Test it on Windows
  * [X] Cmd
  * [X] PowerShell
  * [X] GitBash/CygWin
  * [X] WSL ⚠️  we can only launch `Hyper` and not install/uninstall/list plugins due to Env var problem: see  https://github.com/Microsoft/WSL/issues/1494

Bonus:
* [X]  `hyper <path>` should open Hyper in this path
  * [X] macOS
  * [X] Windows
  * [X] Linux
  * [X] Check if path exists

Screencast: 
hyper-cli


chabou:
There is an error on Linux
```
Failed to symlink Hyper CLI { Error: EACCES: permission denied, unlink '/usr/local/bin/hyper'
  errno: -13,
  code: 'EACCES',
  syscall: 'unlink',
  path: '/usr/local/bin/hyper' }
```

```
parallels@ubuntu:~/hyper$ ls -lah /usr/local/bin/
total 8.0K
drwxr-xr-x  2 root root 4.0K Oct 23 00:02 .
drwxr-xr-x 10 root root 4.0K Jul 19  2016 ..
lrwxrwxrwx  1 root root   16 Oct 23 00:02 hyper -> /opt/Hyper/hyper
```

chabou:
One cool side effect is that Hyper CLI is our `open <path> in new tab` feature (on macOS):
hyper-cli-openinnewtab


But the more important thing is: THIS PR IS READY TO MERGE 🎉 🎉  🎉

peterfraedrich:
I've submitted a PR for `hpm` that deprecates the use of `npm-name` and adds authenticated proxy support. 
[Link To PR](https://github.com/zeit/hpm/pull/100)

Stanzilla:
I can't add reviewers but I nominate @albinekb :)

htkoca:
Instead of going to the path in the first argument, can it be a command? hyper `cd /path` rather than just `/path` . This will allow `cd` to handle path existence (which it is designed to do), and also allow other applications to run things in hyper (such as an SFTP program opening a SSH connection in hyper)

cli/index.js
```diff
@@ -0,0 +1,204 @@
+// This is a CLI tool, using console is OK
+/* eslint no-console: 0 */
+const {spawn, exec} = require('child_process');
+const {isAbsolute, resolve} = require('path');
+const {existsSync} = require('fs');
+const pify = require('pify');
+const args = require('args');
+const chalk = require('chalk');
+const opn = require('opn');
+// const columnify = require('columnify');
```
albinekb: remove

cli/index.js
```diff
@@ -0,0 +1,204 @@
...
+// const ora = require('ora');
+const api = require('./api');
+
+let commandPromise;
+
+const checkConfig = () => {
+  if (api.exists()) {
+    return true;
+  }
+  let msg = chalk.red(`Error! Config file not found: ${api.configPath}\n`);
```
albinekb:
```js

Comment (albinekb) [inline cli/index.js]:
```const

Comment (albinekb) [inline cli/index.js]:
```

cli/index.js
```diff
@@ -0,0 +1,204 @@
...
+  msg += 'Please launch Hyper and retry.';
```
albinekb:
```js
const message = `
${chalk.red(`Error! Config file not found: ${api.configPath}`)}
Please launch Hyper and retry.
`
```

app/utils/cli-install.js
```diff
@@ -0,0 +1,86 @@
...
+exports.addBinToUserPath = () => {
+  // Can't use pify because of param order of Registry.values callback
+  return new Promise((resolve, reject) => {
+    const envKey = new Registry({hive: 'HKCU', key: '\\Environment'});
+    envKey.values((err, items) => {
+      if (err) {
+        reject(err);
+        return;
+      }
+      // C:\Users\<user>\AppData\Local\hyper\app-2.0.4\resources\bin
```
albinekb: `// C:\Users\<user>\AppData\Local\hyper\app-<version>\resources\bin`

cli/api.js
```diff
@@ -0,0 +1,112 @@
+const fs = require('fs');
+const os = require('os');
+const npmName = () => Promise.resolve(); //require('npm-name');
```
albinekb: 🤔

cli/index.js
```diff
@@ -0,0 +1,204 @@
...
+
+  if (plugins) {
+    console.log(plugins);
+  } else {
+    console.log(chalk.red(`No plugins installed yet.`));
+  }
+  process.exit(1);
+});
+/*
+const lsRemote = () => {
```
albinekb: Remove?

cli/index.js
albinekb: Should we add an issue to make the cli ask to create the file if it does not exists?

cli/index.js
chabou: It will be reactivated when NPM API request (URL and criteria) will be chosen.

app/utils/cli-install.js
chabou: I'll fix this

cli/index.js
chabou: It will be used by lsRemote

cli/index.js
chabou: Yes. 
Config creation should remain an `Hyper` task. And yes, we should improve this automatic creation. But this is not a blocker. Most of case, config file will be present.

albinekb: 👌

cli/api.js
chabou: I hoped that this PR would have been merged sooner: https://github.com/dominictarr/rc/pull/94

chabou:
@htkoca thank you for your feedback. You're right. Passing command could be very usefull.

Imo, I prefer command as an option like your `--run` suggestion. Most of time, users will only want to open `Hyper` on a (default) location. I don't want to oblige them to add `cd` to their path.

@albinekb @rauchg @leo  Thoughts ?

cli/index.js
chabou: fixed

app/utils/cli-install.js
chabou: fixed

cli/api.js
chabou: fixed

cli/index.js
chabou: fixed

chabou:
For `install`, `unsinstall`, `docs` commands, `hyper-` prefix is added if not present.
For `search` and `ls-remote` commands, results are filtered to show only `hyper-` prefixed packages.

build/win/hyper
```diff
@@ -5,6 +5,9 @@ NAME="Hyper"
 HYPER_PATH="$(dirname "$(dirname "$(dirname "$(realpath "$0")")")")"
 ELECTRON="$HYPER_PATH/$NAME.exe"
 if grep -q Microsoft /proc/version; then
+	echo "Warning! Due to WSL limitations, you can use CLI commands here. Please use Hyper CLI on cmd, PowerShell or GitBash/CygWin."
```
amckinlay: What does this mean? Do you mean "you cannot use Hyper CLI here"?

amckinlay:
@chabou I left a comment about the language of the warning message in `build/win/hyper`.

As an aside, I wonder why my comment was on an outdated commit.

build/win/hyper
Stanzilla: yeah just missing the `'t`

chabou:
Yes this is a typo.
Can you send a PR or open an issue if you can't?
Thank you 🙏

PR #2382, opened 2017-10-21
Remember the window size on close and launch accordingly

KunalKene1797:
Implement Electron Window State Manager

This will remember the last known position and window size of the app and relaunch it accordingly.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -55,6 +55,7 @@ const {resolve} = require('path');
 const {app, BrowserWindow, Menu} = require('electron');
 const {gitDescribe} = require('git-describe');
 const isDev = require('electron-is-dev');
+const WindowStateManager = require('electron-window-state-manager');
 
 const AppMenu = require('./menus/menu');
 const config = require('./config');
@@ -67,6 +68,11 @@ const plugins = require('./plugins');
 const Window = require('./ui/window');
 
 const windowSet = new Set([]);
+
+const mainWindowState = new WindowStateManager('hwin', {
+  defaultWidth: 1024,
+  defaultHeight: 768
+});
 
 // expose to plugins
 app.config = config;
@@ -146,12 +152,17 @@ app.on('ready', () =>
           }
         }
 
+       
         const hwin = new Window({width, height, x: startX, y: startY}, cfg, fn);
         windowSet.add(hwin);
+        if (mainWindowState.maximized) {
+          hwin.maximize();
+        }
         hwin.loadURL(url);
 
         // the window can be closed by the browser process itself
         hwin.on('close', () => {
+          mainWindowState.saveState(hwin);
           hwin.clean();
           windowSet.delete(hwin);
         });
--- a/app/package.json
+++ b/app/package.json
@@ -17,6 +17,7 @@
     "electron-config": "1.0.0",
     "electron-is-dev": "0.3.0",
     "electron-squirrel-startup": "1.0.0",
+    "electron-window-state-manager": "false0.3.2",
     "file-uri-to-path": "1.0.0",
     "git-describe": "4.0.2",
     "lodash": "4.17.4",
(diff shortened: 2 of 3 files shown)


KunalKene1797:
Travis CI error Need Help?
The command "npm install -g yarn" failed and exited with 1 during .

Stanzilla:
Nice work!

app/index.js
```diff
@@ -55,6 +55,7 @@ const {resolve} = require('path');
 const {app, BrowserWindow, Menu} = require('electron');
 const {gitDescribe} = require('git-describe');
 const isDev = require('electron-is-dev');
+const WindowStateManager = require('electron-window-state-manager');
```
albinekb: Is this only used to save and restore if the window was maximized....?

app/index.js
```diff
@@ -148,10 +155,14 @@ app.on('ready', () =>
 
         const hwin = new Window({width, height, x: startX, y: startY}, cfg, fn);
         windowSet.add(hwin);
+        if (mainWindowState.maximized) {
```
albinekb: because here is the only place you read something from it?

app/index.js
albinekb: If yes, why can't this be in a plugin like https://www.npmjs.com/package/hyperfullscreen or https://github.com/simonmeusel/hyperfull 🤔

app/index.js
KunalKene1797: hwin on (close) is the place where it reads the data of the current window on line 165.
Then it reads the data when next time the app is launched!

KunalKene1797: the one u mentioned just checks the maximized state of the window and sets the window state to either maximized or restored.

app/index.js
KunalKene1797: the ones u mentioned here are to make it full screen. whereas this code will make it dynamic depending on the clients need, whether he wants the app to be full screen partially filled or any custom dimension that he wants. This will store the data from the previous interaction with hyper and apply the same to a new instance when created.

app/index.js
KunalKene1797: usage of this small package overrides the default width and height parameters which were being set earlier.

app/index.js
```diff
@@ -148,10 +155,14 @@ app.on('ready', () =>
...
+          hwin.maximize();
```
KunalKene1797: The maximize/restore button state remains the same, though everything else works fine. the maximize button state is defined in header.js. Could use some help over here fixing the maximize button not showing the maximized window state.

```
const maxButtonHref = this.props.maximized
      ? './renderer/assets/icons.svg#restore-window'
      : './renderer/assets/icons.svg#maximize-window';

```
```
handleMaximizeClick() {
    if (this.props.maximized) {
      this.props.unmaximize();
    } else {
      this.props.maximize();
    }
  }
```

ppot:
@albinekb  Hyper already remember window size and location from the last opened window on restart. This should be a plugin if window was maximized.

Stanzilla:
eh, that is some pretty basic functionality and the behaviour is probably expected to be like that by most people. should be in core imo.

KunalKene1797:
@ppot which operating system are u running? @Stanzilla I agree!

ppot:
@KunalKene1797 
https://github.com/zeit/hyper/blob/canary/app/config/windows.js
https://github.com/zeit/hyper/blob/canary/app/ui/window.js#L242
https://github.com/zeit/hyper/blob/canary/app/index.js#L112

KunalKene1797:
@ppot saw that code it still doesn't work! please try executing it on Windows both that code and this PR.

KunalKene1797:
@ppot @albinekb its still broken on Windows on version 2.0.4, if the window is fullscreen and it is closed then relaunched, it goes out of the screen a little bit and is not maximized at all! :p

Stanzilla:
So we talked about it and is there a way you could do this without adding a dependency and maybe keep in mind that it has to work with session restoration in the future?

KunalKene1797:
Alright, I'll give that a look!

Stanzilla:
I actually think Electron Window State Manager is the best way.

Stanzilla:
Rebased here https://github.com/zeit/hyper/pull/2914

PR #2390, opened 2017-10-24
copyOnSelect support in canary

derrickpelletier:
PR is ready to be merged: tested on mac, ubuntu, and windows 10

fixes #2260 (partially)

Fairly straightforward: on mouseUp, if there is a selection and copyOnSelect is true, it'll copy it.

----------

I had also added quickEdit support in before I noticed there was an existing PR. I backed it out though, see commit here: https://github.com/derrickpelletier/hyper/commit/9dd1d27a96c5236ba3a21971186d29a08c0a9778#diff-268710614c2274f611be667a5f72bb13R141

I'll add it back in if we're going to merge it but @albinekb seemed unsure of that functionality at this time. If you'd rather maintain the existing PR for it (#2348), note that it uses a different method for click handling that I did here.

PR #2392, opened 2017-10-25
quickEdit support

### LINKED ISSUE #2260: [v2] copyOnSelect and quickEdit not working with v2

derrickpelletier:
tested on mac, ubuntu 16.04, windows 10

closes #2260, #2318 

Note that this should close #2348, which is the same functionality, very slightly different implementation. I didn't intend to duplicate their effort, but @chabou asked me to submit this (in https://github.com/zeit/hyper/pull/2390#issuecomment-339148926)

As far as I know, and based on a simple test in windows command prompt, this works as expected:
- Select text
- right click anywhere to copy the selection
- selection clears
- right click anywhere to paste

If there are any nuances to quick-edit beyond that, I'm not aware of them, but even still this is probably enough for a first iteration at the very least.

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -137,6 +137,7 @@ export default class Term extends PureComponent {
       clipboard.writeText(this.term.getSelection());
     }
 
+    // handle quickEdit copy/paste with right click
     if (this.props.quickEdit && e.button === 2) {
       if (this.term.hasSelection()) {
         clipboard.writeText(this.term.getSelection());
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -132,12 +132,6 @@ export default class Term extends PureComponent {
   }
 
   onMouseUp(e) {
-    // handle copying selected text
-    if (this.props.copyOnSelect && this.term.hasSelection()) {
-      clipboard.writeText(this.term.getSelection());
-    }
-
-    // handle quickEdit copy/paste with right click
     if (this.props.quickEdit && e.button === 2) {
       if (this.term.hasSelection()) {
         clipboard.writeText(this.term.getSelection());
@@ -145,6 +139,8 @@ export default class Term extends PureComponent {
       } else {
         document.execCommand('paste');
       }
+    } else if (this.props.copyOnSelect && this.term.hasSelection()) {
+      clipboard.writeText(this.term.getSelection());
     }
   }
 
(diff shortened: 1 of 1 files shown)


lib/components/term.js
```diff
@@ -131,11 +131,21 @@ export default class Term extends PureComponent {
...
+  onMouseUp(e) {
     // handle copying selected text
     if (this.props.copyOnSelect && this.term.hasSelection()) {
       clipboard.writeText(this.term.getSelection());
     }
+
+    // handle quickEdit copy/paste with right click
+    if (this.props.quickEdit && e.button === 2) {
+      if (this.term.hasSelection()) {
+        clipboard.writeText(this.term.getSelection());
```
albinekb: Won't this be run twice? if you have both `quickEdit` and `copyOnSelect` activated?
can't we just change the if here https://github.com/zeit/hyper/pull/2392/files#diff-268710614c2274f611be667a5f72bb13R136

```
if ((this.props.copyOnSelect || this.props.quickEdit && e.button === 2) && this.term.hasSelection()) {
```
Otherwise it will be copied twice which could be a problem
🤔

derrickpelletier: Ah yeah, if you select text _and then_ right click, the right click would copy the same text twice. It still works but you're right, unnecessary.

That condition you suggested isn't enough, because in the quickEdit case, it needs to clear the selection after copying. I modified it to just do one or the other, thoughts?

albinekb: Looks good to me! @derrickpelletier 👍

Stanzilla:
Tested and works well!

chabou:
🙏  Thank you @derrickpelletier

Issue #2395, opened 2017-10-25
Plugin Error


rao-abdul-mannan:
I'm on mac 10.12.6 and with basic hyper installation, i'm getting following error sometimes whenever i close a split.

Also I get notifications  sometimes when i drag the split borders  and with each drag the count increases in that notification showing how much I dragged  I guess. 

![2017-10-25_18-59-32](https://user-images.githubusercontent.com/8604174/32002863-eda322ac-b9b6-11e7-8719-511f6a6191bb.png)

        -
        Hyper 2.0.4
        Electron 1.7.8
        darwin x64 16.7.0

Stanzilla:
Did you try turning off your plugins to find out which one it is?

rao-abdul-mannan:
It is random. I couldn't reproduce it but i'll check it again.

KevinRLeC:
I've run into this issue as well. Turned off all plugins and this error still persisted

Console error: 

```
TypeError: this.term.measureResize is not a function
    at TermGroup_.componentWillReceiveProps (bundle.js:5)
    at updateClassInstance (bundle.js:4)
    at beginWork (bundle.js:4)
    at e (bundle.js:4)
    at k (bundle.js:4)
    at p (bundle.js:4)
    at y (bundle.js:4)
    at D (bundle.js:4)
    at Object.enqueueSetState (bundle.js:4)
    at Connect.v.setState (bundle.js:4)
```

chabou:
Fixed by #2396

PR #2396, opened 2017-10-25
Removing call to undefined function on resize

KevinRLeC:
To Address #2395 

Looks like the `measureResize` function was removed in [2273](https://github.com/zeit/hyper/pull/2273) but this file was never updated. This removes the error.


albinekb:
Yep, looks like `measureResize` is gone 👍 

image
Tested, no errors 👍

chabou:
Nice catch 💯 
@KevinRLeC Thank you for your help 🙏

rauchg:
My bad 🙈  Thansks @KevinRLeC

PR #2422, opened 2017-11-03
Unify wording and spelling in config and other places

Stanzilla:

### DIFF
--- a/app/config/config-default.js
+++ b/app/config/config-default.js
@@ -20,7 +20,7 @@ module.exports = {
     // `'BEAM'` for |, `'UNDERLINE'` for _, `'BLOCK'` for █
     cursorShape: 'BLOCK',
 
-    // set to `true` for blinking cursor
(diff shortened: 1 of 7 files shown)


app/config/config-default.js
```diff
@@ -32,23 +32,23 @@ module.exports = {
...
 
-    // custom css to embed in the main window
+    // custom CSS to embed in the main window
     css: '',
 
-    // custom css to embed in the terminal window
+    // custom CSS to embed in the terminal window
     termCSS: '',
 
-    // set to `true` (without backticks and without quotes) if you're using a
```
chabou: Please keep this. We had so much user errors like:
`showHamburgerMenu: 'true',`

Stanzilla: Done and added to a few more places as well so it is consistent.

chabou:
Thank you @Stanzilla !!

PR #2425, opened 2017-11-04
Improved context menu

leo:
I changed the order of a few context menu entries to make more sense.

As an example, "horizontal splitting" has to come before "vertical splitting" because "x" always comes before the "y" axis. In addition, I replaced "Close Session" with "Close Tab" since we're saying "New Tab" and not "New Session" above.

### DIFF
--- a/app/menus/menus/shell.js
+++ b/app/menus/menus/shell.js
@@ -39,7 +39,7 @@ module.exports = (commandKeys, execCommand) => {
         type: 'separator'
       },
       {
-        label: 'Close Pane',
+        label: 'Close',
         accelerator: commandKeys['pane:close'],
         click(item, focusedWindow) {
           execCommand('pane:close', focusedWindow);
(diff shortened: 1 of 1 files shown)


app/menus/menus/shell.js
```diff
@@ -5,41 +5,41 @@ module.exports = (commandKeys, execCommand) => {
...
       },
       {
         type: 'separator'
       },
       {
-        label: 'Close Session',
+        label: 'Close Tab',
         accelerator: commandKeys['pane:close'],
         click(item, focusedWindow) {
           execCommand('pane:close', focusedWindow);
```
chabou: You should modify this command if you change its name

chabou:
What do you think to call this item `Close` like iTerm2

image

Its suit well if there is only one session and it closes app directly (on Windows/Linux) and if there is multiple panes in tab and it closes only one pane.

leo:
@chabou You're right. Done!

PR #2436, opened 2017-11-06
Fix selectAll command

chabou:
I'm not totally satisfied by this implementation but it works


lib/containers/hyper.js
```diff
@@ -66,6 +74,7 @@ class Hyper extends PureComponent {
 
   componentDidMount() {
     this.attachKeyListeners();
+    window.rpc.on('term selectall', this.handleSelectAll);
```
albinekb: Why lowercased here and camelCased in the other place (selectAll)?

PR #2440, opened 2017-11-06
Enhance plugin installation

chabou:
This PR enhance plugin installation:
* No more double error notification triggered when plugin is added to config before launching Hyper
* Error notification message is better when yarn failed:
  * Before : `Command failed: /Users/chabou/Documents/Projets/hyper/node_modules/electron/dist/Electron.app/Contents/MacOS/Electron /Users/chabou/Documents/Projets/hyper/bin/yarn-standalone.js install --no-emoji --no-lockfile --cache-folder /Users/chabou/.hyper_plugins/cache
error Couldn't find package "lklfdklfkl" on the "npm" registry.`
  * After: `error Couldn't find package "lklfdklfkl" on the "npm" registry.`
* Delay before launching plugin installation is reduced from 5s to 1s

### DIFF
--- a/app/plugins.js
+++ b/app/plugins.js
@@ -57,8 +57,7 @@ function updatePlugins({force = false} = {}) {
 
     if (err) {
       //eslint-disable-next-line no-console
-      console.error(err.stack);
-      notify('Error updating plugins.', err.message);
+      notify('Error updating plugins.', err);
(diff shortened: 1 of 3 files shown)


Stanzilla:
Very nice, that was always super confusing to users.

PR #2452, opened 2017-11-11
npm start script runs in single terminal

theRemix:
- ready to merge
- to test: `yarn` then `yarn start`
- results: runs `yarn run dev` and `yarn run app` in a single terminal
- logs are easy to read
- exits nicely
- does not inhibit the previous ability to run in two separate terminals

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "repository": "zeit/hyper",
   "scripts": {
-    "start": "echo 'please run `yarn run dev` in one tab and then `yarn run app` in another one'",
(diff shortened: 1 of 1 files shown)


chabou:
I am not totally satisfied with this solution for multiple reasons.
The major one is:
* Launch with Hyper with `yarn start`
* Stop hyper
* Make a modification in code source
* Launch with Hyper with `yarn start`
* Previous builded code is used. Dev must make a ForceReload to see his modifications.

A minor one:
* Remove previous build `rm -Rf app/renderer`
* Launch with Hyper with `yarn start`
* Electron app hangs because renderer build is missing
* You need to force a reload or open devtools to make Hyper work


We can't launch these task at the same time, because Electron app launching should always occur after renderer build.

PR #2459, opened 2017-11-15
Fixing #1869. Ensuring that the restored window position is valid

### LINKED ISSUE #1869: config.json does not update Window Position after 2nd display is unplugged
Dell XPS-9560 (3840x2160) connected via HDMI to a 27" Dell monitor running at 1920x1080. I had Hyper term window on the 27" Dell monitor. Finished for the day and shutdown all applications including Hyper. Shutdown down laptop. Disconnect HDMI.

Later in the evening at home, turned on laptop. Started VS Code, Chrome and Postman (all apps that were running on the external monitor earlier), they all came up fine with the window sizes a bit wonkey due to the resolution switch.

However the Hyper windows was not visible even though the program was running. I looked at config.json and the Window Position was set to [4000, 50]. Obviously outside of the laptop's visible area. Closed down hyper and manually changed to [0,0]. 

Not sure if this is a bug. Unable to replicate as I was working at that remote location just for today.

-

Hyper 1.3.3
Electron 1.4.16
win32 x64 10.0.15063

Xapphire13:
Fixes #1869 

I have also added unit tests for the change. And removed unit tests for files that no longer exist (does this imply UTs aren't run for this project?).


chabou:
Thank you so much for this PR.
Indeed UTs are not systematically run for this project. And this is bad.
I have broken these UTs without being warned about this.

We need to include them in `yarn test`. Not only linting.
Do you mind to fix this test command in your PR?

Otherwise, I'll make another PR.

Xapphire13:
I shall modify the test command =]. Once this is in, do I need to cherry pick it to master?

Xapphire13:
Ping

chabou:
We definitely need to run UTs on push.
It will not prevent commiting early and often. (commit !== push). And if you need to push with broken UTs, you still can do : `yarn push --no-verify` (but this is bad 😜 ), but CI tests will fail.

Can you make UTs run on prepush?
Thanks !

Xapphire13:
Certainly! I'll get around to it either tomorrow or later this week =]

Xapphire13:
Actually, turns out I can edit on my phone =]. Done!

chabou:
I will test & merge ASAP

app/config/windows.js
```diff
@@ -8,11 +9,28 @@ const cfg = new Config({
   }
 });
 
+function validateAndFixWindowPosition(position, size) {
+  const displays = electron.screen.getAllDisplays();
+  const [x, y] = position;
+  const positionIsValid = displays.some(({workArea}) => {
```
chabou: We can't test with any of available displays, only with targeted screen.
Screen targeting and some verification are done here: https://github.com/zeit/hyper/blob/5d0c6355e12ec8d9a89182556fa73227374b488f/app/index.js#L113-L148

It is a little bit messy and we probably should refactor it to add your position/size validation.

app/config/windows.js
```diff
@@ -8,11 +9,28 @@ const cfg = new Config({
...
+    return x >= workArea.x && x <= workArea.x + workArea.width && y >= workArea.y && y <= workArea.y + workArea.height;
```
chabou: We should test if size is valid too. We don't want to have a valid position and an almost invisible window. For example if `x === workArea.width - 1` or `y === workArea.height - 1`.

app/config/windows.js
```diff
@@ -8,11 +9,28 @@ const cfg = new Config({
...
+  });
+
+  if (!positionIsValid) {
+    const {workArea} = electron.screen.getPrimaryDisplay();
+    position[0] = workArea.x + (workArea.width - size[0]) / 2;
+    position[1] = workArea.y + (workArea.height - size[1]) / 2;
+    cfg.set('windowPosition', position);
```
chabou: We don't need to calculate a valid position. I think that fallback to default position/size is ok.

app/config/windows.js
Xapphire13: >We can't test with any of available displays, only with targeted screen.

I'm having trouble understanding what you mean by this?

>It is a little bit messy and we probably should refactor it to add your position/size validation.

Sounds reasonable. Would you have the logic persist a fixed location? Or just assume that a correct location will be written on next close, and if not, the verification logic will take care of it again.

app/config/windows.js
Xapphire13: I had contemplated this, but decided to keep it simple and follow YAGNI. If we were to get into this case, the use would have had to positioned the window there themselves and if it's on the screen the standard <kbd>ctrl</kbd> + <kbd>arrow</kbd> shortcuts would work.

If we go with this method we would need to define how big a work area is permissible. We could perhaps change the logic to check the center of the window, if the center of a terminal window is in the workarea, then we're good, if not, relocate? (This would cover both the edge case you mention, and the original "my monitor is no longer plugged in" case)

app/config/windows.js
Xapphire13: The defaults are private to the [windows config](https://github.com/zeit/hyper/blob/5d0c6355e12ec8d9a89182556fa73227374b488f/app/config/windows.js#L4-L9) module. If I refactor my change to be in `index.js` how would I go about invalidating the set values?

Xapphire13:
@chabou 

I finally got around to refactoring this and addressing your comments. Could you please take a look?

Xapphire13:
Ping @chabou

cnt000:
Only for me this bug is a block? I can't use hyper.
There are issues or we will have possibilities to acept this? Gently @chabou :-D

Xapphire13:
@albinekb, could we get a second reviewer on this please?

chabou:
I am really sorry for the delay.
I will review/merge it before next week.

app/config/windows.js
chabou: My first point: I misunderstood that a windows referential is always the same. I was thinking that we specified a screen when opening a new window. I was wrong.

My second point: I would assume that a correct location will be written or fixed if needed. But never mind, let's merge this as it is.

app/config/windows.js
chabou: I think that considering window center is really better than top-left corner.
I know there is some shortcuts (different on each OS), but I am pretty sure that some users know nothing about them.

I will fix this by myself, because this PR had a too long wait time. Sorry again for this

app/config/windows.js
chabou: Thank you to have fixed this

app/config/windows.js
chabou: No need to consider new window center because startX/startY is already corrected by existing code.

chabou:
Thanks a lot for your help 🙏

Xapphire13:
Thanks so much @chabou!

Issue #2451, opened 2017-11-11
Background opacity (with rgba) not working on Ubuntu 17.10


Jeffarese:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [X] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Ubuntu 17.10
- **Hyper.app version**: 2.1.1
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/Jeffarese/65bd9e3d48f2f4bfaf4453f82d246d4e
- **Relevant information from devtools** _(CMD+ALT+I on macOS, CTRL+SHIFT+I elsewhere)_: N/A
- **The issue is reproducible in vanilla Hyper.app**: Is Vanilla

## Issue

I'm trying to lower the opacity of the background color, using `backgroundColor: 'rgba(0,0,0, 0.5)'` but it's not working, the background has 100% opacity no matter what.

If I change to another color (not caring about the opacity) it works, but it ignores the opacity.

SergioGeeK7:
opacity is not supported by electron on linux/windows. just in mac

albinekb:
We should probably put a notification when a user tries to get transparency in windows/ubuntu and link to the comment by electron.

chabou:
Adding a comment in config file should be ok

Jeffarese:
Yeah, I didin't know this. With a comment in the config file it would have been enough.

PR #2463, opened 2017-11-16
Making it explicit that opacity is only supported in macOS  

AndreFCruz:
PR is ready to be merged.
Makes it more explicit that background colour opacity is not supported in windows/linux.
Related to #2451

### DIFF
--- a/app/config/config-default.js
+++ b/app/config/config-default.js
@@ -27,6 +27,7 @@ module.exports = {
     foregroundColor: '#fff',
 
     // terminal background color
+    // opacity is only supported on macOS
     backgroundColor: '#000',
 
     // border color (window, tabs)


app/config/config-default.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
     foregroundColor: '#fff',
 
     // terminal background color
+    // opacity is only supported on macOS
```
chabou: Maybe `transparency` is better. Or `Changing opacity`...

albinekb: I think this is good, it's called opacity everywhere i css.

chabou:
Thank you @AndreFCruz for this first contribution! 🙏 ❤️

PR #2467, opened 2017-11-17
Fixing zoom out keymap

baukevanderlaan:
I think I fixed the broken keymap for zoom:out.

I found out that when I replaced command+minus with command+- the keyboard shortcut works again on a Mac. I'm not sure why this suddenly stopped working because I couldn't find a recent commit that changed this keymap – maybe it's an Electron thing? Anyway, it makes sense, since that particular key produces a hyphen-minus and not a minus.

I can't test on Windows or Linux but I took the liberty to change those keymaps as well. Can somebody test them?


chabou:
Thanks a lot ! 🙏

PR #2477, opened 2017-11-21
Prompt before closing the app when a process is running (Fixes #684)

### LINKED ISSUE #684: Prompt before closing the app when a process is running
Related to #623. Would it be possible to have a prompt before quitting the app when a process is running? It could help to prevent quitting the app with an ongoing process that we won't be able to kill without manually killing the PID.

antonioalmeida:
Currently works on macOS fully, and on linux when closing tabs. Doesn't work on linux when closing windows directly because we're overriding a different mechanism. Could use some guidance to know what events to prevent for that case.
On Windows it doesn't' work at all because the process we're running to check if the processes is unix based. Runs without any issues though.

Could use some feedback to improve the way we're obtaining the current sessions PID


Stanzilla:
Nice, that is definitely a good feature to be in the core!

antonioalmeida:
Update on this: now works on Linux when closing tabs individually or the entire window

diogotorres97:
@albinekb what do you think about this fix/feature?

lib/utils/dialogs.js
```diff
@@ -0,0 +1,23 @@
+const {dialog} = require('electron').remote;
```
albinekb: this file is exactly the same as `app/utils/dialogs.js` or? 🤔

antonioalmeida: Kind of, yeah. We did that because the closure is suspended in different processes when we're a closing a single tab vs closing all the windows at once (one is on electron's main process, the other on the renderer process). The was the simplest way we found to go around this (different files with different electron includes), but there's definitely a better way, sure.

diogotorres97: @albinekb @chabou What is the current situation?

albinekb: Interesting @antonioalmeida, do you think we should add a comment to the top of both files to describe that (why there's two copies of the file) so that the next person coming along have a chance to understand what's going on 🤔 💭  😄

antonioalmeida: Sure thing 👍

chabou:
You could take a look here: https://github.com/weinand/vscode-processes/blob/master/src/ps.ts
Good exemple to list process on linux/mac but on Windows too

Stanzilla:
@antonioalmeida willing to pick this up again?

antonioalmeida:
It's been a couple of years, I'm not sure if what I made is still valid

Stanzilla:
Alright, I'll close this now then, but if you feel this is still needed, please make a new PR :)

PR #2482, opened 2017-11-22
add additional sentence To run hyper

blackhands77:
5. To run hyper
yarn run app from another terminal tab/window/pane
→ `yarn run app` from another terminal tab/window/pane **OR press ctrl+c then `yarn run app`**
--------------------------------------------------
Hi, I'm a beginner to use github.
When I install this project I found another path to run hyper so I request it.
And I can't find any contribution guide.
So I'm very sorry to this untyped explanation.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -50,7 +50,7 @@ Regardless of the platform you are working on, you will need to have Yarn instal
 3. Install the dependencies: `yarn`
 4. Build the code and watch for changes: `yarn run dev`
 5. To run `hyper`
-  * `yarn run app` from another terminal tab/window/pane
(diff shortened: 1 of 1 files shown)


chabou:
If you interrupt `yarn run dev`, you'll need to relaunch it each time you want to test something.
It is really better to keep it launched in a tab, because webpack will watch changes and will rebuild renderer code when needed (and only what have changed). You'll just have to relaunch electron by using `yarn run app` or VSCode launch task.

Thank you really much for your contriubtion but this PR will not be accepted. I am really sorry.

If you want to describe what I said about renderer code watched/rebuilt and why you need to have 2 tabs (or 1 tab + VSCode), I will happy to merge it.

Issue #2468, opened 2017-11-18
Text cut off at bottom when scrollable


levrik:
- [ ] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 10 1709
- **Hyper.app version**: 2.1.1
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/levrik/ba854eeab882a9759f7f50b768ab4f6b
- **Relevant information from devtools** _(CMD+ALT+I on macOS, CTRL+SHIFT+I elsewhere)_: _N/A_
- **The issue is reproducible in vanilla Hyper.app**: yes

## Issue
![image](https://user-images.githubusercontent.com/9491603/32980014-971e3e6a-cc5f-11e7-8e85-81e7c02a4b36.png)

Does only happen when the scrollbar appears. I'm not able to scroll down further.
Makes it especially difficult with underscores in the last line since you cannot see them.

**Edit:** Starting from `canary.9` the whole last line is cut-off.

Magneticmagnum:
Happens to me as well.  No plugins.
Sometimes, the entire bottom row doesn't scroll down, nor am I able to scroll further (Not able to see what I'm typing in the terminal)

Nufflee:
Same here, when I run a command that fills up the whole screen, the text gets cut off - it doesn't autoscroll nor I can scroll myself. It's incredibly annoying.

ineeve:
I can not replicate this issue. Can you give me any instructions on how to replicate it ?

bennycooly:
I also have this issue. I'm on Windows running bash (WSL). Whenever the output of a command exceeds the bottom of the terminal, the text gets cut off and I can't even scroll or see the output. The only way to fix this is to issue a clear command.

ineeve:
Can this be related to [https://github.com/xtermjs/xterm.js/issues/1136](https://github.com/xtermjs/xterm.js/issues/1136)?

Ehekatl:
the fix for xtermjs is already merged https://github.com/xtermjs/xterm.js/pull/1157
any updates here?

chabou:
the fix for xtermjs is included in our `2.0.0-canary.9`.

Maybe using xterm fit addon could help with this issue: https://github.com/zeit/hyper/pull/2594

[...]

PR #2492, opened 2017-11-23
Subtracted 1 from the number of rows, to prevent text from being cut …

ineeve:
…off at bottom when scrollable. Solves issue #2468

PR is ready to be merged. I think this is the correct solution for the problem related to the issue #2468.
I've tested with powershell and it is working as it should.


ineeve:
This might be only a temporary fix, as it is probably related with [this xterm issue](https://github.com/xtermjs/xterm.js/issues/1136)

Stanzilla:
Hey there, please try again with the new Hyper v2 release we picked up the xterm version that fixes this.

PR #2497, opened 2017-11-27
Use same `hazel` endpoint to notify update to Linux users 

### LINKED ISSUE #2476: Prevent canary update notification for stable Linux users.
## Context
On Linux, we don't use autoUpdater (and `hazel`) because it is not compatible.
Instead, we are making some request to a news feed `https://hyper-news.now.sh` with some params:
https://github.com/zeit/hyper/blob/5d0c6355e12ec8d9a89182556fa73227374b488f/app/notifications.js#L19-L24

## Problem
`updateChannel`  is not given as parameter. That's why stable users currently see notification about v2.1.1 release 😞 

## Solution
We should use this news feed only for important messages and not as an autoUpdater fallback.
Update notifications for linux should use same feed (`hazel`) than autoUpdater and notify users if update is available without downloading archive.

chabou:
Fixes #2476

This PR implements a pseudo `auto-updater-linux`. It fetches the same URL and trigger same events. All data related to the available release (name, note, date etc...), are passed to `update-available` event instead of `update-downloaded`.
Internal updater use Electron autoUpdater or this one on Linux.

### DIFF
--- a/app/updater.js
+++ b/app/updater.js
@@ -1,5 +1,6 @@
 // Packages
-const {autoUpdater, app} = require('electron');
+const electron = require('electron');
+const {app} = electron;
 const ms = require('ms');
 const retry = require('async-retry');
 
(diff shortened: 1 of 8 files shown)


app/updater.js
```diff
@@ -1,23 +1,27 @@
...
 
 // Utilities
 // eslint-disable-next-line no-unused-vars
 const notify = require('./notify');
 const {version} = require('./package');
-const {getConfig} = require('./config');
+const {getDecoratedConfig} = require('./plugins');
 
 const {platform} = process;
+const isLinux = true; //platform === 'linux';
```
albinekb: This should be fixed? 🤔

chabou: indeed...... Sorry for that

PR #2499, opened 2017-11-28
Fix branch detection when setting the canary icon

matheuss:
When a tag is being built, `TRAVIS_BRANCH` equals to the tag name. Therefore, the canary icon was never being copied over 😄 

This also copies the Windows (.ico) icon over 👌

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,9 +33,10 @@ install:
   - yarn
 
 after_success:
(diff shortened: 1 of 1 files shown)


leo:
@matheuss Awesome! Can you fix the tests. Also, is it still [WIP]? If not, please remove that prefix from the title. 🚀

chabou:
CircleCI was not failing because of tests. I relaunched build and everything is ok.

leo:
Ok, wonderful!

PR #2503, opened 2017-11-29
Keep Arabic letters without wrapping them with span

MortadaAK:
Fix the Arabic letter connection where splitting the chars into multiple spans will cause the words to be broken.

Ex. currently when you write مرحبا it will display it as اب ح ر م.

### DIFF
--- a/lib/hterm.js
+++ b/lib/hterm.js
@@ -31,7 +31,7 @@ hterm.Terminal.prototype.onMouse_ = function (e) {
 };
 
 function containsNonLatinCodepoints(s) {
-  return /[^\u0000-\u00ff]/.test(s);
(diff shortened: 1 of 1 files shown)


chabou:
`hterm` will be replaced by `xtermjs` in our v2 already available on our canary update channel.

Can you try this v2 and confirm your issue is fixed?

MortadaAK:
it has the same issue

chabou:
We are using xermV3 that use canvas and not DOM anymore.
Can your try with our latest canary (>= canary.9)?

MortadaAK:
Sorry for the late response but still it has same issue

Stanzilla:
Please open an issue on https://github.com/xtermjs/xterm.js, the PR can't be merged since we no longer use hterm.

PR #2507, opened 2017-11-30
request

chillininvt:

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -26,6 +26,7 @@
     "node-pty": "0.7.3",
     "queue": "4.4.2",
     "semver": "5.4.1",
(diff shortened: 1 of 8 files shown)


Stanzilla:
A little explanation would be nice

chabou:
This is obivously a mistake :)
BTW nice idea to use Hyper with serial port and not a pty.

Do no hesitate to poke us when your fork is finalized.

albinekb:
Very interesting to use serial port indeed!   👌 We should take this into mind in the future of hyper, if we want to embed hyper into stuff like Atom and websites, we should be "backend" agnostic, so having a plugin that talks over Serial, SSH etc would be really cool.

PR #2510, opened 2017-11-30
Save and restore window size after restart (Fixes #2249)

### LINKED ISSUE #2249: Save and restore window size after restart
- **OS version and name**: Windows 10
- **Hyper.app version**: 1.4.7
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/someanon/e45a11f297f38882cc3eac5c9053d2af
- **Relevant information from devtools** _(CMD+ALT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: N/A
- **The issue is reproducible in vanilla Hyper.app**: resize to needed size, maximize, close hyper, open hyper

## Issue
If hyper closed in maximized state after restart it doesn't restore the common size. Istead it restores maximized size (as it was common) but not maximized state and shifted bottom-right a little from top-left screen corner.

I think hyper should start maximized state if it closed so. And it should save and restore common size after maximized close, so if i minimize maximized window after restart it would become common size before restart.

Maybe it would be better to make optional parameter in .hyper.js for window size of any new hyper windows and option for start maximized if it closed so before.

Carolina-Rocha:
This commit ensures that hyper is restarted in a maximized state if it is closed in a maximized state. Minimizing a maximized window after restart gives a window with default dimensions.

Fixes: [#2249](https://github.com/zeit/hyper/issues/2249)
This issue is only for windows

### DIFF
--- a/app/config/config-default.js
+++ b/app/config/config-default.js
@@ -54,7 +54,7 @@ module.exports = {
 
     // default window width
     windowWidth: 500,
-    
+
     // default window heigth
     windowHeight: 370,
 
--- a/app/config/config-default.js
+++ b/app/config/config-default.js
@@ -52,6 +52,12 @@ module.exports = {
     // custom padding (CSS format, i.e.: `top right bottom left`)
     padding: '12px 14px',
 
+    // default window width
+    windowWidth: 500,
+    
+    // default window heigth
+    windowHeight: 370,
+
     // the full list. if you're going to provide the full color palette,
     // including the 6 x 6 color cubes and the grayscale map, just provide
     // an array here instead of a color map object
--- a/app/config/import.js
+++ b/app/config/import.js
@@ -60,3 +60,15 @@ exports.getDefaultConfig = () => {
   }
   return defaultConfig;
 };
+
+exports.updateState = function(path, data) {
+  _write(path, data);
+};
+
+exports.importState = function(path) {
+  try {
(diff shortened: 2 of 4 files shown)


app/config/import.js
```diff
@@ -60,3 +60,15 @@ exports.getDefaultConfig = () => {
...
+  _write(path, data);
+};
+
+exports.importState = function(path) {
+  try {
+    return readFileSync(path, 'utf8');
+  } catch (err) {
+    return 0;
+  }  
+};
```
albinekb: Add an empty line after this one, all files should end with a blank line to make git diffs better 🙂

app/config/config-default.js
```diff
@@ -52,6 +52,12 @@ module.exports = {
     // custom padding (CSS format, i.e.: `top right bottom left`)
     padding: '12px 14px',
 
+    // default window width
+    windowWidth: 500,
```
albinekb: maybe `defaultWindowSize: {width:500, height: 370}` to make it clear that it's default without needing the comment. I don't know what's best 😄 @chabou ?

Fleshgrinder:
No progress for a long time here. Just gave Hyper another look and this was instantly the first thing that caught my eye. Any chance to get this merged? Can I help?

dimuls:
Any progress? How can i build hyper with this patch?

Issue #1325, opened 2016-12-31
Auto-update if all windows closed


leo:
IMHO, we should just restart automatically after downloading an update if no Hyper window is open.

rauchg:
This is really clever

albinekb:
I wouldn't expect that to happen...

pablopunk:
From the [electron auto-updater docs](https://github.com/electron/electron/blob/master/docs/api/auto-updater.md):

> Note: autoUpdater.quitAndInstall() will close all application windows first and only emit before-quit event on app after that. This is different from the normal quit event sequence.

If there is no window open, shouldn't `quitAndInstall()` just restart it? I mean, I would assume that that's the expected behavior, isn't it working like that right now?

ppot:
I prefer the macOS way.  Give user the ability to update when they want.

rauchg:
I like the Chrome way. Always run the _latest_, the least you bother the user the better.

pablopunk:
> I like the Chrome way. Always run the latest, the least you bother the user the better.

Thumbs up for that 🙌🏻

IMO for patches or even minor versions it should be automatic.

rauchg:
Yep, in the case of no terminal sessions being open.

AlbertoElias:
That would be great. How does the updating process work currently? Is it manual? I'm on Windows 10

jhonnymoreira:
@rauchg is this issue valid yet? I started using Hyper a few days ago and I would like to contribute.

leo:
Yep, I'd say we're happy to accept a PR for it! 😊

jhonnymoreira:
thanks, @leo! Expect a PR soon. [=

Stanzilla:
https://github.com/zeit/hyper/pull/2515

urmila1907:
It would be great to be given a notice before restarting, even if no hyper window is open.

jumpman-cmd:
It’s great to see the progress on this issue! I agree that automating the update process when no windows are open would streamline things for users. Implementing this feature could enhance the user experience by ensuring they always have the latest version without manually updating. I’m curious if there are any plans to include user notifications or options for handling updates in different scenarios, such as when windows are open. I'm looking forward to seeing this feature in action!

PR #2515, opened 2017-12-03
Auto-update if all windows closed  (#1325)

3rdvision:
This commit enables hyper to install already downloaded new updates when all windows are closed.

It comes as a solution for #1325.


Stanzilla:
This should probably be a setting?

Issue #2319, opened 2017-10-02
Suggestion: Password input icon


lucasqueiroz:
## Suggestion

In the latest version of iTerm, they have included an icon that is a replacement for the current cursor icon when you have to type a password, so it looks like a key. I think this is very interesting.

iTerm:
![screen shot 2017-10-02 at 13 07 13](https://user-images.githubusercontent.com/29466590/31087231-a6dc0d94-a772-11e7-9fb7-52cf96571814.png)

MAPESO:
Thank you for your suggestion.

jbergstroem:
FWIW, this has been in `terminal.app` for a release or three as well. I prefer the implementation too:

<img width="573" alt="screen shot 2017-10-10 at 1 57 43 pm" src="https://user-images.githubusercontent.com/176984/31399546-0d09284e-adc3-11e7-95f9-cca579bb3121.png">

PR #2527, opened 2017-12-07
Added Password Icon when Password is requested

bernardoleite:
Hello!
Through this Pull Request I implemented a solution to the suggestion presented in "Suggestion: Password input icon #2319". 
Fixes: #2319 

Hope you enjoy!

### DIFF
--- a/app/config/import.js
+++ b/app/config/import.js
@@ -4,6 +4,7 @@ const {defaultCfg, cfgPath, plugs, defaultPlatformKeyPath} = require('./paths');
 const {_init, _extractDefault} = require('./init');
 
 let defaultConfig;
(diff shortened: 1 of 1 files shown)


bgw:
I think it would be better to detect a password prompt by examining the TTY state, [similar to how iTerm2 does it](https://github.com/gnachman/iTerm2/blob/c71a6a682b4f1eaf52682037dbba0d9beb37eb5e/sources/PTYTask.m#L426). It might be difficult to do this, depending on the hooks that xterm.js provides.

If you want to use this sudo-specific implementation, I think it would be better to encourage people to put it in their shell's configuration (zshrc/bashrc).

Issue #2243, opened 2017-09-20
Error while installing a plugin


omarmhaimdat:
<img width="532" alt="screen shot 2017-09-20 at 11 27 33 pm" src="https://user-images.githubusercontent.com/2160742/30671365-6e5e5ffc-9e5e-11e7-9caf-f5a410b12542.png">
        

## How I got this problem:

I installed hyper a couple of days ago, installed some themes, changed the shell, and then worked with it without any bug or glitch and not even some lag. Worked just perfectly!

And then figured I could add some cool plugins, so I added some, and then I guess I restarted hyper because of an update, and then suddenly every time I installed a new plugin (via hpm), I got the pop-up box (ScreenShot above).

- OS Version 10.12.6
- MacBook Pro 2.3 GHz Intel Core I7 | 16GB
- Hyper 1.4.7
- Electron 1.7.6
- darwin x64 16.7.0

**Link of a Gist with the contents of the .hyper.js:** https://gist.github.com/omarmhaimdat/0aa32ff5f9455303a0b1f1eb79cacc77

MAPESO:
@albinekb  @CodeTheory  **hpm** will have to investigate because it is causing errors in stable and canary versions. I'll see if I can correct this problem.

MAPESO:
@omarmhaimdat  If you are going to customize Hyper I recommend you do it through the **hyper.js** file 😄

Stanzilla:
Hey there, please try again with the new Hyper v2 release and open a new issue if the problem stills exists. Thank you!

omarmhaimdat:
I think the problem has been solved, Good Job \

--------------
Omar MHAIMDAT

PR #2530, opened 2017-12-08
Fix reading config file concurrency

NadiaCarvalho:
This PR is supposed to correct the issue:
#2243

We discovered it was a concurrency problem and the plugins were giving this error not on the first installation but on the second or third because they were reading the 'not yet written' configuration by the last installed plugin.
It was added a flag to check if the configuration is being modified. If it is, it should bypass new changes until the configuration is completely read and the changes are absolutely done.

The Pull Request is able to be merged.

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -8,6 +8,7 @@ const {cfgPath, cfgDir} = require('./config/paths');
 const watchers = [];
 let cfg = {};
 let _watcher;
+let _isWatching = false;
 
 const _watch = function() {
   if (_watcher) {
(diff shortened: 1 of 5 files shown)


app/config.js
```diff
@@ -8,17 +8,22 @@ const {cfgPath, cfgDir} = require('./config/paths');
 const watchers = [];
 let cfg = {};
 let _watcher;
+let _isWatching = false;
```
chabou: `isImporting` seems to be a better var name

app/config.js
```diff
@@ -8,17 +8,22 @@ const {cfgPath, cfgDir} = require('./config/paths');
...
   if (_watcher) {
     return _watcher;
   }
 
   const onChange = () => {
-    cfg = _import();
-    notify('Configuration updated', 'Hyper configuration reloaded!');
-    watchers.forEach(fn => fn());
-    checkDeprecatedConfig();
+    if (_isWatching == false) {
```
chabou: You should use `===` 

But more important, I think that `_isWatching` will ever be `true` because all next actions are synchronous.
Can you explain how it can be false ?

Issue #2432, opened 2017-11-06
Rebinding the 'break' command from ctrl+c?


isseiler:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 10 Pro 
- **Hyper.app version**: 1.4.8 stable
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/isseiler/0fa72e513af1852b7df180e62a19e071 

## Issue

Ask: How can I get the ctrl+c 'Break' behavior to be bound to ctrl+x or ctrl+d? 

I've rebound ctrl+c to be Copy so I need a replacement for the break action. I've scoured the internet and Hyper's docs looking for this keybind but I can't find it. Is it not possible?

I love Hyper and I nearly have a perfect in environment but this is a bit of a hold up :/

This is the most up to date key map I can find: https://github.com/zeit/hyper/blob/master/app/keymaps/win32.json

ineeve:
I've managed to solve this issue and I'll submit the pull request soon.

albinekb:
Thanks to @ineeve this will be available in the next `canary` release, read more on how to try the canary releases at https://zeit.co/blog/canary. We're working hard on finishing up hyper 2.0, but until that's ready you can get it via the canary releases.

🎉

PR #2531, opened 2017-12-08
Added keymap to rebind the 'break' command - Issue #2432

ineeve:
Created a keymap to rebind the break/interrupt command as asked [here](https://github.com/zeit/hyper/issues/2432)
I have only tested on Windows 10, but I think it should also work on Linux and MacOS.
Any feedback is appreciated.


albinekb:
Doesnt this break `ctrl+c` for copy in windows and linux? @ineeve @chabou

PR #2543, opened 2017-12-11
Use public API to remove xterm listeners

Tyriar:
This uses the public API to do the same thing. Note that I didn't verify this but I think it should work.


chabou:
Thank you so much for your help @Tyriar

PR #2593, opened 2018-01-12
Make CLI perform static analysis of config lazily

bgw:
zeit/hyper#2375 adds a wrapper CLI tool that parses the ~/.hyper.js configuration file. However, the static analysis logic can't possibly handle every possible configuration file without executing it.

In my case, [my configuration file][0] defines a constant at the top, which means that the `module.exports` definition isn't the first statement. This breaks the parsing logic.

However, this static analysis is only necessary if one wishes to use the CLI subcommands. They're not needed if one just wants to launch hyper.

Unfortunately, the existing implementation tries to read from the AST immediately when `cli/api.js` is imported. This changes it to be lazy, and uses memoization retain the same reference, allowing the AST to still be mutated in-place.

This PR should be ready to merge. I've tested it to the best of my ability by building a new version with `yarn run dist`, modifying my configuration file, and seeing that a few `hyper` subcommands seem to work.

[0]: https://github.com/bgw/dotfiles/blob/665e83fd7a13be675ad0ae984478d5b0de502ee1/hyper.js#L4

### DIFF
--- a/cli/api.js
+++ b/cli/api.js
@@ -6,37 +6,51 @@ const recast = require('recast');
 
 const fileName = `${os.homedir()}/.hyper.js`;
 
-let fileContents;
-let parsedFile;
-let plugins;
-let localPlugins;
-
-try {
-  fileContents = fs.readFileSync(fileName, 'utf8');
-
-  parsedFile = recast.parse(fileContents);
-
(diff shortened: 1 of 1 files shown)


chabou:
Amazing work!

Thank you so much @bgw for your help 🙏

PR #2594, opened 2018-01-12
Use xterm.js's fit addon when calling fitResize

bgw:
xterm.js doesn't use `this.term.charMeasure.width` when rendering. It
actually uses:

```
Math.floor(this._terminal.charMeasure.width * window.devicePixelRatio)
```

This is important, because character widths may be non-integer values,
which means that we can end up with an empty space on the right side of
the terminal window.

Instead of trying to match xterm's behavior, it's probably better if we
just use xterm's `fit` addon. This seems to behave better for me.

We might not want to land this as-is, becuase this addon forces an
annoying hard-coded 17px margin on the right side to compensate for a
scrollbar (see xtermjs/xterm.js#400), but at least for my use-cases,
this is still better than it was.

Before:
workspace 2_020

After:
workspace 2_021

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -1,11 +1,14 @@
 /* global Blob,URL,requestAnimationFrame */
 import React from 'react';
 import {Terminal} from 'xterm';
+import * as fit from 'xterm/lib/addons/fit/fit';
 import {clipboard} from 'electron';
 import * as Color from 'color';
 import {PureComponent} from '../base-components';
 import terms from '../terms';
 import processClipboard from '../utils/paste';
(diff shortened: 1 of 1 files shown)


lib/components/term.js
```diff
@@ -195,10 +198,7 @@ export default class Term extends PureComponent {
     if (!this.termWrapperRef) {
```
chabou: Is it necessary to keep this term DOMNode reference?

chabou:
Thank you so much !
This 17px margin is annoying, but this better than now.

Issue #2579, opened 2018-01-10
Separate a style sheet from index.html


ParkSB:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate
## Issue

There is a internal style sheet using ```<style>``` tag in the website code([website/index.html](https://github.com/zeit/hyper/blob/master/website/index.html)) now, so the code is too long and difficult to read. I think it would be better to create a external style sheet.

ParkSB:
PR #2598

PR #2598, opened 2018-01-12
Add external style sheet for the website (#2579)

ParkSB:
Issue #2579 
I removed the internal style sheet from index.html, and added external style sheet(style.css).


chabou:
Good catch! thank you

Issue #2607, opened 2018-01-14
Default --supports-dual-gpus=true (Hyper disabling iGPU (OSX))


ethanx94:
- [X] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [X] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OSX 10.13.2**: 
- **Hyper.app version 1.10.16**: 
- **.hyper.js (default config)**: 
- **Relevant information from devtools: N/A** _(CMD+ALT+I on macOS, CTRL+SHIFT+I elsewhere)_: 
- **Is Vanilla**: 

## Issue
I noticed that Hyper, as well as many other Chrome/Electron based apps, were actively disabling the iGPU in my MacBook. I am running an eGPU which would normally not be an issue but the entire laptop display goes black and on the external display performance is brought to a crawl. Therefore I'm proposing a supports-dual-gpu switch to be added to the main index.js that seems to remedy this problem.
This may also alleviate another issue with artifacting on a vanilla MacBook Pro that has discrete graphics without a performance hit. (Although I have not been able to reproduce that issue to test this theory)
See this potentially related issue here: #747

Let me know your thoughts, if adding a switch into the main app/index.js is an issue maybe we can support adding a .hyper.js configuration variable to append commandLine switches instead.

Greenek:
This is already fixed in v2.0.0-canary and I believe old versions won't be supported anymore.

ethanx94:
My mistake, I listed 1.10.16 but I actually cloned down the latest canary branch to make these modifications. Without that flag the app is completely unusable in an eGPU setup (A setup which will become more common soon as Apple has officially released a developer eGPU kit). If the change is too invasive or if it seems like it may cause unintended side effects maybe we could go with the second option of a config variable.

ethanx94:
....On further inspection, although I feel like this is a Chrome issue. I was able to remedy the problem by changing the order of my dGPU/eGPU set up. This fixed graphics switching as it now falls back to the factory AMD card. That being said others with MacBooks which lack a dGPU may still be affected by this.
If you know that Google has removed the --supports-dual-gpus flag from more recent versions of Chromium/Chrome feel free to disregard this pull. Although if I have time I might write up something to read in a config variable for appendSwitch arguments to be passed in to meet this edge case if that is something that might benefit the project.

mckernanin:
Still an issue on latest version

[...]

PR #2608, opened 2018-01-14
fixed issue with dual-gpus getting disabled

ethanx94:
#2607

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -70,6 +70,8 @@ const Window = require('./ui/window');
 
 const windowSet = new Set([]);
 
+app.commandLine.appendSwitch('supports-dual-gpus', 'true');
+
 // expose to plugins
 app.config = config;
 app.plugins = plugins;


albinekb:
If I understand this correctly, this flag will tell macOS that the app supports switching between GPUs? (iGPU & discrete card / eGPU) whereas without this flag, it will force the better GPU to be on all the time? 🤔 

What are the drawbacks of adding this flag? Are there issues with GPU switching and electron/chrome? If not, why isn't this default?

Sorry for all the questions but I want to make sure this doesn't negatively effect other users 👍 @ethanx94

ethanx94:
Drawbacks I can think of: None except I believe this is considered an experimental flag. I think that’s why it isn’t default. I haven’t throughly tested this on other platforms however.

Furthermore this doesn’t seem to be supported for me in newer versions of Chrome/Chromium. Meaning it doesn’t hurt anything, but just doesn’t do anything different. If Electron is ever updated to use a newer version of Chromium this argument may become useless again.

This has a lot to due with how Chrome is forcing graphics switching as soon as Hyper/VSCode/Any Electron app is opened. See my comments on this here: https://github.com/zeit/hyper/issues/2607#issuecomment-358201484

A more thorough look at what is going on would be to go to chrome://gpu with this flag on and look at the GL_RENDERER. I’m not sure how to get to here from the Electron app.
Someone with a deeper understanding of Chrome might be able to shed some light on this but all I can say with confidence is that I had an issue and this flag fixed it.

albinekb:
Thanks for that @ethanx94 👍 Yes, chrome and electron is a bit messy. I'm not sure what to do. @chabou thoughts on this one?

Maybe we should put this in a plugin and link people that want this fix to activate the plugin? It feels to risky to change it for everyone when we only ever got 1 report of this being an issue. 

VS Code, Slack and all the other Electron apps would also need to add this flag for it to be effective, since in most cases Hyper isn't the only electron app running on the system.

Thoughts? 🌡

saamalik:
The 'gfxCardStatus' app indicates which apps are using iGPU vs eGPU. Anyway, from my _limited_ testing, _Hyper_ continues to run in 'Integrated' GPU _even_ with this PR. Is there some special command needed to kick in the dedicated graphics card?

My computer: MacBook Pro 15in mid-2017.

PR #2612, opened 2018-01-15
Fix offscreen term size calculation

chabou:
`fit` xterm's addon, use parent size to calculate canvas sizes. But when term is offscreen (another tab), ancestor have a `display:none` and prevents xterm to calculate real size.

Using absolute positioning and offscreen positioning (if unfocused) rather than using `display:none` fixes this.

Thanks to IntersectionObserver, `xterm` pauses rendering when canvas are offscreen: https://github.com/xtermjs/xterm.js/pull/1144

### DIFF
--- a/lib/components/terms.js
+++ b/lib/components/terms.js
@@ -151,17 +151,16 @@ export default class Terms extends Component {
       },
 
       termGroup: {
-        opacity: 0,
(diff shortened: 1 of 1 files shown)


bgw:
Oops! Thanks for fixing this.

PR #2614, opened 2018-01-16
Fix initial cursor blinking

### LINKED ISSUE #2596: cursorBlink don't work on canary 9
- **OS version and name**: Windows 10
- **Hyper.app version**: Canary 9

## Issue
On the new canary 9 update the parameter cursorBlink does not work anymore. Even when true the cursor is static.

chabou:
This is a workaround for https://github.com/xtermjs/xterm.js/issues/1194

Fixes #2596


albinekb:
@chabou can you add a comment linking to the xterm issue?

PR #2618, opened 2018-01-17
Fix xterm fork version

chabou:
Using short commit sha was a bad idea. 😞


chabou:
Short commit was a good idea BUT...

I noticed that when you are using a github dependency, and you want to change pointing Branch/Commit/Tag, you MUST change version number of dependency in its package.json. Otherwise, previous fetched "package" can be use. I noticed this with `yarn`.

Ready to merge

PR #2619, opened 2018-01-17
Put offscreen terms far far away

chabou:
`top: 100%` is not enough when using `hyper-statusline`:
image

### DIFF
--- a/lib/components/terms.js
+++ b/lib/components/terms.js
@@ -155,12 +155,12 @@ export default class Terms extends Component {
         width: '100%',
         height: '100%',
         position: 'absolute',
-        top: '100%', // Offscreen to pause xterm rendering, thanks to IntersectionObserver
-        left: 0
+        top: 0,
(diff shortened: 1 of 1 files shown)


albinekb:
The workaround should be made in hyper-statusline, not in hyper imho.

chabou:
@albinekb we need to ensure that offscreen tabs will not appear, whatever plugins do.

Tyriar:
@chabou why isn't `display:none` being used?

chabou:
@Tyriar Related PR: https://github.com/zeit/hyper/pull/2612

With `addon`, unfocused terms are resized to 100x83. Because this plugin uses parent elt size: https://github.com/xtermjs/xterm.js/blob/0379ef7bc788d3ccca72c52ff9fb5ae158eee594/src/addons/fit/fit.ts#L21-L22

And when upper parent has `display: none` attribute, term parent have `width=100%` and `height=100%` (wrongly parseInt-ed to 100).

When Term become active, it is recalculated and works great on macOS but on Windows, history keep his 100x83 size.

Issue #2578, opened 2018-01-10
modifierKeys altIsMeta on Canary 9 build


saamalik:
- [x] I am on the latest Hyper.app version
- [x] I have searched the issues of this repo and believe that this is not a duplicate * (there were but closed now)

HI guys,
I'm on the latest `canary.9` build of hyper. Unfortunately, the `altIsMeta` modifier is not working as expected. Whenever the key combinations `alt+f` (move word forward), `alt+.` (last argument of the previous command) are pressed, some random silly looking special character is added. 

I'm on Mac OSX and trying to use the Option button as the _alt_ key. Here is the relevant section from the `.hyper.js`:

```
module.exports = {
  config: {
    modifierKeys: { altIsMeta: true },
...
}
```

saamalik:
Related bugs:
https://github.com/zeit/hyper/issues/330 (Closed now)
https://github.com/zeit/hyper/issues/2251

saamalik:
Can anyone else please confirm the same issue?

albinekb:
Hello @saamalik! This is probably due to the upgrade of the terminal rendering/handling lib `xterm` to the latest v3. I don't know exactly why this has stopped working but thanks for reporting!

saamalik:
Corresponding issue on underlying terminal: xtermjs/xterm.js#487

saamalik:
Found a temporary solution. Add the following to your `~/.inputrc` file:
```
"ƒ": forward-word
"∫": backward-word
"≥": yank-last-arg
```

And restart your terminal (or run `bind -f ~/.inputrc`). It's a temporary solution since still other bashrc shortcuts do not work (e.g: `alt+1`, `alt+2`, ..). Hopefully a more generic solution is found.

saamalik:
@albinekb proposed fix in underlying xterm.js library: [Mac OSX: altIsMeta support](https://github.com/xtermjs/xterm.js/pull/1225

albinekb:
A w e s o m e  work @saamalik 🙏 😍

PR #2623, opened 2018-01-20
Mac: altIsMeta support

saamalik:
* Reimplement altIsMeta supprt with new xterm library

_PR is ready to be merged _when_ underlying xterm.js is updated to this version: https://github.com/xtermjs/xterm.js/commit/616958e2c6c004133e4c5fe728702852bb1d2a5f. We just need to update the hyper xterm version in package.json._

Fixes: https://github.com/zeit/hyper/issues/2578

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ language: node_js
 matrix:
   include:
     - os: linux
-      node_js: 7.9.0
+      node_js: 8.0.0
       env: CC=clang CXX=clang++ npm_config_clang=1
       compiler: clang
 
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -21,6 +21,7 @@ const getTermOptions = props => {
   // Set a background color only if it is opaque
   const backgroundColor = Color(props.backgroundColor).alpha() < 1 ? 'transparent' : props.backgroundColor;
   return {
+    macOptionIsMeta: props.modifierKeys.altIsMeta,
     cursorStyle: CURSOR_STYLES[props.cursorShape],
     cursorBlink: props.cursorBlink,
     fontFamily: props.fontFamily,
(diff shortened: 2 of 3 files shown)


saamalik:
The new _xterm_ option `macOptionIsMeta` was recently added in this PR: https://github.com/xtermjs/xterm.js/pull/1225

chabou:
Amazing work @saamalik! You PR in `xterm` is awesome!

albinekb:
Thank you and again awesome work @saamalik ❤️

saamalik:
@albinekb my pleasure! Thank you for building and maintaining such an awesome terminal!

MindTooth:
Hi,

Not sure if this is working as intended.  At least I can't managed to make it work with WeeChat.
Expected behaviour is that I can Alt + Up/Down to change buffers.  It only acts as regular up and down arrow keys, switches between previous commands.  This works when using the Command key however.  But that doesn't work with other combos since the key is used for other things on macOS.

I've set `altIsMeta: true`.

It works as intended when using iTerm2.app,   In Terminal.app only Alt + Up/Down works.

Any thoughts?  Is there any bindings in hyper.js that takes precedence?

Ref.keys: https://weechat.org/files/doc/devel/weechat_user.en.html#key_bindings_buffers_windows

saamalik:
@MindTooth -- from the xterm.js code, I can see that `alt + up/down` is already supported even before this change. The changes introduced in the `xterm.js` PR were more for `alt + [a-z]` sequences. I'll try out the `alt + up/down` sequences later tonight to see if I can reproduce the issue.

Logan2211:
I'm seeing a similar issue as @MindTooth... meta seems to be working for everything except up/down/left/right in tmux for me. `altIsMeta: true` is set.

hozza:
@logan2211 same here `modifierKeys: { altIsMeta: true }` set inside `config` but not working with tmux over ssh.

PR #2631, opened 2018-01-23
Improve Windows default editor detection

brandon93s:
The current default editor check for Windows suffers from false-positives. The false-positives result in the config being opened with WScript and the user being prompted with an error:

image

The current implementation checks _all_ values in `...FileExts\\.js`. The updates in this PR filter down the values to just those inside of `...FileExts\\.js\\UserChoice` which is where the user's configured editor lives. `OpenWithList`, which powers the right click -> open with context menu, is one of the sub-keys. This can be populated with valid targets without the user having ever provided a default program:

image

The shell does not use values from this list. By including them in the check, the code can wrongfully assume a default editor has been provided and proceeded without the notepad fallback. 

**TL;DR** - The check is more restrictive now, and should more accurately fallback to notepad avoiding errors. 

---

Tested:
- Unconfigured (opens notepad)
- WScript explicitly set as default (opens notepad)
- VS Code as default (opens code)
- Atom as default (opens atom)

### DIFF
--- a/app/config/open.js
+++ b/app/config/open.js
@@ -3,61 +3,52 @@ const {cfgPath} = require('./paths');
 
 module.exports = () => Promise.resolve(shell.openItem(cfgPath));
 
+// Windows opens .js files with  WScript.exe by default
+// If the user hasn't set up an editor for .js files, we fallback to notepad.
 if (process.platform === 'win32') {
   const Registry = require('winreg');
   const {exec} = require('child_process');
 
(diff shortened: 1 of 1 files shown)


chabou:
Wow ! Awesome work @brandon93s! And very well documented PR 😍 

Thank you so much for your help! 🙏

brandon93s:
Awesome! Thank you both for the quick review. 👍

PR #2638, opened 2018-01-24
Display full download table on mobile

Angristan:
On mobile, the 64 bits column is hidden, which is kind of misleading since all the 32 bits cells are "coming soon", except macOS:

screenshot_24-01-2018_19-44-53

What it looks like with this PR:

screenshot_24-01-2018_19-44-42

PR #2639, opened 2018-01-24
Remove 32 bits column

Angristan:
As discussed in https://github.com/zeit/hyper/pull/2638, I'm removing the 32 bits column.

I also included the CSS fix in case you re-add it.

How it looks now:

screenshot_24-01-2018_20-44-07

screenshot_24-01-2018_20-43-50


Angristan:
🙌

PR #2659, opened 2018-02-04
Allow Hyper to run using local config if it exists (make Hyper portable)

MrMaxie:
Some people (like me) prefer to working on portable applications. Sometimes it is just a needed. I did the same as they did with Atom.

The application checks whether a configuration file exists next to the application's executable file. In contrast to the development version - the executable file is the file combined with the update - not electron directly.

If you need to make Hyper portable, just paste your configuration file in the application folder.

The only change I've made is to check if a file with a local configuration exists. And if it exists, it will replace directory and path of config. Thanks to this, the plugins will also appear in the folder next to the executable file.

### DIFF
--- a/app/config/paths.js
+++ b/app/config/paths.js
@@ -1,15 +1,25 @@
 // This module exports paths, names, and other metadata that is referenced
 const {homedir} = require('os');
-const {statSync} = require('fs');
-const {resolve, join} = require('path');
+const {statSync, existsSync} = require('fs');
+const {resolve, join, dirname} = require('path');
 const isDev = require('electron-is-dev');
+const {app} = require('electron');
 
 const cfgFile = '.hyper.js';
 const defaultCfgFile = 'config-default.js';
 const homeDir = homedir();
+const appDir = resolve(dirname(app.getPath('exe')), '..');
+
+let cfgLocalPath = join(appDir, cfgFile);
 
 let cfgPath = join(homeDir, cfgFile);
 let cfgDir = homeDir;
+
+// make portable if can
+if (existsSync(cfgLocalPath)) {
+  cfgPath = cfgLocalPath;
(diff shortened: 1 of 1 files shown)


ppot:
This shouldn't be done at the `paths.js` level. I will let @chabou  give it's take on this.

DRSDavidSoft:
As stated in #2932, I'd appreciate it if you give us an option in CLI to specify the path for it.

Currently, Hyper accepts _absolutely no arguments_ other than which path to start in, which is mildly annoying.

Enteleform:
Not sure what the build process is like for Hyper, but just saw that [**electron-builder**](https://www.electron.build) has portable builds.

DRSDavidSoft:
Curious... any updates on this PR?

MrMaxie:
> Not sure what the build process is like for Hyper, but just saw that [**electron-builder**](https://www.electron.build) has portable builds.

There aren't problem with "portability" of Hyper, but Hyper's config. It's static placed in (if I remember correctly) "documents" directory. It's absolutly fine when you work on static enviroment, but when you use your drive in multiple computers or just need to remove os from partition for some reasons - you will lose "your" config. And no - remembering about every programs' config file from each path isn't solution.

IMHO Hyper should in firstly check existence of config placed in the same directory which executable is. It's easiest way to create optional possibility to create Hyper portable without any problems for people which don't need that.

caseykey:
Would be really cool if hyper.js could be portable, I'd like to get a terminal on a pendrive with pure oh-my-zsh theme.

DRSDavidSoft:
@Stanzilla I'm assuming this issue is closed without being solved, if I'm not mistaken?

Stanzilla:
Correct, since there was no response for 2y

PR #2663, opened 2018-02-07
Electron 1.8 is out

Stanzilla:
Time to upgrade! :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ language: node_js
 matrix:
   include:
     - os: linux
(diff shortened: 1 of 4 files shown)


MindTooth:
Sweet!

@chabou:  Did more testing with dependabot: https://github.com/MindTooth/hyper/pulls?q=is%3Apr+is%3Aclosed

They all merge with success on Travis.  It pulled in 20 updates.

chabou:
@MindTooth Thank you very much for your insights. We definitely need to decide to use something like this!

Issue #1683, opened 2017-03-18
How to show native title bar and close icon on Arch Linux / Gnome?


alberto2000:
I guess most people wanted the contrary / to hide the title bar - now I'm asking if it's possible to have an option to let electron show the native window title bar. Is it possible? Thanks

This might be similar to https://github.com/zeit/hyper/issues/1675

alberto2000:
I figured if switching the browserDefaults frame value to true also for linux process platform, the windows will have the native frame and therefore title bar and menu. Now I'd like to move the working directory title from where it is originally to the native title bar title - so the window has the same, native configuration as any other linux app in my desktop environment, see screenshots comparing my current hyper build to gnome terminal:

![My current hyper build](http://i.imgur.com/RtDV4Mw.png)

![Gnome terminal](http://i.imgur.com/3v9Vld2.png)

a-eid:
how did you get the default os decorations to work?

jcrben:
If you're on Linux and you don't have the native title menu, then you may not be able to get to the menu at all. You need to enable `showHamburgerMenu` - discussed at https://github.com/zeit/hyper/issues/1266.

Stanzilla:
Here you go https://www.npmjs.com/package/hyper-native-window-decoration

Issue #1373, opened 2017-01-10
Hyper does not show menus on Fedora 25 with Gnome


possebon:
- [X] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [X] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Fedora 25 Kernel 4.8.15
- **Hyper.app version**: the latest downloaded yesterday (09/01/2017)
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/possebon/0b1a0169dc28e264bbc55c10278e96f0
- **Relevent information from devtools** _(CMD+SHIFT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: http://imgur.com/a/07CkU
- **The issue is reproducible in vanilla Hyper.app**: 

## Issue
After downloading the Hyper.app I run with sucess but there are no menus. I tried on Windows and MacOS with success, but on Linux (at least on my workstation), it does not work as expected.

This is the screenshot of my screen: http://imgur.com/a/xl50w

I saw a video on Youtube of Hyper running on Fedora 25 without any problems.

stefanivic:
@possebon in ~/.hyper.js set the showHamburgerMenu to true. 
So it's `showHamburgerMenu : true,`

possebon:
Thanks for you reply.

I changed but did not showed the menu anyway.

I just downloaded the version 1.1.0 too, and did not work either.

bryantee:
@possebon I just got it working myself on F25. Make sure you're not using quotes as `true` should be a boolean not a string.

possebon:
I got it working too now. It was the true value that I was using quotes.

Thanks Bryan.

Issue #1234, opened 2016-12-15
Menu only show when maximized on Ubuntu


RomuloPBenedetti:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Ubuntu 16.04 LTS 
- **Hyper.app version**: 1.0.0 
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/RomuloPBenedetti/09c06cbdba7bbbfed16584f5d21554fc
- **Relevent information from devtools** _(CMD+SHIFT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: N/A 
- **The issue is reproducible in vanilla Hyper.app**:  Vanilla

## Issue

I've configured my ubuntu to show menu in title bar, It's the only relevant setting I believe can be a cause of this problem.

saule1508:
Hello,

On Fedora the menu does not show at all. With the previous version (0.8) there is a normal menu, I've seen from the release notes that this got changed to an hamburger menu in 1.0 but when I run 1.0 on Fedora there is no menu at all.

Thanks

swhammack:
This is also an issue in Arch running Cinnamon as desktop environment. Can't see the menu maximized or any other way.

penance316:
I have seen the same issue on Linux Mint XFCE desktop ( no hamburger menu even with config set to true )

saule1508:
As indicated in #1266, editing .hyper.js in the home directory and adding this in the config object solved the problem on Fedora
```
module.exports = {
  config: {
     showHamburgerMenu: true,
     // default font size in pixels for all tabs
     fontSize: 14,
     etc..
```

albinekb:
@RomuloPBenedetti can you see if @saule1508 's solution does it for you? :) I think it's intended. Re-open if i'm wrong.

PR #2666, opened 2018-02-10
on Linux default to "showHamburgerMenu": true

jcrben:
I believe this is ready to merge, but I didn't go through and run from source to check. This is partly to open a discussion and get a better understanding of why this default was selected. Another possible resolution to the underlying issue is figuring out how to get the title menu bar to work consistently on Linux.

If a minimal flow is felt to be super important, then I recommend an in-app onboarding wizard to help users understand how to access their menu bar. But until that's built, a menu should be visible by default.

From the code comment `// show by default on Windows and Linux`, it appears that the author expected the window menu to appear on Linux, but this isn't actually happening. Therefore I think it would be better to default to showing the hamburger menu and letting experienced users configure it away.

When I loaded up hyper on Linux, I couldn't access the menu at all. I saw this - and I didn't have an overall menu bar like macOS on Ubuntu 17.10 running KDE Plasma thru Virtualbox:
image

Issues from this that I found quickly:
* https://github.com/zeit/hyper/issues/1266 (most commonly-cited)
* https://github.com/zeit/hyper/issues/1118
* https://github.com/zeit/hyper/issues/1168
* https://github.com/zeit/hyper/issues/1373
* https://github.com/zeit/hyper/issues/1234
* https://github.com/zeit/hyper/issues/1195

[Blame](https://github.com/jcrben/hyper/blame/887fb5ac1a56d50d2f7ab694b8b1e92a2cebc101/lib/components/header.js#L75) points to https://github.com/zeit/hyper/pull/946 which does not discuss this issue.

There is a mention of the frameless thing which I haven't messed around with (https://github.com/zeit/hyper/issues/795 - maybe that could help with https://github.com/zeit/hyper/issues/1683 and showing the title menu bar).

### DIFF
--- a/app/config/config-default.js
+++ b/app/config/config-default.js
@@ -51,9 +51,8 @@ module.exports = {
     // custom CSS to embed in the terminal window
     termCSS: '',
 
-    // set to `true` (without backticks and without quotes) if you're using a
-    // Linux setup that doesn't show native menus
-    // default: `false` on Linux, `true` on Windows, ignored on macOS
+    // if you're using a Linux setup which show native menus, set to false
+    // default: `true` on Linux, `true` on Windows, ignored on macOS
     showHamburgerMenu: '',
 
     // set to `false` (without backticks and without quotes) if you want to hide the minimize, maximize and close buttons
--- a/lib/components/header.js
+++ b/lib/components/header.js
@@ -72,7 +72,7 @@ export default class Header extends PureComponent {
     const {showHamburgerMenu, showWindowControls} = this.props;
 
     const defaults = {
-      hambMenu: process.platform === 'win32', // show by default on windows
+      hambMenu: !this.props.isMac, // show by default on windows and linux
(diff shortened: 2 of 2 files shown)


chabou:
I am agreed. We need to enable it by default on Linux.
Thank you so much for your PR.

Do you mind to modify comment in default config file too?
https://github.com/zeit/hyper/blob/76beaa32f5f3285bc8f7083fff8cd722ba97f60a/app/config/config-default.js#L56

jcrben:
Updated. I think this gets into the whole Client Side Decorations going on in the Linux world. GNOME has an initiative to focus on client-side window decorations (https://wiki.gnome.org/Initiatives/CSD).

Using the hamburgerMenu aligns hyper with GNOME, but also I'm running KDE on Ubuntu 17.10 and it also isn't showing native menus, so it seems to cover modern KDE too (granted, I'm building Ubuntu from a server ISO https://gitlab.com/jcrben-play-learn/packer-desktop).

PR #2684, opened 2018-02-19
Add support for platform-specific shell config

### LINKED ISSUE #2584: To support a single cross-platform settings file some of the settings should be platform specific (like in VS Code)

- **OS version and name**: Windows 10 1709, Ubuntu 16.04, Ubuntu 17.10, macOS Sierra
- **Hyper.app version**: 1.4.8

## Feature Request
At present, with only a single "shell", "shellArgs", and "env" options I must maintain three different configs for the different OS platforms (where the only difference is in these three settings).  This requires me to synchronize the rest of the settings manually in order to have the same experience across platforms.  This kind of defeats the purpose of a cross-platform tool such as this.

VS Code gets around this problem by providing settings that are platform specific and the tool figures out which setting to use based on which platform it is running on.   This is how I can have a single setting for VsCode that is shared across all machines.

In VS Code the relevant settings section looks like this:

"terminal.integrated.env.linux": {},
"terminal.integrated.env.osx": {},
"terminal.integrated.env.windows": {},

"terminal.integrated.shell.linux": "sh",
"terminal.integrated.shell.osx": "sh",
"terminal.integrated.shell.windows": "C:\\WINDOWS\\System32\\WindowsPowerShell\\v1.0\\powershell.exe",

"terminal.integrated.shellArgs.linux": [],
"terminal.integrated.shellArgs.osx": [
    "-l"
  ],
"terminal.integrated.shellArgs.windows": [],

Perhaps we could use:

shell.linux
shellArgs.linux
env.linux
shell.windows

and so on.

brandon93s:
This PR introduces platform-specific config support for `windows`, `osx` and `linux` as discussed in #2584. This has been implemented by transforming / extracting platform specific values at cfg load time if available and promoting them to the root cfg properties. This adds support without the need to modify up-stream consumers of the configuration.

This functionality _could_ be expanded to all config options, but is currently limited to just those options that would need to differ by platform. 

**TODO:**
- [x] Document new functionality (website or config-default.js?)
---
Closes #2584

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,6 +8,15 @@
       "runtimeExecutable": "${workspaceRoot}/node_modules/.bin/electron",
       "program": "${workspaceRoot}/app/index.js",
       "protocol": "inspector"
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Run AVA test",
+      "program": "${workspaceRoot}/node_modules/ava/profile.js",
+      "args": [
+        "${file}"
+      ]
     }
   ]
 }
--- a/app/config/config-default.js
+++ b/app/config/config-default.js
@@ -98,13 +98,17 @@ module.exports = {
     //
     // PowerShell on Windows
     // - Example: `C:\\WINDOWS\\System32\\WindowsPowerShell\\v1.0\\powershell.exe`
+    //
+    // supports platform specific configuration: shell.windows|linux|osx: ''
     shell: '',
 
     // for setting shell arguments (i.e. for using interactive shellArgs: `['-i']`)
     // by default `['--login']` will be used
+    // supports platform specific configuration: shellArgs.windows|linux|osx: []
     shellArgs: ['--login'],
 
     // for environment variables
+    // supports platform specific configuration: env.windows|linux|osx:{}
     env: {},
 
     // set to `false` for no bell
--- a/app/config/init.js
+++ b/app/config/init.js
@@ -1,6 +1,7 @@
 const vm = require('vm');
 const notify = require('../notify');
 const mapKeys = require('../utils/map-keys');
+const platformSpecific = require('./platform-specific');
 
 const _extract = function(script) {
   const module = {};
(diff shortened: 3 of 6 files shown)


.vscode/launch.json
```diff
@@ -8,6 +8,15 @@
       "runtimeExecutable": "${workspaceRoot}/node_modules/.bin/electron",
       "program": "${workspaceRoot}/app/index.js",
       "protocol": "inspector"
+    },
```
brandon93s: Debug the currently open AVA Unit Tests

brandon93s:
Website changes: 

image

chabou:
Amazing work! Thank you so much!

I'll review and merge it once v2 is out: soooooon!

albinekb:
Isn't this already possible, our config file is in JS

```js
module.exports = {
  env: process.platform.os === 'darwin' ? { THIS_IS_MAC_ONLY: 'test' } : {}
```

Why is this needed?

timothyis:
@albinekb The original solution seems cleaner and easier to read 🤔 
Also offloads work for the user

albinekb:
Then why do we even have the config in JS... this will be PITA to maintain and test when it's already possible (even more powerful config can be made as it's in JS) I think we should document this or add helper functions to select between different config on different OS, having it like this PR implements it will add mode things that can and will go wrong and open up to more issues. My 5c.

bdougherty:
@albinekb unless I'm doing something wrong, `process` is not available in the config file.

albinekb:
@bdougherty haha, you're correct. I fixed it with this change:  https://github.com/zeit/hyper/pull/2716, after this patch it works.

brandon93s:
Closing in favor of #2716.

chabou:
I really want this PR to be merged. I think that is a better way to handle platform specifics than let users adding some algorithm in config file.

chabou:
@albinekb I think it will be easier to maintain with this PR than introducing some algorithms in config file. And I think this PR adds really nice user-friendly feature. A lot of our users have never written a line of JS.

The only reason that we have our config in JS is to have some comments I think. It would be awesome to migrate it to a [Json5 format](https://github.com/json5/json5).

Using some per-OS configuration will maybe lead to some misunderstanding and some issue opening. But I think that it will be the case with your method too.

chabou:
@brennanfee 
I really think that your request should be implemented as this. I don't want that people understand javascript and nodejs (`process.*` thing), to implement some platform specific config. VSCode is a good example as you said in #2584

But we can have both :) (But I really don't like to expose `process.*` in config but let discuss this on #2716)

albinekb:
I'm starting to think we should go towards changing the config to `.json`, since we're modifying it programatically when you install plugins via reading and chaning the AST, which is really fragile.

If people start to modify their config too much (for example looping together their `plugins`, the installation would fail, if we are using JSON it will always be correct and also easier to verify with json schema. 🤔

Stanzilla:
@brandon93s willing to pick this up again?

Stanzilla:
closing in favor of https://github.com/zeit/hyper/pull/1347

PR #2689, opened 2018-02-20
Fix transparency at startup

### LINKED ISSUE #2688: Semi-transparent background not working as expected
How to reproduce:

1.  I have `2.0.0-canary.13`
2. I have this line in my `.hyper.js`
````js
backgroundColor: 'rgba(0, 0, 0, .7)',
````
3. I open hyper and see a completely transparent background
4. I open preferences, background becomes completely black
5. I click on hyper to regain focus, background becomes as expected (semi-transparent black)

chabou:
Fixes #2688

### DIFF
--- a/app/ui/window.js
+++ b/app/ui/window.js
@@ -61,6 +61,7 @@ module.exports = class Window {
 
     rpc.on('init', () => {
       window.show();
+      updateBackgroundColor();
 
       // If no callback is passed to createWindow,
       // a new session will be created by default.


cooperaj:
This appears to have broken Vibrancy on MacOS. 

My lovely blurred terminal window is now a solid grey. 

Plugins broken that did work "hyper-smyck" and "hyper-vibrancy"

chabou:
I think this not related to this fix but has the same cause (race condition at window init)

I'll try to investigate deeper soon. Thank you for your feedback 🙏

chabou:
@cooperaj Vibrancy works well for me
image

Can you open an issue if it really doesn't work?

Issue #2687, opened 2018-02-20
Reset splits either side of a divider to an even ratio on double click of the divider


timothyis:
## Enhancement
Would be great if double clicking the divider between two splits would reset those two splits to an even size. Otherwise you have to drag and guess.

![image](https://user-images.githubusercontent.com/1695613/36427984-957b5bd4-1646-11e8-9a5e-5a71ee8dbded.png)

maikwoehl:
I think `` hyper/lib/components/split-pane.js `` ist the relevant part. 

I will take some time of my break at work to think about it. 

Maybe I will come up with a solution. 

Please don't assign me yet. I don't know if I can find time to contribute. But I will try.

lesce:
@maikwoehl I can take a look also , if you haven't started it .
I already have a working prototype , could probably create a PR today 
Let me know what your status is on this

lesce:
added a PR for this , would love some feedback on it @maikwoehl @timothyis

timothyis:
Merged in #2692 🎉 

Huge thanks to you @lesce ❤️ and for your review and interest @maikwoehl 👏 

I hope we see you contributing again 🙏

PR #2692, opened 2018-02-21
Reset splits either side of a divider to an even ratio on double click of the divider #2687

lesce:
This feat was requested here https://github.com/zeit/hyper/issues/2687
Should be ready to merge

### DIFF
--- a/lib/components/split-pane.js
+++ b/lib/components/split-pane.js
@@ -31,11 +31,11 @@ export default class SplitPane extends PureComponent {
 
     this.setupPanes(ev);
 
-    let sizes_ = this.getSizes();
+    const sizes_ = this.getSizes();
     sizes_[this.paneIndex] = 0;
     sizes_[this.paneIndex + 1] = 0;
 
-    let availableWidth = 1 - _.sum(sizes_);
+    const availableWidth = 1 - _.sum(sizes_);
     sizes_[this.paneIndex] = availableWidth / 2;
     sizes_[this.paneIndex + 1] = availableWidth / 2;
 
@@ -66,7 +66,7 @@ export default class SplitPane extends PureComponent {
   }
 
   getSizes() {
-    let {sizes} = this.props;
+    const {sizes} = this.props;
     let sizes_;
 
     if (sizes) {
@@ -81,7 +81,7 @@ export default class SplitPane extends PureComponent {
   }
 
   onDrag(ev) {
-    let sizes_ = this.getSizes();
+    const sizes_ = this.getSizes();
 
     const i = this.paneIndex;
     const pos = ev[this.d3];
--- a/lib/components/split-pane.js
+++ b/lib/components/split-pane.js
@@ -1,11 +1,13 @@
 /* eslint-disable quote-props */
 import React from 'react';
 import {PureComponent} from '../base-components';
+import _ from 'lodash';
 
 export default class SplitPane extends PureComponent {
   constructor(props) {
     super(props);
     this.handleDragStart = this.handleDragStart.bind(this);
+    this.handleAutoResize = this.handleAutoResize.bind(this);
     this.onDrag = this.onDrag.bind(this);
     this.onDragEnd = this.onDragEnd.bind(this);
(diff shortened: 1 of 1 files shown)


lib/components/split-pane.js
```diff
@@ -53,9 +75,13 @@ export default class SplitPane extends PureComponent {
...
       const count = new Array(total).fill(1 / total);
 
-      sizes = count;
       sizes_ = count;
     }
+    return sizes_;
+  }
+
+  onDrag(ev) {
+    let sizes_ = this.getSizes();
```
albinekb: Does this need to be run on every drag event?

timothyis: I guess it was already doing that. This just makes it accessible from elsewhere.

albinekb: Oh, never mind. Yep it was 👍

lib/components/split-pane.js
```diff
@@ -18,6 +20,28 @@ export default class SplitPane extends PureComponent {
...
+    this.paneIndex = this.panes.indexOf(ev.target);
+    this.paneIndex -= Math.ceil(this.paneIndex / 2);
+  }
+
+  handleAutoResize(ev) {
+    ev.preventDefault();
+
+    this.setupPanes(ev);
+
+    let sizes_ = this.getSizes();
```
albinekb: `const`, you're not reassigning the reference later?

lib/components/split-pane.js
```diff
@@ -18,6 +20,28 @@ export default class SplitPane extends PureComponent {
...
+    sizes_[this.paneIndex] = 0;
+    sizes_[this.paneIndex + 1] = 0;
+
+    let availableWidth = 1 - _.sum(sizes_);
```
albinekb: same here

lib/components/split-pane.js
albinekb: `const` here too I think 😃

lesce:
@albinekb I've addressed your feedback , thanks for the quick review 👍 💯 🥇

timothyis:
Thank you 🙌

lib/components/split-pane.js
```diff
@@ -18,6 +20,28 @@ export default class SplitPane extends PureComponent {
     }
   }
 
+  setupPanes(ev) {
```
chabou: This function deserves a better name. It doesn't setup anything. It saves dragging pane Index.

lib/components/split-pane.js
```diff
@@ -18,6 +20,28 @@ export default class SplitPane extends PureComponent {
...
+    this.panes = Array.from(ev.target.parentNode.childNodes);
```
chabou: (This is not your code originally but..)
Why do not use a local variable ?

Btw, I think it is better to not determine index from DOM. Certainly better to add it to handler when rendering divider.

lib/components/split-pane.js
```diff
@@ -18,6 +20,28 @@ export default class SplitPane extends PureComponent {
...
+    this.paneIndex = this.panes.indexOf(ev.target);
```
chabou: (This is not your code originally but..)
paneIndex is valide/set only after a dragStart.
Maybe better to have a name like : `this.draggingPaneIndex`

But if we use variable when rendering, we don't need it anymore

lib/components/split-pane.js
```diff
@@ -18,6 +20,28 @@ export default class SplitPane extends PureComponent {
...
+  handleAutoResize(ev) {
+    ev.preventDefault();
+
+    this.setupPanes(ev);
+
+    const sizes_ = this.getSizes();
+    sizes_[this.paneIndex] = 0;
+    sizes_[this.paneIndex + 1] = 0;
+
+    const availableWidth = 1 - _.sum(sizes_);
```
chabou: I prefer using map/filter/reduce when it is possible :)

`const availableWidth = sizes_.reduce((total, size) => total - size, 1);`

chabou:
@lesce Thank you so much for your work 🙏

Issue #2674, opened 2018-02-15
CMD + R breaks Hyper


kuoppa:
When pressing CMD + R when window is active it breaks the app. it clears all texts and no tab is working except the one you're in. So it results in a reboot of the  app.

- **OS version and name**: 10.13.1 High Sierra
- **Hyper.app version**: 1.4.8

lesce:
I can take a look 👀

kuoppa:
This still breaks Hyper. Even in 2.0.0.

chabou:
Reload and ForceReload will be removed soon

graciano:
Is there a way to configure it to do nothing when I press cmd+R?

timothyis:
Possibly by registering a custom keymap, although I'm not sure @graciano

timothyis:
I'm going to merge this issue into #921. I feel like this issue belongs in the same realm.

PR #2698, opened 2018-02-22
CMD + R breaks Hyper #2674

lesce:
As mentioned in this issue https://github.com/zeit/hyper/issues/2674 the current Reload functionality is broken .
I'm not really sure what the difference should be between Reload ( cmd+R ) and Clear buffer ( cmd+K ) if there is one . Tested it on Terminal.app and iTerm2 and they seem to do the same thing .

### DIFF
--- a/app/commands.js
+++ b/app/commands.js
@@ -36,7 +36,7 @@ const commands = {
     updatePlugins();
   },
   'window:reload': focusedWindow => {
-    focusedWindow && focusedWindow.rpc.emit('reload');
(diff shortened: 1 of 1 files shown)


chabou:
Our reload is not working as expected but this is not attended to clear output.
Actually, `Reload` should force a React update without removing any output. But it seems to be more a development need.

I think that in near future, "Force reload" should, at least, replace "Reload". But user can restart application/window, thus imo, we'll remove both.

Thank you for your help and your work 🙏 
But I must reject this PR.

PR #2700, opened 2018-02-23
CLI: better existence check

chabou:
NPM registry is broken for some (unpublished?) package:
https://registry.npmjs.org/hyper-cwd
https://api.npms.io/v2/package/hyper-cwd (error 500)

Checking headers on npm is not enough.

Now, `versions` key existence is checked in content

### DIFF
--- a/cli/api.js
+++ b/cli/api.js
@@ -64,18 +64,11 @@ function save() {
 
 function existsOnNpm(plugin) {
   const name = plugin.split('#')[0].split('@')[0];
-  return got
-    .get(registryUrl + name.toLowerCase(), {timeout: 10000, json: true})
-    .then(res => {
-      if (!res.body.versions) {
-        throw new Error();
-      }
-    })
-    .catch(() => {
-      const err = new Error(`${plugin} not found on npm`);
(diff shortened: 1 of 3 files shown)


cli/api.js
```diff
@@ -62,14 +63,19 @@ function save() {
...
 function existsOnNpm(plugin) {
-  plugin = plugin.split('#')[0].split('@')[0];
-  return npmName(plugin).then(unavailable => {
-    if (unavailable) {
+  const name = plugin.split('#')[0].split('@')[0];
+  return got
+    .get(registryUrl + name.toLowerCase(), {timeout: 10000, json: true})
+    .then(res => {
+      if (!res.body.versions) {
+        throw new Error();
```
ppot: Error should be defined with a message.

brandon93s: Meh, just used to get into the `.catch` block here. Could alternatively `return Promise.reject();`.

ppot: Your ready to go. 🚀

chabou:
@brandon93s @ppot Thank you for your reviews. You were both right. I refactored it and looks a way better now.
Bonus: There is a convenient message for errors.

PR #2711, opened 2018-02-25
Added menu item to to change update channel

### LINKED ISSUE #2708: Add a menu checkbox to switch to canary/stable update channel
We really need a better way to change Hyper update channel.

The idea is to add an item `Canary Releases` right above `Preferences...` menu item in:
* `Hyper` menu on macOS
* `About` menu on Windows/Linux

This checkbox should be synchronized with config file:
* Checking it will update/add an `"updateChannel"` setting with a `"canary"` value.
* Unchecking it will set `"updateChannel"` setting to `"stable"` value.

If config file is updated manually, checkbox should reflect the new value. If the value is not `'canary'` (or missing), it should be considered as `'stable'`.

All other part of config file should remain unmodified.
To achieve this, using `recast` could be a great option. Example CLI that modifying `plugins` array in config file:
* Parse file content: https://github.com/zeit/hyper/blob/76beaa32f5f3285bc8f7083fff8cd722ba97f60a/cli/api.js#L38
* Finding the right setting: https://github.com/zeit/hyper/blob/76beaa32f5f3285bc8f7083fff8cd722ba97f60a/cli/api.js#L42
* Modifying the value: https://github.com/zeit/hyper/blob/76beaa32f5f3285bc8f7083fff8cd722ba97f60a/cli/api.js#L84
* Writing modified file: https://github.com/zeit/hyper/blob/76beaa32f5f3285bc8f7083fff8cd722ba97f60a/cli/api.js#L61

aditya0099:
I added a checkbox to the edit menu that will switch the updateChannel between 'canary' and 'stable'. To do this, I used recast to parse and edit the config file.

Fixes #2708

### DIFF
--- a/app/config/edit.js
+++ b/app/config/edit.js
@@ -3,27 +3,27 @@ const fs = require('fs');
 const recast = require('recast');
 
 function toggleUpdateChannel() {
-  // Get the config file using recast
-  const configFile = recast.parse(getFileContents(cfgPath));
+  let updateChannel = '';
+  let configFile = '';
 
   // Search the config file for the property with the name 'updateChannel'
   try {
     // Get the config file using recast
-    const configFile = recast.parse(getFileContents(cfgPath));
+    configFile = recast.parse(getFileContents(cfgPath));
 
     // Search the config file for the property with the name 'updateChannel'
-    const updateChannel = configFile.program.body[0].expression.right.properties
+    updateChannel = configFile.program.body[0].expression.right.properties
       .find(property => property.key.name === 'config')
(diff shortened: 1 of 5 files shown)


app/config/edit.js
```diff
@@ -0,0 +1,55 @@
+const {cfgPath} = require('./paths');
+const fs = require('fs');
+const recast = require('recast');
+
+function toggleUpdateChannel() {
+  let updateChannel = '';
+  let configFile = '';
```
albinekb: This can be just

PR #2716, opened 2018-02-26
Allow process to be used inside config

albinekb:
As discussed in https://github.com/zeit/hyper/pull/2684 it should be possible to use process.platform and process.env inside your config file so you can do stuff like:

```js
module.exports = {
  config: {
    env: process.platform === 'darwin' ? { THIS_IS_MAC_ONLY: 'test' } : {}
  }
}
```
```js
module.exports = {
  config: {
    shell: { win32: 'powershell.exe', darwin: 'fish', linux: 'bash' }[process.platform]
  }
}
```

```js
module.exports = {
  config: {
    env: { win32: {}, darwin: {}, linux: {} }[process.platform]
  }
}
```

### DIFF
--- a/app/config/init.js
+++ b/app/config/init.js
@@ -4,7 +4,7 @@ const mapKeys = require('../utils/map-keys');
 
 const _extract = function(script) {
   const module = {};
-  script.runInNewContext({module});
+  script.runInNewContext({module, process});
   if (!module.exports) {
     throw new Error('Error reading configuration: `module.exports` not set');
   }


brandon93s:
+1 I much prefer this to my proposed PR. I am however curious why we're loading this sandboxed to begin with and just just requiring in a try...catch? Once we start passing things like the process to the sandbox we've lost the purpose of the sandbox. 

The code is a good bit more complex in a VM. 

Original: https://github.com/zeit/hyper/commit/b6990f754367b8d45bd302d54eeb7fde0e7dce11

albinekb:
Awesome! 🙏  The reason we run it in a VM context is to get the syntax errors so we can display that in a notification letting the user know something is up. But we might get the same result from using `require` inside `try...catch`?

brennanfee:
I opened the original request #2584.  

I also would vote for this implementation rather than @brandon93s implementation (even though his was really good as well - thank you Brandon).  This version would allow configs to differ not only by platform (as I requested) but also by bit level (process.arch) or by environment variables (process.env) and everything else in the process object.  That gives the user maximum flexibility in having a single config file that can be used in lots of different situations (which was the real ask).

While I do agree the syntax is not "cleaner" the power involved is worth a little bit of an uglier API so-to-speak.

albinekb:
@chabou from https://github.com/zeit/hyper/pull/2684#issuecomment-378838888
>But I really don't like to expose process.* in config but let discuss this on #2716

Why don't you want people to be able to write their config with `process`? Hyper is marketed as the hackable terminal built **with javascript**, I think it makes more sense to allow our users to leverage this rather than a boring old school config file seen in most other apps.

chabou:
Separation of concerns.

Config file should be declarative. We already have a wonderful plugin system if someone needs more power. I really feel that we'll introduce some forever evil with that.
This is already possible to have some test/functions in our config file but I really don't want to encourage this.
The more config file will use runtime values, the less we'll be able to reproduce an issue.

But this is only my very humble opinion.

LinusU:
> Config file should be declarative.

_If_ we want it to be that way, we should probably change the config file from a JavaScript file to JSON/TOML/YAML/whatever...

chabou:
`json5` would be perfect imho.

I totally agree with you: it is quite nice to offer some power to users to make crazy things in the config. But I really think that we'll pay this power later.

albinekb:
I think we should close this in favour of #2684, and go towards changing our config to `json5`, to minimize potential issues.

Issue #2702, opened 2018-02-24
Add flow toolchain/workflow


chabou:
We really want to add [flow](https://flow.org/) in our project.

First step is to add it to our dev workflow: add dependencies, linter compatibility, automatically run before push...

Please help us to have some benefits of typed system.

albinekb:
I think it would be better if we used TypeScript since it has more types available and specifically `xterm` which Hyper is now heavily dependent on is also written in TS.

jcrben:
+1 for TypeScript. TypeScript seems more widely known (I use TypeScript, have not yet used Flow) and seems to have more momentum. It can also pick up jsdoc...

rauchg:
I'm ok with either as long as we have an incremental adoption story. For example, if we can start replacing one .js file at a time with one .ts file, then we can incrementally make lots of little PRs to get to where we want to be.

So far, the appeal of Flow to me has been that incremental adoption strategy. If we can do that with TS, I'm game

chabou:
No problem to have incremental adoption with TS.
If there is no bonus to use flow (to be compliant with other Zeit's projects, for example), I think that @albinekb has right: Maybe better to use TS because `xterm` uses TS.

I let this issue opened to let other people add their inputs. I suggest to definitely choose our type system in 7 days.

Tyriar:
Re-adding my thoughts here from Slack. xterm.js was originally a 5k+ line file and separating it out into different files was one of the main reasons to use TS. I incrementally pulled modules out into their own files while converting, you can check what these sort of PRs look like with the linked PRs on https://github.com/xtermjs/xterm.js/issues/335, though it will certainly be easier with Hyper as you already use a newer style of JS.

saadq:
I'm a user of both TypeScript and Flow. For what it's worth, the [`xterm.d.ts` defs](https://gist.github.com/anonymous/057d178976966811c061d6514ca8d3a2) are pretty simple and it would be pretty trivial to write a Flow libdef from it (just taking a glance, I think the file is already 99% compatible with Flow). So if that's the deciding factor, I could submit a PR to `flow-typed` with a libdef for `xterm`.

With that said, TypeScript is definitely in the lead when it comes to tooling, community size, and developer support. The Flow team seems to not have too much time to deal with GitHub issues unfortunately, so the support you get from the repo will mainly be from Flow users rather than the Flow team itself:

[...]

PR #2717, opened 2018-02-26
Add flow type checking

Ameya05:
Added flow type checking support as requested in #2702 

Also added flow annotations and types in a few library util files.


rauchg:
We are going to switch to TypeScript

Issue #2740, opened 2018-03-07
SSH url scheme handling should not be added on dev mode


chabou:
Since #2516, Hyper is set as default program to handle `ssh://` url scheme.
It should not be done on dev mode.
Condition should be added here:
https://github.com/zeit/hyper/blob/0162a9f84dab44a5be395f4faf7579217822e5a6/app/index.js#L186-L195

cc @antonioalmeida

ajomadlabs:
Can I take up this issue

antonioalmeida:
Sure @ajomadlabs, you just need to check `isDev`, before applying the change

ajomadlabs:
@antonioalmeida Since this is my first issue, can you provide some resource, where to start off

diogotorres97:
@ajomadlabs literally do:
 if (! isDev){ line 188,189} 
else 
nothing;

Same thing to second case

antonioalmeida:
https://github.com/zeit/hyper/blob/0162a9f84dab44a5be395f4faf7579217822e5a6/app/index.js#L186-L195

 As @chabou mentioned, in this area you should check if Hyper is running on dev mode, and prevent the chances if so. To do this, you can check `isDev`'s value, similar to what is done here:
 https://github.com/zeit/hyper/blob/0162a9f84dab44a5be395f4faf7579217822e5a6/app/index.js#L88-L108

Is that clear?

ajomadlabs:
I followed the contribution guidelines but I am unable to set up. I am encountering with the following error
![errorhyper](https://user-images.githubusercontent.com/21251118/37132946-9028929c-22b6-11e8-86f0-daedd256f3af.png)

ajomadlabs:
@chabou I created a pull request https://github.com/zeit/hyper/pull/2745. Please do suggest me changes if any

antonioalmeida:
@chabou no problem! 👍 
good job @ajomadlabs

PR #2745, opened 2018-03-08
Do not register app for `ssh://` url scheme handling on dev mode

ajomadlabs:
Have added a condition to check if its in Dev mode or not

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -183,17 +183,19 @@ app.on('ready', () =>
       // expose to plugins
       app.createWindow = createWindow;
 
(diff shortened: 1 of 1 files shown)


chabou:

Some advices:
For your next PR, keep in mind that the title of the PR will be shown in the Release Notes. It should be chosen to be understable by final user.
And this is a good practice to add issue number with `Fixes` prefix in your comment as your commit messages.

Fixes #2740

ajomadlabs:
@chabou Sure. :+1:

PR #2757, opened 2018-03-13
Remove auto prefix plugin name

### LINKED ISSUE #2739: [CLI] Do not automatically prefix package name when installing
For now, CLI automatically adds `hyper-` prefix when installing/uninstalling plugins.
`hyper i pane` do exactly the same thing than `hyper i hyper-pane`.
But if user wants to install a plugin with a non standardized name like `hyper i hyperline`, it will try to install `hyper-hyperline`.
It leads to some security problems. If someone forks `hyperline`, adds some malicious code and published it as `hyper-hyperline`...

Please remove any automatically prefixing in CLI.

albinekb:
Fix #2739

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -23,10 +23,6 @@ const assertPluginName = pluginName => {
   }
 };
 
-const getPluginName = arg => {
-  return arg;
-};
-
 const checkConfig = () => {
   if (api.exists()) {
     return true;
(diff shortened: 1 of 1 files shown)


cli/index.js
```diff
@@ -24,9 +24,6 @@ const assertPluginName = pluginName => {
 };
 
 const getPluginName = arg => {
```
chabou: We don't need this function anymore (identity)

PR #2758, opened 2018-03-13
Update to CircleCI 2

timothyis:
Should be ready to merge, I think.

Updates our config and build step with CircleCI to their version 2 system


.circleci/config.yml
```diff
@@ -0,0 +1,88 @@
...
+      - install
+      - build:
+          requires:
+            - install
+          filters:
+            branches:
+              only:
+                - master
+                - canary
+                - circleci-2
```
albinekb: Should this be here in this PR or was it for testing 🕵️

timothyis: Good spot, I just didn't save the file 😅

leo:
amazing, thank you!

PR #2761, opened 2018-03-14
Migrate styling to styled-jsx

### LINKED ISSUE #2737: Styled-jsx migration
I would love to replace `aphrodite` by `styled-jsx`.
It would be more consistent to use Zeit tool :)

timothyis:
This PR should be ready to go, pending a review.

Basically, I wanted to learn more about the Hyper codebase instead of sitting behind the scenes so much. The way I began to plan this was mentioning that it'd be a good idea to transition Hyper to use styled-jsx. I wanted to do this for a few reasons.
1. It's consistent with other ZEIT codebases using React (since styled-jsx is by ZEIT)
2. It lessens the barrier for contribution for those that only want to change some styling

I had a few problems along the way. The biggest problem was that styled-jsx would leave the classname as it's written in the code but it'd generate a new class, e.g. `jsx-2394298342`, for the component. The styling would then target `.og-class.jsx-2394298342` making the weight of the styles higher than the custom styling that just targets the original class.
To get around that I applied a scope around Hyper using the ID `#hyper` on the root div. I then used `stylis` to prepend each selector definition in the customCSS with that ID. This makes the weight of the customCSS higher and more specific.

Another problem is that, since we used to use aphrodite and that generated classes for every element with a class (not leaving the original), we had some logic to add a global class based on the component and the class of each element. This was no longer necessary since styled-jsx left the original class behind. But since the original class didn't match the class we generated ([component]_[class]), I changed each class to follow that naming convention individually. 
This offers a few benefits:
1. It makes sense when reading the code why the class ends up with that name
2. If we want to add components in the future we can either follow that naming convention or not. There's no logic needed for it.

So there it is. This should close #2737.
- I migrated all the styling to `styled-jsx` from `aphrodite-simple`
- Remove the need for `base-components` so we can use just direct React Components or Pure Components (easier to read and navigate, hooray!)
- Introduced a naming convention for the CSS
- Scoped the custom CSS to have more weight
- Prettier watches the CSS and enforces code style (a big bonus!)

cc @giuseppeg (thanks for the help!)

*Note*: I have only tested this on MacOS High Sierra (not that the version makes a difference). I'd love for this to be tested on Windows and Linux too!


lib/components/header.js
```diff
@@ -109,30 +108,42 @@ export default class Header extends PureComponent {
 
     return (
       <header
-        className={css('header', isMac && 'headerRounded')}
+        className={`header_header ${isMac && 'header_headerRounded'}`}
```
giuseppeg: have you checked the result of this? when `isMac` is falsy it should render `header_header false`

giuseppeg: you might want to use a ternary

lib/components/notification.js
```diff
@@ -63,51 +62,54 @@ export default class Notification extends PureComponent {
...
-      }
-    };
+          .notification_dismissLink {
+            position: relative;
+            left: 4px;
+            cursor: pointer;
+            color: #528D11;
+          }
+
+          .notification_dismissLink:hover {
```
giuseppeg: maybe you could also style the `:focus` pseudo class

lib/components/split-pane.js
```diff
@@ -134,80 +133,79 @@ export default class SplitPane extends PureComponent {
...
+          .splitpane_divider_horizontal {
+            height: 11px;
+            margin: -5px 0;
+            border-top: 5px solid rgba(255, 255, 255, 0);
+            border-bottom: 5px solid rgba(255, 255, 255, 0);
+            cursor: row-resize;
+            width: 100%;
+          }
+
+          {/*
```
giuseppeg: Just use CSS comments `{/* */}` is invalid css and might break very easily

lib/components/terms.js
```diff
@@ -131,39 +130,40 @@ export default class Terms extends Component {
...
+        <style jsx>{`
+          .terms_terms {
+            position: absolute;
+            margin-top: 34px;
+            top: 0;
+            right: 0;
+            left: 0;
+            bottom: 0;
+            color: #fff;
+            transition: ${isMac ? '' : 'margin-top 0.3s ease'};
```
giuseppeg: probably `transition: ${isMac ? 'none' : 'margin-top 0.3s ease'};` is better

lib/components/split-pane.js
giuseppeg: need to fix all over the PR

lib/components/header.js
```diff
@@ -109,30 +108,42 @@ export default class Header extends PureComponent {
...
-                <svg className={css('shape', left && 'maximizeWindowLeft')} onClick={this.handleMaximizeClick}>
+                <svg
+                  className={`header_shape ${left && 'header_maximizeWindowLeft'}`}
+                  onClick={this.handleMaximizeClick}
+                >
                   <use xlinkHref={maxButtonHref} />
                 </svg>
-                <svg className={css('shape', 'closeWindow', left && 'closeWindowLeft')} onClick={this.handleCloseClick}>
+                <svg
+                  className={`header_shape header_closeWindow ${left && 'header_closeWindowLeft'}`}
```
albinekb: won't this add the class name `.false` to this?
image

maybe use https://github.com/JedWatson/classnames or similar 🤔

lib/components/header.js
albinekb: or https://github.com/JedWatson/classnames which helps a lot, and makes it cleaner for when you have more than one dynamic class name 👍 Also helps with consistency

ppot: `header_header` sound like a bad name. You are already inside `Header` component. Since you are at the root className of the component. You could leave it `header` and then `header_rounded`

lib/components/split-pane.js
```diff
@@ -116,7 +115,7 @@ export default class SplitPane extends PureComponent {
       sizes = new Array(children.length).fill(1 / children.length);
     }
     return (
-      <div className={css('panes', `panes_${direction}`)}>
+      <div className={`splitpane_panes splitpane_panes_${direction}`}>
```
ppot: `splitpane_${direction}`

ppot:
It's a good start. My only concern is with the `class name`, the nomenclature can be revisited.

lib/components/header.js
timothyis: Thanks for spotting that @giuseppeg. I changed to a ternary in a few places but I obviously missed a lot somehow.

@ppot I didn't make those names up, that's the names that we were using before but they were just generated from `base_components`. I did speak about that in the PR description.
Changing these classnames would break users custom CSS, themes, and some plugins.

lib/components/split-pane.js
timothyis: I'll fix this 👌

lib/components/header.js
timothyis: I'm not sure we have enough of this case to use classnames. But could do it in another PR later.
I'll just change them to ternaries for now

lib/components/split-pane.js
timothyis: As I said about the other classnames, I didn't make it up. It's what has to be used to keep consistency and backwards compatibility.

lib/components/header.js
ppot: Yeah. That's why my comment at the end: `My only concern is with the names ` haha  ;)

ppot: It is something we need to revisit for the codebase. But for now it's okay to leave it like that.

timothyis:
@giuseppeg @albinekb Feedback addressed 👌

Stanzilla:
Looks and seems to work fine on Windows 👍

giuseppeg:
Nice! I am not familiar with the codebase so I am the last one who should give you a shipit. If I were you I would take this as a chance to add some visual diffing test to catch regressions.

giuseppeg:
fwiw you can take screenshots with electron - open the dev tools in hyper and paste this snippet

```js
require("electron")
  .remote.getCurrentWindow()
  .capturePage(function(data) {
    const buffer = data.toJPEG(100)
    const blob = new Blob([buffer], { type: "image/jpeg" })
    const objectUrl = window.URL.createObjectURL(blob)
    const a = document.createElement("a")
    a.href = objectUrl
    a.style = "display: none"
    a.download = "test.jpeg"
    document.body.appendChild(a)
    a.click()
    window.URL.revokeObjectURL(objectUrl)
    document.body.removeChild(a)
  })
```

chabou:
Amazing job @timothyis!!
Thank you so much 😍

PR #2767, opened 2018-03-17
Disable VSCode type checking

chabou:
Type checking was introduced here: #2272

But there are some confusing errors.
If someone wants to re-enable this, please fix errors too

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,6 +1,5 @@
 {
   "compilerOptions": {
-    "checkJs": true,
     "jsx": "react",
     "target": "es6"
   },


albinekb:
Why? @chabou I don't think turning it off is the correct solution..

chabou:
I don't have time to correct these errors (some false positive). I constantly disable it when I create a branch from canary and pay attention to not commit `jsconfig.json` file.

We made an error to accept the previous PR without resolving found errors.

Feel free to reactive this type checking with all error resolved and, if possible a Husky hook to prevent errors to be committed.

Tyriar:
It might be better to do this either at the user settings level:

```
"javascript.implicitProjectConfig.checkJs": true
```

Or the file level

```
// @ts-check
```

PR #2771, opened 2018-03-18
Fix Terminal colors hot reloading

chabou:
Before this PR, Terminal theme colors (background, foreground, selection, ANSI, cursor etc...) were never hot reloaded.

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -240,9 +240,9 @@ export default class Term extends React.PureComponent {
     // Do we need to update theme?
     const shouldUpdateTheme =
       !this.termOptions.theme ||
-      Object.keys(nextTermOptions.theme).some(option => {
-        nextTermOptions.theme[option] !== this.termOptions.theme[option];
-      });
+      Object.keys(nextTermOptions.theme).some(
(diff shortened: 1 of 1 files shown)


timothyis:
This seems to work great on install/uninstall of a theme/plugin but when I hard-refresh the app reverts to the themes background colour (even after uninstall). Xterm doesn't though.

image

chabou:
Can you give me exact steps to reproduce your issue?

timothyis:
@chabou 
1. `hyper i hyper-snazzy`
2. hard-refresh
3. Notice how the background colour reverts its state.

Not sure if this is the biggest issue, though.

chabou:
Ok reproduced.
Frankly, hard-refresh is totally broken. I think we need to respawn a window (but with keeping/restore state) and remove soft refresh.

Stanzilla:
Can repro as well.

timothyis:
@chabou Let's merge this in then? hard-refresh is a different issue and we should remove or redefine refresh. 

Hard-refresh: App reload
Refresh: All terms reset to their initial state
How does that sound?

chabou:
agreed

PR #2774, opened 2018-03-20
Add SESSION_WILL_REQUEST and TAB_GROUP_WILL_REQUEST actions

hharnisc:
**I think this PR is ready to merge**

This fires two new actions `SESSION_WILL_REQUEST` and `TAB_GROUP_WILL_REQUEST` the provides a hook for tools like [hypercwd](https://github.com/hharnisc/hypercwd) to set the current working directory before creating a new session/tab group. This removes the need for having to check the processes `pwd` on (nearly every keystroke -- for OS X + Linux).

I'm working on a canary release of `hypercwd` to consume these actions.


hharnisc:
Just published a canary release (this version of `hypercwd` is untested on windows -- I'll be testing that out soon but doesn't impact this PR)

```js
 plugins: [
    'hypercwd#1.2.2-canary-01'
  ],
```

chabou:
@hharnisc Hi!
This is a really good news that `hypercwd` will be compatible with our upcoming v2 🎉.

I think you can do exactly the same thing without adding these 2 new action types.
I see you already intercept them in your middleware: https://github.com/hharnisc/hypercwd/blob/9b9ebe5a4496eb6635d5df17f351b6c446502a3a/linuxActionHandler.js#L25-L31

But I wonder why you don't intercept `SESSION_REQUEST`/`TERM_GROUP_REQUEST` directly?
If you pass these actions to `next()` handler only after awaiting current directory getter, and change `cwd` action parameter on the fly, it should work. But I certainly miss something.

I can really help you on this, I really love and NEED your plugin. Do not hesitate to ask 🙏

hharnisc:
So cool to hear you use `hypercwd` @chabou :smiley: :raised_hands: 

Would you be up for sharing a snippet of intercepting the `SESSION_REQUEST` / `TERM_GROUP_REQUEST` actions? I'm all for using existing actions if possible since that might make the `hypercwd` changes backwards compatible :smile: 

The reason why I _think_ we might need these new actions is because the cwd is retrieved before the `SESSION_REQUEST` is sent. Then the new session is emitted with the old `cwd`.

Let's work together on this to get the best solution :raised_hands:

chabou:
OK you're right, `cwd` is evaluated too early and prevent plugins to change it.

I made a PR to correct this: #2777

chabou:
With #2777, you just have to remove `WILL_` from action name here: https://github.com/hharnisc/hypercwd/blob/9b9ebe5a4496eb6635d5df17f351b6c446502a3a/linuxActionHandler.js#L25-L26

And it just works 🎉

hharnisc:
Thanks so much @chabou :raised_hands: I'll close this PR. I've published a new canary release of `hypercwd` to reflect these changes:

```
 plugins: [
    'hypercwd#1.2.2-canary-02'
 ],
```

I'll keep an eye out for the canary of `hyper` to be officially released -- then I can push `hypercwd` canary -> master

Looking forward to this amazing release, its so smooth! :rocket:

chabou:
@hharnisc This is time to push ;)

PR #2777, opened 2018-03-21
Make SESSION_REQUEST/TERM_GROUP_REQUEST hackable

chabou:
`cwd` should be lazily evaluated to let plugins override it

This PR should replace #2774


hharnisc:
Tested this out and works great :+1: thanks so much @chabou

PR #2787, opened 2018-03-26
Prevent build job when releasing

chabou:
Running `build` job on tags because `release` job will do the same


chabou:
I think we don't need these artefact because they are automatically uploaded to Github Release

PR #2792, opened 2018-03-28
CLI: use .hyper.js from repo root dir when running in dev mode

tool3:
Ready to be merged automatically.

I went with process.env.NODE_ENV because it is set automatically to 'production' when running in..well..production.

Thank you so much ! if I missed anything be sure to let me know please :)

1st contribution!


chabou:
Thank you for this first contribution 🙏 

But some part must be modified:

We don't need a new config file. We need to have exactly same behavior with CLI than with Hyper. If it is in dev mode, it checks the existence of config file `.hyper.js` in root repository directory. If it exists, it uses this and not the one in user home directory

tool3:
Please see latest commit

tool3:
oops... sorry about that 👼 ..please see latest commit

cli/api.js
```diff
@@ -5,7 +5,12 @@ const registryUrl = require('registry-url')();
 const pify = require('pify');
 const recast = require('recast');
 
-const fileName = `${os.homedir()}/.hyper.js`;
+const devConfigFileName = '../.hyper.js';
```
chabou: You need to resolve this path with something like [__dirname](https://nodejs.org/docs/latest/api/modules.html#modules_dirname). It should not be impacted by the current working directory.

tool3:
Used `__dirname` with `path.join` to be relative to the root directory. Please check 🤞

PR #2795, opened 2018-03-30
always set js to lf

Stanzilla:
if on auto, Windows will change them to crlf locally and this can make lint fail

@paulcbetts thoughts?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
-* text=lf
-*.sh eol=lf
+* text=auto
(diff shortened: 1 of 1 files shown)


sindresorhus:
This will corrupt images, as you're now saying that Git should convert all files to `lf`, even binary.

What you want is:

```
* text=auto
*.js text eol=lf
```

I wish Git supported a way to say "make all *text* files LF"…

Stanzilla:
Oh thanks!

PR #2797, opened 2018-03-31
Update readme

mapitman:
Add info about installing in Arch Linux from the AUR
Add section headers for each OS in the Usage section


chabou:
Thank you for your help 🙏

PR #2811, opened 2018-04-05
Use xterm winptyCompat addon

chabou:
See https://github.com/xtermjs/xterm.js/pull/1096

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -3,6 +3,7 @@ import React from 'react';
 import {Terminal} from 'xterm';
 import * as fit from 'xterm/lib/addons/fit/fit';
 import * as webLinks from 'xterm/lib/addons/webLinks/webLinks';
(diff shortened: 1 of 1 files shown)


lib/components/term.js
```diff
@@ -3,13 +3,15 @@ import React from 'react';
 import {Terminal} from 'xterm';
 import * as fit from 'xterm/lib/addons/fit/fit';
 import * as webLinks from 'xterm/lib/addons/webLinks/webLinks';
+import * as winptyCompat from 'xterm/lib/addons/winptyCompat';
```
timothyis: Does this not need `xterm/lib/addons/winptyCompat'/winptyCompat` to link to a certain file, like with webLinks?

chabou: 😮 Good catch! Thank you!

chabou:
Tested on Windows10 (cmd) and everything seems OK

PR #2824, opened 2018-04-16
Removed website

leo:
It's now located in [zeit/hyper-site](https://github.com/zeit/hyper-site).


albinekb:
zeit/hyper-site is not public, but I assume it will be opened up? 😊

leo:
@albinekb Correct! 👍 

@timothyis just mentioned it would be good to include this information in the PR template, so I guess we need to wait before merging this until they've adapted the PR.

PR #2825, opened 2018-04-16
New timestamp server

leo:
This makes our Windows builds work again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -148,7 +148,8 @@
       ]
     },
     "win": {
-      "target": ["squirrel"]
+      "target": ["squirrel"],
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -148,7 +148,8 @@
       ]
     },
     "win": {
-      "target": ["squirrel"]
+      "target": ["squirrel"],
+      "timeStampServer": "http://timestamp.globalsign.com/scripts/timstamp.dll"
```
albinekb: Hmm, I don't like that this is http 😱 but it doesn't seem to serve over https 😭

Issue #2850, opened 2018-04-17
Configure line height


hankchizljaw:
- [X] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [X] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: macOS High Sierra 10.13.3
- **Hyper.app version**: Hyper 2.0.0 (stable)
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/hankchizljaw/d1733306ae06854a6ae9d1304bbd0844

## Issue
Hey Hyper Team, 

First of all, the new version looks excellent, so thank you for your ongoing hard work on this 👏 

Now that we've got a canvas based rendering system, is it any simpler to increase the line height at all? 

Here's two screenshots of Vim, the first using iTerm and the second user Hyper 2.0

### iTerm
<img width="1552" alt="screen shot 2018-04-17 at 10 49 04" src="https://user-images.githubusercontent.com/8672583/38862535-19ed852e-422d-11e8-958b-2d82ba9608d0.png">

### Hyper 
![screen shot 2018-04-17 at 10 49 41](https://user-images.githubusercontent.com/8672583/38862548-224a6da4-422d-11e8-9463-517cae7db46d.png)

msikma:
As mentioned by @Spaxe in #75, xterm 3.0 [exposes a `lineHeight` option.](https://github.com/xtermjs/xterm.js/pull/938) It doesn't look like it's being passed on, though, since defining it in the config doesn't do anything.

hankchizljaw:
#2858 looks like a decent solution. Let's hope this is resolved soon, as I'd love to use Hyper as my primary terminal 🙂

albinekb:
@piotr-yuxuan because it's not used by a lot of people. Feel free to create a PR. Now we have #2858 so it will be in soon. We're all doing this in our **spare** time and for **free**.

piotr-yuxuan:
Please forgive the tone of my previous message which was too cold. I know we all contribute for free in our spare time :-)

PR #2858, opened 2018-04-17
Add line height config option

### LINKED ISSUE #75: Add `line-height` configuration parameter

bdougherty:
Add a `lineHeight` configuration parameter. Not sure if what I did is the appropriate way to check that the type is correct, so please let me know if there is a different way that it's normally done and I will update it.

This will fix #75 and #2850.


chabou:
Awesome! Thank you! I'll merge it ASAP

lib/components/term.js
```diff
@@ -263,7 +264,10 @@ export default class Term extends React.PureComponent {
       });
     }
 
-    if (this.props.fontSize !== nextProps.fontSize || this.props.fontFamily !== nextProps.fontFamily) {
+    const fontSizeChanged = this.props.fontSize !== nextProps.fontSize;
+    const fontFamilyChanged = this.props.fontFamily !== nextProps.fontFamily;
+    const lineHeightChanged = this.props.lineHeight !== nextProps.lineHeight;
+    if (fontSizeChanged || fontFamilyChanged || lineHeightChanged) {
```
chabou: I think that storing comparison result in temp vars is not necessary.
These comparisons don't need some explanation with explicit variable names. And without these temporary variables, comparisons will be done only if previous one is invalid.

bdougherty: Updated.

hankchizljaw:
I've built the canary build and this is working really well. Thanks for the hard work @bdougherty!

dongz9:
@chabou I prefer to keep the cursor size unchanged. Do you think it's reasonable to make it configurable?

Baevra:
Line height is awesome!!!
But now we need cursor height option!

JesseLeite:
Yall crazy, love the tall cursor.  Wouldn't a shorter cursor look funny next to background highlighted text?

albinekb:
No, it's not in stable yet @Azzhoe https://github.com/zeit/hyper/releases/tag/2.1.0-canary.1

rchrdnsh:
Any update on this? When will 2.1.0 be stable?

Tuxie:
I tried canary.2 but I was not able to set a lineHeight below 1.0. Is this a bug or a technical limitation? I want to _reduce_ line height because my favourite font has too much spacing. In iTerm2 I have set line height to 85% and it looks perfect.

PR #2901, opened 2018-04-22
Moved version from app/package.json to root

### LINKED ISSUE #2894: Move app version string from app/package.json to root package.json
Having version number in root `package.json` will permit to use the full power of https://github.com/zeit/release

Manish-Giri:
Moved `version` from `app/package.json` to root `package.json`. 

Tested with `yarn run dev` and `yarn run app`, although this change shouldn't break anything - 

hyp2

Fixes #2894

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -2,7 +2,6 @@
   "name": "hyper",
   "productName": "Hyper",
   "description": "A terminal built on web technologies",
-  "version": "2.0.0-canary.18",
   "license": "MIT",
   "author": {
     "name": "ZEIT, Inc.",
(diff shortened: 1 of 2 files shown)


Manish-Giri:
Oops, completely overlooked that `version` is mandatory in `package.json`. Should I leave the original version in `app/package.json` as is?

timothyis:
@Manish-Giri I think it should be fine unless it causes any errors? The top-level package.json didn't have a version before, so it shouldn't matter that the other doesn't have a version now.

chabou:
You should pay attention on 2 things:
* Code that requires this version like: https://github.com/zeit/hyper/blob/2de45245bf46484daef1efdd070f2a8b6329515a/app/notifications.js#L4
* Electron builder: when executing `yarn dist`

timneutkens:
I'm going to close this since the PR is stale.

Issue #785, opened 2016-10-06
Allow window memorization upon forced application close


ppot:
The expected behavior would be to save application screens positions and size and reopen the next time the terminal is opened.

### use case
1. app quit from tray
2. app quit
3.  state restore after forced shutdown / restart
4. hyperterm update

#### Feel free to comment different uses cases.

ppot:
@ekmartin @matheuss @rauchg Do you think of other uses cases?

PR #2914, opened 2018-04-23
Remember the window size on close and launch accordingly

Stanzilla:
This is https://github.com/zeit/hyper/pull/2382 rebased. electron-window-state-manager afaik is the de-facto standard for this in Electron apps so it should be fine. Tested on Windows 10, works well so far

fixes https://github.com/zeit/hyper/issues/785

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -55,6 +55,7 @@ const {resolve} = require('path');
 const {app, BrowserWindow, Menu} = require('electron');
 const {gitDescribe} = require('git-describe');
 const isDev = require('electron-is-dev');
+const WindowStateManager = require('electron-window-state-manager');
 
 const config = require('./config');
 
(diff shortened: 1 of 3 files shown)


Stanzilla:
hyper_2018-04-24_01-04-51
^ default size (1024x768)

Stanzilla:
Since we're in a widescreen dominated world and lots of people use 13" with pretty low resolutions, 480p (1.85:1) / 888×480p could be a nice alternative as well:

hyper_2018-04-24_01-13-01

Stanzilla:
Yes we should, even if it only is to be consistent.

Stanzilla:
That's https://github.com/zeit/hyper/pull/2510 then

zanona:
Nice, is there any way to make Hyper always honour `windowSize`, so regardless the final state before closing it will always start using those settings? Thanks

PR #2938, opened 2018-04-26
Return of the Bell

ShaggyDev:
Hey all!

The purpose of this PR is to restore the "bell" sound toggle on/off from the config file's "bell" field. I have been using Hyper.js again for a bit and noticed this missing feature. For me its extremely helpful when trying to determine if the terminal session lost connection or not (there is some lagging sometimes that without the bell, it is difficult to determine if the terminal window is lagging or there is no action to take). While in the part of the code, I took the liberty at re-enabling the ability to change the bell sound using the "bellSoundURL" field in the config file.

I am unfamiliar with the previous terminal emulator used, but It would seem the root cause for requiring these changes is due to the switch to xterm.js. Xterm.js' configuration for the bell sound currently has two options: 'none' and 'sound'. We only needed to map Hyper.js' "SOUND" value to correlate with Xterm.js' "bellType" field. To change the sound took most of the effort in this PR.

Xterm.js uses a base64 encoded binary string of the audio file data in the "bellSound" configuration field. This encoded string is appended to "base64," to create a simple comma-delimited string like so: "base64,{encodedStringHere}". When xterm.js looks at this field, they split the string by the comma and use the second value, being the encoded string, so we play nicely with xterm and construct the string as they would be used to seeing it as. As long as we can get the binary data of an audio file, it doesn't matter where it comes from, per se, so I added functionality to read a local audio file specified by its absolute file path in the filesystem as well as use a web link to a sound file.

Barring any major issues you all might find, these changes are ready to be merged. 

Let me know if there are any issues you find that I should address. Thank you for your time!


lib/components/term.js
```diff
@@ -61,6 +65,60 @@ const getTermOptions = props => {
   };
 };
 
+const getFileData = soundFileLocation => {
+  return new Promise(async resolve => {
+    const isRemoteUrl = isValidUrl(soundFileLocation);
```
albinekb: One problem I see with accepting remote URLs is that your terminal won't have any sound if you start without internet access?

We could cache it and add more logic, but I think it's best just to only accept local files and let the user download it manually? This is a very novel feature that not many will use but it can cause a lot of issues.

lib/components/term.js
```diff
@@ -234,12 +292,20 @@ export default class Term extends React.PureComponent {
...
-  componentWillReceiveProps(nextProps) {
+  async componentWillReceiveProps(nextProps) {
     if (!this.props.cleared && nextProps.cleared) {
       this.clear();
     }
+
     const nextTermOptions = getTermOptions(nextProps);
 
+    if (nextProps.bellSoundURL) {
+      const value = await getFileData(nextProps.bellSoundURL);
```
chabou: I think it should not be done here.
It will be done each time that you open a new pane/tab/window.

Maybe better to do this in the main process but I don't know how to do this fetch asynchronously. Maybe by reemitting a `config change` RPC event when done, to dispatch a new `CONFIG_RELOAD` redux action in each window.

It will maybe require adding some caching mechanism when decorating configuration.

lib/components/term.js
```diff
@@ -234,12 +292,20 @@ export default class Term extends React.PureComponent {
...
+      if (value) {
+        nextTermOptions.bellSound = value;
```
chabou: `bellSound` option seems unused 🤔

lib/components/term.js
dongreenmedl: Thats a good point about using remote urls and lack of internet at the time of spawning the terminal. Maybe for "restoring functionality" we can stick with only local files.

lib/reducers/ui.js
```diff
@@ -86,7 +86,7 @@ const initial = Immutable({
   messageURL: null,
   messageDismissable: null,
   bell: 'SOUND',
-  bellSoundURL: 'lib-resource:hterm/audio/bell',
+  // bellSoundURL: 'lib-resource:hterm/audio/bell',
```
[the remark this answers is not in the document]
dongreenmedl: Good point, ill have to comb through better when I circle back on things like this.

lib/components/term.js
albinekb: yes, also it would slow down startup significantly to download the sound every launch 👍  I don't think hyper should support remote URLs for this

lib/components/term.js
dongreenmedl: I imagine it isn't the most important feature out of everything, but the default sound provided by Xterm only makes the feature less desirable 😆. The aim was a mix between personal desire (I like having a bell sound) and completeness. I wanted to remove one more thing for someone to complain about 😀

chabou: To be clear: I LOVE your PR. Seriously, it will be useful and magnify what Hyper is: configurable and Hackable.

What I was trying to say is: You assign `value` to `nextTermOptions.bellSound` but it doesn't seem to be used anywhere. I think that you forgot to add it to Terminal options object like you did for `bellStyle`.

lib/components/term.js
chabou: Loading remote sound can be done async. So it would not slow down startup but there is a lot of case to deal with: during loading, after a failure etc...

I agree it could be better to stick with local files only.

lib/components/term.js
dongreenmedl: Thank you! It's not often I find time to contribute to open source and figured this was an interesting and bite-sized task to handle (bonus points for taking a task that hasn't been reported yet 😆). So thank you for taking the time to evaluate this PR; it's quite motivating as I look forward to contributing where I can.

Okay, so it is a little ambiguous on when it gets used. In the lines just below, `Object.keys` ends up evaluating each option in `nextTermOptions` and filters out only those that have changed which will then call the `setOption` method on the terminal for each option that has changed. 

I originally was going to place the logic in the same area as the `bellStyle` change, but it would end up modifying the nature of `getTermOptions` into an asynchronous function which would create a bit of noise in all the areas `getTermOptions` is being called. I was looking for the least disruptive change, but it's more important that it is done right for the project so ill dig deeper to find where it would fit best. 

Considering @albinekb comments, there would be a way of utilizing `fs.readFileSync()` and drop the "remote url" feature due to valid concerns raised in that idea. On the other hand, `fs.readFileSync()` is usually frowned upon and I don't generally gravitate towards. I do want to take a look at optimization too so we would only have to read the file once, even from local file system, and have this be set for all terminal sessions.

albinekb: Sounds great 👍

ShaggyDev:
Hey all, I pushed some updates to this PR that hopefully addresses all the topics we discussed before. Let me know what you all think about it ;D

lib/reducers/ui.js
```diff
@@ -195,7 +196,7 @@ const reducer = (state = initial, action) => {
               ret.bell = config.bell;
             }
 
-            if (config.bellSoundURL) {
+            if (config.bellSoundURL !== state.bellSoundURL) {
```
chabou: I think this is not necessary to compare with value in the state.

lib/constants/config.js
```diff
@@ -1,2 +1,7 @@
 export const CONFIG_LOAD = 'CONFIG_LOAD';
 export const CONFIG_RELOAD = 'CONFIG_RELOAD';
+export const CONFIG_ASYNC = 'CONFIG_ASYNC';
+
+export const BELL_STYLE = {
```
chabou: I think this is not necessary. Or we should use it everywhere. Like here: https://github.com/zeit/hyper/pull/2938/files#diff-7da6d687f5a3beed5725bb8324e4b1adR26
or even here: https://github.com/zeit/hyper/pull/2938/files#diff-6d186b954a58d5bb740f73d84fe39073R32

But using 'SOUND' directly is ok imo.

lib/components/term.js
```diff
@@ -86,6 +89,7 @@ export default class Term extends React.PureComponent {
     this.term.open(this.termRef);
     this.term.webLinksInit();
     this.term.winptyCompatInit();
+    this.termDefaultBellSound = this.term.bellSound;
```
chabou: Are you sure that it works?
I don't see anything in `Terminal.ts`: https://github.com/xtermjs/xterm.js/blob/master/src/Terminal.ts

lib/utils/file.js
```diff
@@ -9,11 +9,29 @@
...
-export default function isExecutable(fileStat) {
+export function isExecutable(fileStat) {
   if (process.platform === 'win32') {
     return true;
   }
 
   return Boolean(fileStat.mode & 0o0001 || fileStat.mode & 0o0010 || fileStat.mode & 0o0100);
 }
+
+export const getFileDataAsync = soundFileLocation => {
```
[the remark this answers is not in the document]
ShaggyDev: The desired file type would be that of a sound file which is why it is so specific. It might even be worth evaluating the extension or file type somehow to potentially prevent reading a file we know we cant use for a bell sound? If desired, we can use a more generic name here.

lib/reducers/ui.js
ShaggyDev: This one does have a specific reason as i initially thought the same before I implemented this. Without comparing against the current state, removing the `bellSoundURL` property from the config file would not properly restore the sound back to a default sound. Instead of only updating the bellSoundURL field if the value exists, we check if it has changed and updated accordingly.

lib/components/term.js
ShaggyDev: I did confirm and test throughly that this indeed works. Within `Terminal.ts` there is a constant variable declared toward the beginning of the file called `DEFAULT_OPTIONS` which contains a property `bellSound` with the value `DEFAULT_BELL_SOUND` which is imported from the xterm packages `SoundManager` class. 

I needed to offload the defaultSound value from the term instance before the method `this.term.setOption()` is called to overwrite the bellSound property so that we can restore it later. Xterm.js does not currently have logic to restore its own default value upon the bellSound being overwritten then being null or undefined later (it will just fall back on whatever was set last).

lib/utils/array.js
```diff
@@ -1,3 +1,21 @@
-export default function last(arr) {
+export function last(arr) {
   return arr[arr.length - 1];
 }
+
+export const arrBuf2Base64 = buffer => {
+  let base64Data = null;
+  try {
+    const binaryString = String.fromCharCode.apply(null, new Uint8Array(buffer));
+    base64Data = 'base64,' + window.btoa(binaryString);
```
chabou: Better to use the [Buffer class](https://nodejs.org/docs/latest/api/buffer.html#buffer_buf_tostring_encoding_start_end), imo.
Something like: 
```
base64Data = Buffer.from(buffer).toString('base64')
```

lib/utils/array.js
```diff
@@ -1,3 +1,21 @@
...
+  } catch (err) {
+    if (err instanceof RangeError) {
+      //eslint-disable-next-line no-console
+      console.warn('The file was liekly too large to use for the "bell" sound:\n', err);
```
chabou: Typo: `liekly`

lib/utils/file.js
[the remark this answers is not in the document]
albinekb: I think this function is more generic than just for sound files, so either rename it to `filePath`, and implement the sound checks in another function, **or** rename this function to `getSoundDataAsync`?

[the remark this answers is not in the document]
chabou: I think that we need to make this function less specific:

We have to export a `getBase64File` function or something like that. And merge `arrBuf2Base64` (updated to use Buffers) code into it.
This function will be responsible to load a filePath asynchronously and returns a base64 string from its content.
Adding `base64,` prefix should be done after using this function because this is really specific and will prevent to reuse this function.

I think that we don't need to make any filetype verification.

lib/reducers/ui.js
```diff
@@ -234,6 +235,13 @@ const reducer = (state = initial, action) => {
         );
       break;
 
+    case CONFIG_ASYNC:
```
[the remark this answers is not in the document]
albinekb: I agree, if we can use the already existing `CONFIG_RELOAD`, that's easier to maintain 👍

lib/utils/file.js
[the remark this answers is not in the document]
ShaggyDev: I see what you all mean now and agree that this should be more generic for better reusability. Ill get this updated with the feedback in mind. @chabou Thank you, btw, for the feedback on using the Buffer class to create the base64 string. I felt that there should be a better way of achieving the same result but decided to begin with a similar implementation to the one being used in the `SoundManager.ts` file in the xterm.js project to mitigate bugs caused by mismatched methods and results.

ShaggyDev:
@chabou @albinekb Thank you guys for looking over this again ;D. My next steps will be:

1. Make the method used to convert file data into a base64 string more generic in naming and in implementation (ie. moving the bit that appends the 'base64' string to the converted string).
2. Attempt to remove the new CONFIG_ASYNC action and instead re-use the CONFIG_RELOAD action and reducer.
3. Fix a typo in one of the console warnings when getting the `RangeError` exception.

If you guys see anything here I might have missed, feel free to remind me ;D. Thanks again for all your helpful feedback

lib/utils/file.js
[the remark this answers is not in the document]
chabou: There is a huge difference between xtermjs SoundManager and your function: yours is executed on Electron MainProcess (ie nodejs, not a browser). And Buffer class is only available in NodeJs.

chabou:
Keep up your hard work.

lib/utils/array.js
```diff
@@ -1,3 +1,15 @@
-export default function last(arr) {
+export function last(arr) {
   return arr[arr.length - 1];
 }
+
+export const arrBuf2Base64 = buffer => {
+  let base64String = null;
+  try {
+    base64String = Buffer.from(buffer).toString('base64');
```
albinekb: you don't need to assign this to a temp var, just do

albinekb: you don't need to assign this to a temp var, just do:

albinekb: you don't need to assign this to a temp var, just do:
```js
return Buffer.from(buffer).toString('base64');
```

lib/utils/array.js
```diff
@@ -1,3 +1,15 @@
...
+  } catch (err) {
+    //eslint-disable-next-line no-console
+    console.error('Could not use the file specified:', err);
```
albinekb: Should we really catch this error here? Isn't it better to re-throw it and handle it in the caller to display the error to the user? 🤔

ShaggyDev: Ah yea you have a point. I could just have the getBase64FileData function use the try catch and resolve with null if it catches an error. Taking it a bit further, does it make sense anymore to wrap `Buffer.from(buffer).toString('base64')` in arrBuf2Base64 anymore? It has turned into a one-liner thing now so technically the `getBase64FileData` can import the Buffer class and perform the one-liner. Thoughts?

albinekb: Hmm this should never throw anyways, if you send a buffer or string to `Buffer.from` it won't throw (afaik), so maybe it's not even needed to try/catch it if you know you got a valid buffer/string?

ShaggyDev:
@chabou @albinekb Just letting you guys know, I just pushed some small updates that I believe should address the minor inefficiencies as well as a sneaky bug that would cause the sound to return to default when saving a config file that does not change the bellSoundURL. Let me know if you see anything that might need more adjustment/fixing. Thank you for your patience ;D

lib/utils/array.js
```diff
@@ -1,3 +1,3 @@
-export default function last(arr) {
+export function last(arr) {
```
albinekb: Hmm, wasn't this used anywhere? I mean changing from default export to named export 🤔

lib/actions/ui.js
```diff
@@ -1,6 +1,6 @@
 import {php_escapeshellcmd as escapeShellCmd} from 'php-escape-shell';
-import last from '../utils/array';
-import isExecutable from '../utils/file';
+import {last} from '../utils/array';
```
albinekb: Only used here? (see comment further down)

chabou: You mean that we have a module with only one function with only one line `arr[arr.length - 1]` imported in only one file and executed only one time: https://github.com/ShaggyDev/hyper/blob/d868721a0661b42d39be3eda086e0fd0d3342561/lib/actions/ui.js#L168

😂 

Ok, we need to remove this `utils/array.js` module

chabou:
This PR needs a final test round.
I'll do ASAP.

MichelYpma:
Any update on this? Can't wait to annoy some colleagues with my custom terminal bell :)

ShaggyDev:
Merged in the latest from this canary branch and retested functionality. Appears that all is working as expected, although I currently don't have the ability to test this in a windows environment. Should there be any problems, let me know and ill take a look ;D.

estebanrao:
Hey everyone! Is this still being planned to be merged? I'm on Mac and the bell hasn't worked for a while now.

tplk:
@chabou Can we have an update on this? I wanted to use hyper as my main term on windows but the bell is really annoying. Thanks in advance.

pelillian:
@chabou What is the status of this PR? It would be great to have the bell working on Mac.

benc-uk:
Switched to canary and I miss my bell 
🚫 🔔 😭

ShaggyDev:
Updated this to the latest canary version to solve the merge conflicts and made some adjustments to fix some regression issues. This should be good to go again.

On another note, I'm not fully up to speed on how releases are organized for this product so please let me know if anything I am doing here is pushing this further away from being merged in.

Happy Hacking all ;D

lib/components/term-group.js
```diff
@@ -82,6 +82,7 @@ class TermGroup_ extends React.PureComponent {
       copyOnSelect: this.props.copyOnSelect,
       bell: this.props.bell,
       bellSoundURL: this.props.bellSoundURL,
```
ppot: This might not be required anymore. Try to look if you can catch this.

ppot: Look like it's in the config for a giving specific sounds. Nevermind.

lib/components/term.js
```diff
@@ -267,8 +271,15 @@ export default class Term extends React.PureComponent {
     if (!this.props.cleared && nextProps.cleared) {
       this.clear();
     }
+
     const nextTermOptions = getTermOptions(nextProps);
 
+    // Check nextProps for the bellSoundURL field explicitly to see if there is a value
```
ppot: `Check bellSoundURL value  in  nextProps`

lib/components/term.js
```diff
@@ -267,8 +271,15 @@ export default class Term extends React.PureComponent {
...
+    // Only if there is a value do we want the field "bellSound" to exist in extTermOptions
```
ppot: `If found, then let "bellSound" exist in extTermOptions`

lib/components/term.js
```diff
@@ -267,8 +271,15 @@ export default class Term extends React.PureComponent {
...
+    // If "bellSound" is set to null or undefined, the below logic will call "this.term.setOptions" with
```
ppot: `If null or undefined, call "this.term.setOptions"`

lib/components/term.js
```diff
@@ -267,8 +271,15 @@ export default class Term extends React.PureComponent {
...
+    // the undesired value of null or undefined which causes xterm.js to fail trying to play the bell sound.
+    nextTermOptions.bellSound = nextProps.bellSound || this.termDefaultBellSound;
```
ppot: Try to be more direct in your comment if possible. Less duplication and noice.

lib/constants/config.js
```diff
@@ -1,2 +1,6 @@
 export const CONFIG_LOAD = 'CONFIG_LOAD';
 export const CONFIG_RELOAD = 'CONFIG_RELOAD';
+
+export const BELL_STYLE = {
```
ppot: Put this directly in the fil. it's not reused.

dongreenmedl: @ppot You are right, currently it is not being reused anywhere else. The only thought I had was "maybe" there would be a time where other options would be available, such as "VISUAL" for a visual bell style. It seems out of scope for this iteration so we can remove this and have the condition that is using this just hardcode the string in the conditional instead of referencing this object enum.

lib/components/term.js
dongreenmedl: @ppot Just so I understand, these comments are suggestions on making the comment notes more concise? I want to make sure I'm not missing anything else.

Thank you for taking the time to review and add your feedback btw!

lib/constants/config.js
ShaggyDev: @ppot You are right, currently it is not being reused anywhere else. The only thought I had was "maybe" there would be a time where other options would be available, such as "VISUAL" for a visual bell style. It seems out of scope for this iteration so we can remove this and have the condition that is using this just hardcode the string in the conditional instead of referencing this object enum.

lib/components/term.js
ShaggyDev: @ppot Just so I understand, these comments are suggestions on making the comment notes more concise? I want to make sure I'm not missing anything else.

Thank you for taking the time to review and add your feedback btw!

Stanzilla:
@ShaggyDev hey there, sorry for the super long delay, are you willing to pick this up again and address the outstanding comments? If they no longer apply you can click on `Resolve Conversation` to mark them as fixed!

estebanrao:
Hey @Stanzilla. The issue still persists, version Hyper 3.0.2 (stable) on macOS 10.14.6. Thanks

Stanzilla:
@estebanrao yeah, waiting to hear back from @ShaggyDev

ShaggyDev:
@Stanzilla The latest commit address the feedback in the comments. I marked those conversations as resolved. I apologize for not marking those earlier lol.

Please let me know if there's anything else you'd like me to address.

Stanzilla:
Hey there, it mainly needs a rebase, but I still see a few open comments

ShaggyDev:
@Stanzilla Ok, ill rebase the latest and solve the conflicts as well as run through the comments again and resolve those that have been addressed.

Thank you for taking a look at this again; I'm happy to see activity on this again ;D.

ShaggyDev:
@Stanzilla Just FYI, I should be taking care of this in the next day or so. I started on this last week to check out what it was going to take to merge and noticed there was some refactoring into typescript on one of these files. I ran into some issues when I tried to compile and run the project. Maybe some recent commits already solved this but ill have more details soon.

Stanzilla:
Thank you and sorry that it took so long!

Stanzilla:
@ShaggyDev hey there, I just had this happen 

```
configInfo.bell.toUpperCase is not a function
```

can you add an extra safe guard for that, please?

ShaggyDev:
@Stanzilla Sure thing. Shall I open a new PR with the fix then?

Stanzilla:
@ShaggyDev yes please!

Issue #2861, opened 2018-04-17
Not launching at all on CentOS 7 ("Javascript error occurred in the main process")


darkvertex:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: CentOS Linux release 7.4.1708 (Core)
- **Hyper.app version**: 2.0.0 I guess? I installed from the rpm `hyper-2.0.0.x86_64.rpm`
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: N/A -- it never generated one because it crashes immediately.
- **Relevant information from devtools** _(CMD+ALT+I on macOS, CTRL+SHIFT+I elsewhere)_: N/A
- **The issue is reproducible in vanilla Hyper.app**: Yes.

## Issue

I'm on CentOS and after installing the rpm (`hyper-2.0.0.x86_64.rpm`) as soon as I run `hyper` I am immediately greeted by an error that says:

```
A Javascript error occurred in the main process
Uncaught Exception:
TypeError: Cannot read property 'resize' of undefined
    at Server.Window.rpc.on (/opt/Hyper/resources/app.asar/ui/window.js:146:14)
    at emitOne (events.js:115:13)
    at Server.emit (events.js:210:7)
    at Server.ipcListener (/opt/Hyper/resources/app.asar/rpc.js:33:11)
    at emitTwo (events.js:125:13)
    at EventEmitter.emit (events.js:213:7)
    at WebContents.<anonymous> (/opt/Hyper/resources/electron.asar/browser/api/web-contents.js:266:13)
    at emitTwo (events.js:125:13)
    at WebContents.emit (events.js:213:7)
```
and when I press OK, the Hyper window disappears and quits.

If I run `hyper --help` from a normal terminal, it outputs the help, so it seems something broken UI-wise.

chabou:
So issue is located here: https://github.com/zeit/hyper/blob/57c735cb6da09a109f43d39dae98beede44fa5b5/app/ui/window.js#L146

I really don't know why the session is not found. We can make a test like: https://github.com/zeit/hyper/blob/57c735cb6da09a109f43d39dae98beede44fa5b5/app/ui/window.js#L127-L133
But I'm curious to understand why it happens only in CentOS...

wasabigeek:
[...]

PR #2940, opened 2018-04-27
Patch for RPC handler race condition causing a crash

umpc:
Hello,

First, thank you all so much for your work on Hyper! I had stopped using version 1 after the UI would freeze from attempting to display a large amount of output. I was so glad to see the switch to using a canvas element.

This PR addresses another occurrence of issue #401, where the `session` reference is `undefined`:

screenshot 2018-04-26 23 10 03

A user mistake that causes a large amount of output, leading to a `session` `exit` event, can cause Hyper to crash.

I suspect that the RPC handler for `session` `exit` events and its map `delete` call can, in rare cases, be scheduled to run before the final `data` event handler associated with its `uid` runs, which tries to call `write` on the now `undefined` `session` reference.

I could see where PR #404 added a conditional to catch this issue but was closed, though I did find its change in another RPC event handler within the same file added at a different time, attempting to prevent a crash in an `exit` event handler involving the same `undefined` `session` reference.

I was using version 2.0.0-stable on macOS 10.13 when this occurred. I was purposely fuzzing Hyper with `/dev/zero` and `/dev/urandom` output to test its reliability in handling large amounts of output. I triggered this using well-known output within the first hour of having Hyper installed.

I could not reproduce this issue on-demand, though the cause of the problem seems apparent considering the error in the screenshot, the previous issue/PR, and the existing conditional check of a `session` reference that calls `console.log`.

This PR removes the single `console.log` call and does not add them to the two new conditionals. I mention this as perhaps you would like to keep them there to make a different fix later, though I left them out as it's been nearly two years and I would guess that the proper fix isn't worth the extra code or design changes when a total of three conditionals, with one already present in the codebase, should resolve the crash bugs simply and with no side-effects. If this is acceptable, then this PR is ready to be merged.

I know that this issue was closed before, with a more satisfying fix desired, though the problem can occur in some rare but possible situations, and crash bugs affecting all tabs and windows were my only issue with using Hyper previously on a daily-basis. That is why I have been so thorough in explaining this now.

Thanks again for writing Hyper!

### DIFF
--- a/app/ui/window.js
+++ b/app/ui/window.js
@@ -127,9 +127,6 @@ module.exports = class Window {
       const session = sessions.get(uid);
       if (session) {
         session.exit();
-      } else {
-        //eslint-disable-next-line no-console
-        console.log('session not found by', uid);
       }
     });
     rpc.on('unmaximize', () => {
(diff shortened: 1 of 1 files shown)


chabou:
I am satisfied with this solution.
Fixes #2861 too

PR #2955, opened 2018-04-30
Adding ability to send error object to notify()

sonnyg:
In short, this pull request is to assist with error reporting, and diagnostics.

After the 2.0 upgrade, a number of plugins I use stopped working. Unfortunately, while I would see the notification message to check 'Developer tools', there was no further output or logs available.

After looking through the source, I found that some catch blocks would output the error to console.error, but others would not. Rather than modifying all the catch blocks, I opted to add an additional parameter 'error', to the notify method. If present, the error is logged to the console. I then just modified the catch block to pass the error reference.

Please do not be bashful with your feedback, and thank you so much for this incredible tool.

### DIFF
--- a/app/notify.js
+++ b/app/notify.js
@@ -27,9 +27,13 @@ app.on('ready', () => {
   });
 });
 
-function notify(title, body) {
+function notify(title, body, error) {
   //eslint-disable-next-line no-console
   console.log(`[Notification] ${title}: ${body}`);
+  if (error) {
+    //eslint-disable-next-line no-console
+    console.error(error);
+  }
   if (win) {
     win.webContents.send('notification', {title, body});
   } else {
--- a/app/notify.js
+++ b/app/notify.js
@@ -27,12 +27,12 @@ app.on('ready', () => {
   });
 });
 
-function notify(title, body, error) {
+function notify(title, body, details = {}) {
   //eslint-disable-next-line no-console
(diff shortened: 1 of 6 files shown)


sonnyg:
BTW, I wanted to add a test for the modifications, but I was not quite clear on asserting calls to the console. If someone can provide me some guidance, I would be more than happy to add theses tests.

app/notify.js
```diff
@@ -27,9 +27,13 @@ app.on('ready', () => {
   });
 });
 
-function notify(title, body) {
+function notify(title, body, error) {
```
albinekb: I think it would be a good idea to do {error} as an object on the end, so if we need to send more info to notify in the future we don't need to have another positional argument 👍

sonnyg: It's funny you mention that, because that was actually my first thought!

sonnyg:
@albinekb I modified the parameter as you suggested. Are these changes what you were thinking?

sonnyg:
Excellent, thank you so much!

sonnyg:
I created a corresponding PR in hyper-site, which updates the API description for 'notify'.

https://github.com/zeit/hyper-site/pull/47

chabou:
Thank you for your valuable help 🙏 
Great work and a good thing that you made another PR for our documentation 💯

Issue #2978, opened 2018-05-05
HTM Integration


MisterTea:
Many people would like to use tmux -CC with Hyper.js (see https://github.com/zeit/hyper/issues/2406 ).

Unfortunately, tmux -CC has no public spec and is hard to port to another terminal emulator.  I decided to write a replacement for tmux -CC called "Headless Terminal Multiplexer" (HTM). Here are the reasons why I'm writing something new:

- tmux -CC is tied to iTerm, which is itself tied to OS/X. I am adding support for HTM in hyper.js, which is a terminal emulator that runs in most desktop OSes. One day we will be able to use laptops with a hardware escape key again.
- HTM is an open spec that can be adopted by many terminal emulators. Wez Furlong, author of Wezterm, has helped write the API.
- Some minor design changes, such as having HTM own a single window.

I've been making steady progress and should have a Hyper.js PR soon. Here's a link to a video showing HTM working with Hyper.js: https://youtu.be/Ea6Y08tP-RI

Issue #2406, opened 2017-10-29
Tmux Integration


williscool:
- [ ] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [ ] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**:  Windows 10 Home Build 15063.674
- **Hyper.app version**:  1.4.8
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**:  Vanilla no changes
- **Relevant information from devtools** _(CMD+ALT+I on Mac OS, CTRL+SHIFT+I elsewhere)_:  N/A
- **The issue is reproducible in vanilla Hyper.app**:  Is Vanilla

## Issue

Hi 

Was looking to see about a tmux integration for hyper like in recent versions of `iterm2`

video of it in action:

https://www.youtube.com/watch?v=nGhfE9n1RLo

https://gitlab.com/gnachman/iterm2/wikis/TmuxIntegration

Then you also get all of tmux's plugins and stuff for free too.

williscool:
Super down to help both research and implementation.

Just seeing what other interest or information I can get by opening an issue

MisterTea:
Hey, check out #1912 for some more details on how tmux -cc works.

MisterTea:
Hey all.  I decided to write a replacement for tmux -CC called "Headless Terminal Multiplexer" (HTM).  Here are the reasons why I'm writing something new:

- tmux -CC is tied to iTerm, which is itself tied to OS/X.  I am adding support for HTM in hyper.js, which is a terminal emulator that runs in most desktop OSes.  One day we will be able to use laptops with a hardware escape key again.
- HTM is an open spec that can be adopted by many terminal emulators.  Wez Furlong, author of Wezterm, has helped write the API.
- Some minor design changes, such as having HTM own a single window.

I've been making steady progress and should have a Hyper.js PR soon.  Here's a link to a video showing HTM working with Hyper.js: https://youtu.be/Ea6Y08tP-RI

MisterTea:
I decided to make a new issue since HTM is somewhat off topic from the OP.  https://github.com/zeit/hyper/issues/2978

PR #2988, opened 2018-05-08
Add support for Headless Terminal Multiplexer (htm)

MisterTea:
As discussed in https://github.com/zeit/hyper/issues/2406 and  https://github.com/zeit/hyper/issues/2978 , a terminal multiplexer that integrates with the terminal emulator would be a great addition to Hyper.js.  I initially looked into integrating the existing solution, tmux control center, but tmux -CC has no defined spec and is difficult to reverse engineer.  As a result, I wrote my own headless terminal multiplexer called HTM.  HTM will be packaged with the next version of Eternal Terminal ( https://mistertea.github.io/EternalTerminal/ ).

The core idea is that htm runs inside of ssh/et and sends commands to Hyper.js (or any other terminal emulator).  Hyper.js gets a special ANSI escape sequence which puts hyper in HTM mode, and then the current state of the tabs/panes.  As users make tabs, delete tabs, or type in panes, all of this is funneled through htm.  If someone exits hyper.js and comes back, running htm restores the state when the person exited.  To see a demo of how this works, check out https://www.youtube.com/watch?v=Ea6Y08tP-RI&feature=youtu.be

This PR is a WIP because I assume it's a big enough change that it will require more iterations before it's ready to merge.  It's 100% functional and can be tried by following these instructions:

- First, check out the htm branch of eternal terminal and build the source (see https://github.com/MisterTea/EternalTerminal#building-from-source ).
- Then, ensure that htm and htmd are in your path by copying to /usr/local/bin or modifying your .bashrc/.zshrc
- Then, build & run yarn with this PR.
- Inside hyper (can be in an ssh session or not), run htm
- Do some work
- Close hyper.js
- re-open and re-run htm
- your state is preserved.

### DIFF
--- a/app/commands.js
+++ b/app/commands.js
@@ -9,16 +9,16 @@ const commands = {
   },
   'tab:new': focusedWindow => {
     if (focusedWindow) {
-      focusedWindow.rpc.emit('termgroup add req');
+      focusedWindow.rpc.emit('termgroup add req', {});
     } else {
       setTimeout(app.createWindow, 0);
     }
   },
   'pane:splitVertical': focusedWindow => {
-    focusedWindow && focusedWindow.rpc.emit('split request vertical');
+    focusedWindow && focusedWindow.rpc.emit('split request vertical', {});
   },
   'pane:splitHorizontal': focusedWindow => {
-    focusedWindow && focusedWindow.rpc.emit('split request horizontal');
+    focusedWindow && focusedWindow.rpc.emit('split request horizontal', {});
   },
   'pane:close': focusedWindow => {
     focusedWindow && focusedWindow.rpc.emit('termgroup close req');
--- a/app/ui/window.js
+++ b/app/ui/window.js
@@ -147,6 +147,17 @@ module.exports = class Window {
       session.resize({cols, rows});
     });
     rpc.on('data', ({uid, data, escaped}) => {
+      window.handleSessionInput(uid, data, escaped);
+    });
+    window.initSession = (opts, fn_) => {
+      fn_(uuid.v4(), new Session(opts));
+    };
+
+    window.handleSessionData = (uid, data, handleSessionCallback) => {
+      // By default, just execute the callback.  Plugins can override.
+      return handleSessionCallback(uid, data);
+    };
+    window.handleSessionInput = (uid, data, escaped) => {
       const session = sessions.get(uid);
 
       if (escaped) {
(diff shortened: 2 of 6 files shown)


ppot:
Can you create a new one with only your changes by pulling the latest version? 90k line of addition if very difficult to track. @chabou What's your take on this?

MisterTea:
@ppot Sorry about that: it was using the canary as the base instead of master.

albinekb:
Really cool! However I think this is better suited as a plugin, since it's not a thing that will be used by even 1% of hyper users, I think most of the stuff you're doing is already possible from a plugin, and if we need to expose more API for you to be able to pull it off I think that's a better way to do it than to integrate this deeply into hyper itself. 

Thoughts?

MisterTea:
@albinekb Sure, I'm fine with making it a plugin, but we will need to make some core changes either way.  I'll see how much I can move over.

albinekb:
@chabou what do you think?

MisterTea:
@albinekb So I started looking into it.  Most of the logic can go into a plugin.  The hardest part is the code in window.js.  I have a new type of Session class that doesn't open a pty but instead forwards all the keystrokes to the session that's running htm.  I may change that from Session/HtmSession to PtySession/DummySession and make the DummySession go through RPC to send/receive data.

albinekb:
Awesome to hear, yes, that sounds like a plan that could support more other interesting plugins too! 👍

MisterTea:
@albinekb Alright, I've moved as much as I can out to a plugin: https://github.com/MisterTea/hyper-htm

Overall, the core changes remaining can be described as follows:

- Actions to create a term group / session now optionally take a termGroupUid and sessionUid so the HTM uid and Hyper uid remain in sync
- Split actions now take a sessionUid and also a sourceUid instead of splitting from the active session.  These are also optional and revert to the current behavior when empty
- A new class, FollowerSession, that does not contain a PTY but instead forwards all input & resize requests to the window
- Functions in window.js were modified to make it easier to override them in a plugin

I'm leaving the WIP tag on because I still need to test more edge cases and also replace iTerm with this and use it for a few days before I'm ready to merge, but in the meantime it would be great to get feedback.

MisterTea:
I changed the logic so initSession is now a function inside the window class, and this way the plugin can override it.  With this change, the FollowerSession logic can move entirely into the plugin and the overall change to hyper.js becomes pretty small.

MisterTea:
@albinekb I just realized that active development is happening on the canary branch, not the master branch.  I have a branch in my fork called "htmcanary" that is the same changes rebased on the canary branch.  Should I open a new PR with that branch?

albinekb:
Yes please 🙏 Sorry for not noticing 🙈  @MisterTea

MisterTea:
Moved to https://github.com/zeit/hyper/pull/3026

Issue #2886, opened 2018-04-20
Change the tab title in hyper 2.0.0


bvaldes:
Hi,

I am on OSX but I plan to use Hyper on Windows too because of the new update with xterm.js

I've done a fresh install of hyper 2.0.0 and I would like to change the tab title.

The only answer I found was to Hyper 1.X with the hyperterm-title plugin. The issue is that it doesn't work at all on 2.0.0.

I saw that this plugin was merged with something that you call core but I've no idea what people talk about.

I hope that my question is not stupid. Thanks for the help.

Stanzilla:
What would you like the tab title to say? Something dynamic like the current folder? Which shell are you using inside of Hyper?

bdinesh:
How can we rename the tab title like we want in Hyper 2.0?

cr101:
I'm on Windows 10 and the title for all the tabs is "Shell". Is there a way to change the tab title?

rjurney:
I need this too.

cdetrio:
I'd like this feature (on Mac all tabs also say "Shell"), and be able to set window names too.

ankur2002:
I upgraded to 3.0.2 and now the tab titles don't show up anymore. This seemed to be working on my mac when I was still using version 2.x but it stopped working on 3.x.

Can we restore the tab titles again?

dincercelik:
Same here with 3.0.2 on OSX 10.14.5

drbparsons:
I managed to find a solution that changes the tab titles on Mac to the current directory:

Assuming your default shell is zsh, add this to your ~/.zshrc file:

```
case $TERM in
  xterm*)
    precmd () {print -Pn "\e]0;%~\a"}
    ;;
esac
```

![image](https://user-images.githubusercontent.com/1021530/59541176-0b0dd400-8ef8-11e9-923d-c7a843e720a0.png)

(via http://blog.pengyifan.com/how-to-set-terminal-title-dynamically-to-the-current-working-directory/)

Issue #1188, opened 2016-12-13
Tab titles are not behaving correctly on Windows


stereokai:
- **OS version and name**: Windows 10 64bit
- **Hyper.app version**: 1.0.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: [hyper.js gist](https://gist.github.com/stereokai/2b0c8e4d9b997ee8c1d31f73ac43c63f)
- **Relevent information from devtools**: N/A
- **The issue is reproducible in vanilla Hyper.app**: Reproducible in vanilla

## Issue
Current directory is not displayed. All tab titles display "Shell" regardless of which shell I am using (tried cmd and Bash on Ubuntu on Windows).

audunolsen:
This is true of macOS too.

stereokai:
@audunolsen in version 1.0.0?

I assumed it was Windows only because of this notice: https://github.com/parro-it/hyperterm-title/issues/5

audunolsen:
Yes.

<img width="499" alt="screenshot 2016-12-14 01 12 42" src="https://cloud.githubusercontent.com/assets/16224137/21164615/8cd5f4fa-c19a-11e6-963c-e1b3fcb786ca.png">

Tab titles doesn't reflect directory or process running.

stereokai:
I stole your comment for a better issue title :)

parro-it:
Hi, since my plugin `hyperterm-title` is now embedded in core, tabs title is now managed directly by your shell. There is a special escape sequence command that is meant to set terminal title.

You have to setup a shell prompt that uses this escape sequence in order to avoid "Shell" as tab title.

parro-it:
Checkout [pure prompt](https://github.com/sindresorhus/pure) for zsh. I'm sure there are similar ones for bash or fish too

parro-it:
I agree that a better default behaviour is useful... Show the Shell name should be simple to implement 
The cwd or current foreground process instead was not trivial, previous hyper versions uses a polling procedure that is too resource heavy...

stereokai:
@parro-it thanks for chiming in! I appreciate your work on this plugin as it is a feature probably most users if not everyone take it for granted.

With that said, I think an ideal solution would be supporting all terminals, in particular now that Hyper is meant to be used on all major operating systems. So perhaps additional methods of retrieving the shell title are needed :)

stereokai:
@parro-it you were quicker than me!

parro-it:
[...]

PR #3015, opened 2018-05-14
Tab titles now show cwd (#2907)

### LINKED ISSUE #2907: Revisit the idea of rename tabs

- **OS version and name**: OS X El Capitan 10.11.6
- **Hyper.app version**: 2.0.0

## Issue

I know this has been discussed before in #284, but as a SysAdm I frequently have to ssh to other machines and the current approach will fail, because you can no longer rename the tab through bash tricks.

I feel a better way could be the addition of another hyper command(rename-tab) in hyper-keymap and prompt the user in an os-level dialog and ask the new title.

El-Dringo-Brannde:
Closes #2907 #2886 #1188 

This PR is ready to be merged, ran dist build and it works all the same on my Mac. 

**However,** I don't have a PC, or a VM of a PC to test the windows pathing on. Theoretically... It should work as I have kindly borrowed the same code that gets CWD from [hyper-statusline](https://github.com/henrikdahl/hyper-statusline/blob/master/index.js)


lib/utils/fs.js
```diff
@@ -0,0 +1,25 @@
...
+export default function getCWD(pid, action) {
+  let cwd;
+  if (process.platform == 'win32') {
+    let directoryRegex = /([a-zA-Z]+)/im;
+    if (action && action.data) {
+      let path = directoryRegex.exec(action.data);
+      if (path) cwd = path[0];
+    }
+  } else {
+    cwd = execSync(`lsof -p ${pid} | awk '$4=="cwd"' | tr -s ' ' | cut -d ' ' -f9-`, {encoding: 'utf8'});
```
albinekb: Is this the best way to get the current CWD? 🤔 I wonder if there is an async way of doing it so we don't lock up the app...

El-Dringo-Brannde: @albinekb There are async methods of running child_processes, but I thought in Redux making stuff async is frowned upon (still learning, I've heard of thunks being used for async actions). That's why I only have this function run when the user hits the enter key.

chabou: It will be executed each time that '\n' is output in term. Do not execute `yes` in your fork 😁

chabou:
This is not the right approach.
Performances will be heavily impacted.

BTW I'm using `oh-my-zsh` and my tab title shows my cwd.
But by default, I would have preferred to have current running command. I think that it was the case in hyper 1.X

But feel free to make a plugin with your work 🙏

El-Dringo-Brannde:
If I had it read the `enter` key press to display the current running command would the PR have a chance of being accepted?

El-Dringo-Brannde:
@chabou @albinekb Can I get a re-review, I had it watch on enter key presses only.  Or at the very least a hint on what would be acceptable features for a PR? Trying to get something merged for my final project assignment in my OSS class.

Stanzilla:
@El-Dringo-Brannde willing to pick this up again?

PR #3017, opened 2018-05-14
Adjust clickable region of control buttons

lujjjh:
On Windows, the close button (rather than the border) is expected to be hit when someone clicks the top-right corner of the screen if a window is maximized.

### DIFF
--- a/lib/components/header.js
+++ b/lib/components/header.js
@@ -228,6 +228,7 @@ export default class Header extends React.PureComponent {
             height: 34px;
             justify-content: space-between;
             position: fixed;
+            top: 0;
             right: 0;
           }
 


Stanzilla:
I don't quite understand, when I click on the top right corner I hit close all the time?

lujjjh:
@Stanzilla If the window is maximized, yes.

Here is an example in [User Interface Principles (intuitiveness)](https://msdn.microsoft.com/en-us/library/windows/desktop/ff728831%28v=vs.85%29.aspx#intuitiveness):

> In one of the early Windows Vista preview builds, the Minimize, Maximize, and Close buttons of any window became different. In previous versions of Windows (especially when using a single monitor), you develop a habit of moving the cursor in the top-right corner of the screen and clicking. This always resulted in the window closing. Now in this particular build of Windows Vista, there were approximately 8 pixels worth of space between the Close button and the right-most border of the window. The extra space did make it look cool (and was probably necessary for the cool glow animation the button sported), but was irritating because it did not allow users to close open windows as easily. Reconditioning your mind can be tough. Thankfully, in the following build, this issue was addressed. Now, there still is space between the border of the window and the close button, but clicking that space also causes the window to be closed.

Stanzilla:
Well you said `is expected to be hit when someone clicks the top-right corner of the screen if a window is maximized.` and that is exactly what is happening here?

lujjjh:
There is actually a 1-pixel border on the top so I hit the border if I click **exactly** the top-right corner. I have to move the cursor down and click the close button.

That is why I move `.header_windowControls` 1 pixel higher, which makes it easier to close the window.

Stanzilla:
Alright, fine by me :)

chabou:
@lujjjh Thank you for your help

Issue #3000, opened 2018-05-10
Add startup/working directory config


chabou:
It should be more easy to configure a startup directory.
Currently, it has to be done at shell level for macOS/Linux shells (.bashrc/.zshrc...) and at `hyper` config level for CMD/PowerShell (See https://github.com/zeit/hyper/issues/2794).

I suggest to add an optional `workingDirectory` config parameter. If it is empty/undefined/malformated, default (config location) directory will be used.
Only an absolute path is valid. If a path is passed as parameter to `hyper`, it must be used instead of config/default, as is currently the case.

Magic should occur here: https://github.com/zeit/hyper/blob/f72093c9d75da963544a48a0eee8107ecb170cf8/app/ui/window.js#L93

mtslzr:
Created PR #3021 for this.

DamianRivas:
@chabou perhaps this issue should be closed since it was resolved by commit 164a600805cc8ba24ed01d58ba5cae6455107b2b

I was looking for a good issue to work on. I need to make 10 PRs to OSS projects before I can graduate from my apprenticeship. I use Hyper daily so I would love to help with anything if y'all need it :)

corbanb:
Does this no longer work in 3.0? 

`workingDirectory: "~/Sites/someplace",`

corbanb:
Does this no longer work in 3.0? 

`workingDirectory: "~/Sites/someplace",`

For now I just added `cd ~/Sites/someplace` to one of my custom plugin files for`oh-my-zsh`. ha

kaushikchaubal:
I just added the workingDirectory field in my .hyper.js file as well and it didn't work as expected.

I aim to create a PR with a fix soon.

kaushikchaubal:
PR created for this: https://github.com/zeit/hyper/pull/3769

Ready for review

chabou:
There is a plugin for this: https://www.npmjs.com/package/hyper-working-directory

But maybe it should be a core feature?

kaushikchaubal:
My two cents... this was a core feature which then broke. I think if we want to keep like-fo-like, it should be a core feature. (happy to be overruled)

chabou:
I think it was never implemented 🤔

ppot:
Was never a core feature. But had a flag in the config to define a default for the user.

johnletey:
Has this been implemented? I see that #3769 was merged, but see https://github.com/zeit/hyper/pull/3769#issuecomment-552374299

[...]

PR #3021, opened 2018-05-15
Added check for startup directory and blank option into default config.

mtslzr:
Opening this to tackle #3000. Added a simple check (expanding the check for an argument being passed to also check for a config option named workingDirectory) and then added that option to the default config. Current priority is command-line argument > config file option > home directory.

### DIFF
--- a/app/ui/window.js
+++ b/app/ui/window.js
@@ -41,6 +41,14 @@ module.exports = class Window {
       window.setBackgroundColor(toElectronBackgroundColor(cfg_.backgroundColor || '#000'));
     };
 
+    // set startup directory
+    let workingDirectory = cfgDir;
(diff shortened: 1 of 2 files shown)


app/ui/window.js
```diff
@@ -41,6 +41,14 @@ module.exports = class Window {
       window.setBackgroundColor(toElectronBackgroundColor(cfg_.backgroundColor || '#000'));
     };
 
+    // set startup directory
+    let workingDirectory = cfgDir;
```
chabou: You have to move this block below, in handler passed to `app.config.subcribe()` and after config decoration by plugins.
This way, your working directory config will be hot reloaded and modifiable by plugins.

Bonus: please find a way to get rid of this use of `process.argv` in this module. (Already here before your PR, I know 😁 ) 🙏

Issue #2998, opened 2018-05-10
Revisit CLI installation strategy 


jaydh:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Linux
- **Hyper.app version**: Master
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: Default
- **Relevant information from devtools** _(CMD+ALT+I on macOS, CTRL+SHIFT+I elsewhere)_: 
- **The issue is reproducible in vanilla Hyper.app**: Yes

## Issue
[This issue](https://github.com/zeit/hyper/issues/2891) was brought up a couple weeks ago that more or less got resolved by [2885](https://github.com/zeit/hyper/pull/2885) by silently failing cli installation for linux distros.

I was considering different approaches to symlinking/installing the cli on linux(using policykit to prompt user for privileged access to /usr/local/bin, or adding an alias to .bash_profile) but I got curious about the overall CLI install strategy. Overall I'm not convinced it's secure or tenable cross-platform. I couldn't believe that macOS even allowed an unprivileged user to symlink a file to /usr/local/bin from an executable. I wanted to ask why the cli wasn't enclosed in the hyper app itself by reading command line arguments in the main thread to trigger cli commands. This seems to circumvent issues a lot of these moving target issues with symlinking and checking for the current cliscriptpath.

Namely this issue is prevalent in AppImage builds because the the cli is enclosed in the AppImage which will never be mounted without it's execution to begin with, so a symlink to /tmp/.mount###-hyper/... is dead on exit.

Currently hyper doesn't accept command line arguments anyways so there's no worry of CLI namespace conflicts.

PR #3024, opened 2018-05-16
Bootstrap CLI to hyper startup

jaydh:
This bootstraps the CLI into the hyper startup process. Namely, this bypasses cross-platform installation for a separate binary that has caused issues for linux users. I've outlined some of the details [here](https://github.com/zeit/hyper/issues/2998).

The main CLI function returns a flag for the main thread to execute the terminal emulator if and only if hyper is run without any command line arguments (both in dev and prod). If there are arguments the CLI parses them, executing the commands, or showing a help screen and terminating the node process altogether.


app/cli/index.js
```diff
@@ -82,15 +88,18 @@ const lsRemote = pattern => {
...
+    .finally(() => process.exit(0));
 };
 
 args.command(['s', 'search'], 'Search for plugins on npm', (name, args_) => {
+  commandInvoked = true;
   const spinner = ora('Searching').start();
   const query = args_[0] ? args_[0].toLowerCase() : '';
-
-  commandPromise = lsRemote(query)
+  console.log(1);
```
nazariyv: probably don't need this here?

jaydh: Yup good catch! There's also a separate log statement on 102 we can delete

app/cli/index.js
```diff
@@ -1,21 +1,22 @@
...
 const api = require('./api');
-
+const isDev = require('electron-is-dev');
+const promiseFinally = require('promise.prototype.finally');
+promiseFinally.shim();
 const PLUGIN_PREFIX = 'hyper-';
 
-let commandPromise;
-
+const configLocation = process.platform === 'win32' ? process.env.userprofile + '\\.hyper.js' : '~/.hyper.js';
```
ppot: Will crash with new location of `.hyper.js`

app/cli/index.js
ppot: remove

app/cli/index.js
```diff
@@ -82,15 +88,18 @@ const lsRemote = pattern => {
...
 args.command(['s', 'search'], 'Search for plugins on npm', (name, args_) => {
+  commandInvoked = true;
   const spinner = ora('Searching').start();
   const query = args_[0] ? args_[0].toLowerCase() : '';
-
-  commandPromise = lsRemote(query)
+  console.log(1);
+  lsRemote(query)
     .then(entries => {
+      console.log(2);
```
ppot: remove

Stanzilla:
Hey @jaydh are you willing to pick this up again?

app/cli/index.js
Stanzilla: @jaydh

PR #3032, opened 2018-05-19
update xterm to 3.4.1

Stanzilla:


package.json
```diff
@@ -200,7 +200,7 @@
     "styled-jsx": "2.2.6",
     "stylis": "3.5.0",
     "uuid": "3.1.0",
-    "xterm": "chabou/xterm.js#7cb4b8e"
+    "xterm": "3.4.0"
```
albinekb: 🎉

Stanzilla:
Correct.

chabou:
CI failed because `eslint-plugin-react` version was not pinned and was upgraded to 7.8.2.
We don't use react 16.3.0, only 16.2.0.

Replacing `componentWillReceiveProps()` and upgrading react should take place in another PR.

I am currently fixing this PR (reverting last commit and pining dev dependencies)

PR #3038, opened 2018-05-20
Add scrollback option to config

jaridmargolin:
As I was preparing this PR I was surprised to find that there was very little test coverage. Perhaps I am missing something?

It'd be great to get this pushed through and I am fairly confident the change is working as expected, but it feels strange to add functionality without writing/running tests 🤔

... Additionally, if there is interest in merging this PR, documentation should be updated accordingly.

### DIFF
--- a/lib/components/term-group.js
+++ b/lib/components/term-group.js
@@ -58,6 +58,7 @@ class TermGroup_ extends React.PureComponent {
     const props = getTermProps(uid, this.props, {
       isTermActive: uid === this.props.activeSession,
       term: termRef ? termRef.term : null,
+      scrollback: this.props.scrollback,
       backgroundColor: this.props.backgroundColor,
       foregroundColor: this.props.foregroundColor,
       colors: this.props.colors,
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -26,6 +26,7 @@ const getTermOptions = props => {
   const backgroundColor = needTransparency ? 'transparent' : props.backgroundColor;
   return {
     macOptionIsMeta: props.modifierKeys.altIsMeta,
+    scrollback: props.scrollback,
     cursorStyle: CURSOR_STYLES[props.cursorShape],
     cursorBlink: props.cursorBlink,
     fontFamily: props.fontFamily,
--- a/lib/components/terms.js
+++ b/lib/components/terms.js
@@ -88,6 +88,7 @@ export default class Terms extends React.Component {
             terms: this.terms,
             activeSession: this.props.activeSession,
             sessions: this.props.sessions,
+            scrollback: this.props.scrollback,
             backgroundColor: this.props.backgroundColor,
             foregroundColor: this.props.foregroundColor,
             borderColor: this.props.borderColor,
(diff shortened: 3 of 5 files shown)


chabou:
@jaridmargolin Thank you for this PR 🙏 

You're totally right. We need more unit tests and more functional tests with `spectron`

jcrben:
Never too late to start. At least xterm is decently tested (see e.g. https://github.com/xtermjs/xterm.js/tree/master/src)... looking into tests. At work I'm spoiled by TDD using jest and jsdom with mocking of the API data (which is tested separately) - would like to be able to have a similar TDD flow.

AhmedBM:
I just added the scollback option in v2.1.1 and im not seeing it. It looks like this change was merged sometime ago so it should be there. Am i missing something here?

sankalp-khare:
@AhmedBM I can confirm it works in 2.1.2, though I don't know how to set it to infinite. For now I went with an absurdly large value
```
        scrollback: 2147483647,
```

aecorredor:
> @AhmedBM I can confirm it works in 2.1.2, though I don't know how to set it to infinite. For now I went with an absurdly large value
> 
> ```
>         scrollback: 2147483647,
> ```

I tried this and don't really get that much more when scrolling back up. It's for sure not close to being infinite. Any ideas?

rjurney:
I still don't have additional scrollback after setting:

```json
{
    scrollback: 100000000
}
```

What do I do?

ddimico:
Agree with @rjurney, scrollback appears to not actually do anything to the buffer length. Seems a bit crazy this isn't something of higher priority. I have no idea how people are using this with only an 1000 buffer length.

rjurney:
@ddimico This makes the thing unusable. I abandoned it. If it is fixed I'll look again.

jcruzdzk:
where i set this value? in .hyper.js?

jerch:
hyper prolly solely relies on the scrollback handling done by xterm.js and has no additional infinite scroll shim. If so, you should not try to set it to some arbitrary high value or the app will crash with OOM.

Some calculations - JS engines typically dont allow more than ~2GB to be claimed. A cell takes 12 bytes, thus:
- at 100 cells per line: max of ~1.5M lines
- at 1000 cells per line: max of ~150k lines

and so on...

ajilling:
Echoing this does not work - using config.json

aecorredor:
Just wanted to touch base on this. Did somebody ever find a way to get this working? It still doesn't work for me even on the latest hyper version.

dubbatk:
> Just wanted to touch base on this. Did somebody ever find a way to get this working? It still doesn't work for me even on the latest hyper version.

I would like to confirm that this works. You will need to add the setting under config: {...}, it is the same as adding lineHeight or letterSpacing. Follow the same format as any general setting under the config block :-)

PButcher:
The `scrollback` option should be added to the `~/.hyper.js` by default. The default limit is irritatingly small.

PButcher:
The `scrollback` option should be added to `~/.hyper.js` by default. The default limit is irritatingly small.

The option does work, you just have to add it yourself.

PR #3046, opened 2018-05-22
Add `isWsl` flag to session/action objects

rfgamaral:
The `isWsl` flag denotes if the running shell is a Windows Subsystem for Linux (WSL) shell by pattern matching the user configurable `shell` property.

This flag has support for invoking WSL with multiple executables:

- `bash.exe`
- `debian.exe`
- `kali.exe`
- `opensuse.exe`
- `ubuntu.exe`
- `wsl.exe`

I'm not sure if this is something you guys want to add to Hyper core but it's just a small flag to help plugin developers easily know if the running shell is WSL or not, just like they currently need to know if the host OS is Windows and Linux (provided by `process.platform`) to properly do their plugin magic. This will come in handy to better support WSL within plugins.

Let me know if you have any questions or a better idea to implement this.

### DIFF
--- a/app/ui/window.js
+++ b/app/ui/window.js
@@ -93,6 +93,7 @@ module.exports = class Window {
           cwd: process.argv[1] && isAbsolute(process.argv[1]) ? process.argv[1] : cfgDir,
           splitDirection: undefined,
           shell: cfg.shell,
+          isWsl: /(?:bash|debian|kali|opensuse|ubuntu|wsl)\.exe$/.test(cfg.shell),
           shellArgs: cfg.shellArgs && Array.from(cfg.shellArgs)
         },
         options
@@ -110,6 +111,7 @@ module.exports = class Window {
           uid,
           splitDirection: sessionOpts.splitDirection,
           shell: session.shell,
+          isWsl: sessionOpts.isWsl,
           pid: session.pty.pid
         });
 
--- a/app/ui/window.js
+++ b/app/ui/window.js
@@ -93,7 +93,7 @@ module.exports = class Window {
           cwd: process.argv[1] && isAbsolute(process.argv[1]) ? process.argv[1] : cfgDir,
           splitDirection: undefined,
           shell: cfg.shell,
-          isWsl: /(?:bash|debian|kali|opensuse|ubuntu|wsl)\.exe$/.test(cfg.shell),
+          isWsl: /(?:bash|debian|kali|opensuse|ubuntu(?:1804)?|wsl)\.exe$/.test(cfg.shell),
           shellArgs: cfg.shellArgs && Array.from(cfg.shellArgs)
         },
         options
--- a/app/ui/window.js
+++ b/app/ui/window.js
@@ -13,6 +13,7 @@ const fetchNotifications = require('../notifications');
 const Session = require('../session');
 const contextMenuTemplate = require('./contextmenu');
 const {execCommand} = require('../commands');
+const {isWslShell} = require('../utils/wsl');
 
 module.exports = class Window {
   constructor(options_, cfg, fn) {
@@ -93,7 +94,7 @@ module.exports = class Window {
           cwd: process.argv[1] && isAbsolute(process.argv[1]) ? process.argv[1] : cfgDir,
           splitDirection: undefined,
           shell: cfg.shell,
-          isWsl: /(bash|debian|kali|opensuse|ubuntu(1804)?|wsl)\.exe$/.test(cfg.shell),
+          isWsl: isWslShell(cfg.shell),
           shellArgs: cfg.shellArgs && Array.from(cfg.shellArgs)
         },
         options
--- a/app/ui/window.js
+++ b/app/ui/window.js
@@ -93,7 +93,7 @@ module.exports = class Window {
           cwd: process.argv[1] && isAbsolute(process.argv[1]) ? process.argv[1] : cfgDir,
           splitDirection: undefined,
           shell: cfg.shell,
-          isWsl: /(?:bash|debian|kali|opensuse|ubuntu(?:1804)?|wsl)\.exe$/.test(cfg.shell),
+          isWsl: /(bash|debian|kali|opensuse|ubuntu(1804)?|wsl)\.exe$/.test(cfg.shell),
           shellArgs: cfg.shellArgs && Array.from(cfg.shellArgs)
         },
         options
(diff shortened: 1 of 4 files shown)


app/ui/window.js
```diff
@@ -93,6 +93,7 @@ module.exports = class Window {
           cwd: process.argv[1] && isAbsolute(process.argv[1]) ? process.argv[1] : cfgDir,
           splitDirection: undefined,
           shell: cfg.shell,
+          isWsl: /(?:bash|debian|kali|opensuse|ubuntu|wsl)\.exe$/.test(cfg.shell),
```
Tyriar: I believe there is also `ubuntu1804.exe`

rfgamaral: I must admit I added these by intuition because I couldn't find a document with all the distros exectuables. I only use Debian myself. I'll add that version to the list soon.

However, as the number of distros may grow and some of them may have multiple executables (case in point), a regex is probably not the best approach. But I honestly can't think of another way to detect if Hyper is running a WSL shell.

Any other ideas?

Tyriar: Yes actually :smile: 

I only know about that exe because I recently did something very similar in vscode. Check out the list I came up with https://github.com/Microsoft/vscode/blob/86adb527f12c12c1c04806cfa0cd435f98475a6f/src/vs/workbench/parts/terminal/node/windowsShellHelper.ts#L13 (which also includes powershell and cmd, everything else is wsl).

rfgamaral: Based on your list, I guess I have a couple of entries to fix 😄 

I might follow your approach instead of regex, it will be easier to maintain in the long run.

rfgamaral: @Tyriar Changed the implementation, could you please re-review? I've also took the liberty of using `path.sep` instead of the hard-coded `/` in one other instance.

Tyriar: I don't know the code base too well so can't give a full review, the list of exes looks good though :+1:

app/utils/wsl.js
```diff
@@ -0,0 +1,22 @@
...
+const WSL_SHELL_EXECUTABLES = [
+  'bash.exe',
+  'wsl.exe',
+  'ubuntu.exe',
+  'ubuntu1804.exe',
+  'kali.exe',
+  'debian.exe',
+  'opensuse-42.exe',
+  'sles-12.exe'
+];
```
chabou: Can you transform this into something like:
```
const WIN_SHELLL_EXECUTABLES = [
  'cmd.exe',
  'powershell.exe'
]
function isWslShell(shell) {
  return process.platform === 'win32' && !WIN_SHELL_EXECUTABLES.indexOf(shell.split(path.sep).pop()) !== -1;
}
```
Be careful with empty/missing shell config.

It would be easier to maintain like @Tyriar suggested.

chabou:
Can you give some examples of plugins that need to know if current shell is running WSL or not?

app/utils/wsl.js
Tyriar: I disagree, there are many other non-WSL shells: git bash, python, gchi, etc.

rfgamaral:
@chabou Plugins like `hyper-tab-icons`, `hypercwd` and `hyper-statusline` will need special handling to work within WSL (they currently don't; the last 2 have open issues regarding that). There could be others out there but these are the ones I'm personally interested in using and thus fixing their WSL issues.

To give a more in-depth context, lets use `hypercwd` as an example.

This plugin opens new tabs with the same directory as the current tab in Hyper. It does this by reading the current tab directory with `lsof` on Linux and relies on the current tab directory that is printed on the Command Prompt / PowerShell on Windows (which is always the absolute path).

The Windows technique does not work for WSL because it's a Linux prompt with color codes in the mix and one could always be using a prompt configuration to show a relative path, for instance. You can't also rely on `lsof` because although WSL is Linux, Hyper is still running on Windows and `lsof` doesn't exist on Windows.

In other words, WSL needs special handling and `hypercwd` needs to know which method to apply to detect the absolute path for the current directory.

app/utils/wsl.js
rfgamaral: Yeah, either we have a full list for valid WSL shells, or non-WSL shells. Since the function is called `isWslShell` and the property is `isWsl`, I think it makes more sense to have a list of valid WSL shells. I'll handle the empty/missing shell configuration though.

chabou: Oh yes, you're right.
But, in this case, it will take some effort to keep this list up to date.

rfgamaral: It might, it all depends on how frequent new Linux distros are going to be supported by WSL and if any of them will include different exes than the default ones. But I honestly can't think of any other way to detect if the active/running shell is WSL or not. If you have a suggestion, I'm all hears :)

rfgamaral:
@chabou What do you think of the current implementation? Could we get this merged? It would help immensely to have this implemented so I could start working on fixing some plugins WSL issues.

lib/reducers/sessions.js
```diff
@@ -41,7 +43,8 @@ const reducer = (state = initialState, action) => {
           cols: action.cols,
           rows: action.rows,
           uid: action.uid,
-          shell: action.shell.split('/').pop(),
+          shell: action.shell.split(path.sep).pop(),
+          isWsl: action.isWsl,
```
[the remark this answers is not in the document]
rfgamaral: To be sure, are you suggesting I change this to `isWsl: isWslShell(action.shell)`?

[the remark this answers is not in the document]
rfgamaral: @chabou There's a little problem with that I'm not sure how to handle given that I don't know the code base so well. The `isWsl` function was created as an `app` util function and this reducer is inside the `lib` folder. I'm not sure I can/should import stuff from `app` into `lib` (and vice-versa for that matter).  How should I proceed?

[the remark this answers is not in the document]
chabou: I would suggest to move `app/utils/wsl.js` to `lib/utils/wsl.js` or move your function directly into `lib/reducers/sessions.js` because your function will not be used anywhere else.

innerlee:
according to https://github.com/Microsoft/WSL/issues/2715#issuecomment-386338451, `/bin/wslpath` and `/init` are the only binaries wsl was injected into. So you can check the existence of these two files?

machinshin:
@rfgamaral : any updates on this thread/PR?

Stanzilla:
Instead of hardcoding exe names, could we do something like this instead? https://github.com/microsoft/terminal/pull/1050

PR #3047, opened 2018-05-22
fix mouse wheel close tab event.

WeslyG:
Hey, i am fix Mouse wheel close tab event)

### DIFF
--- a/lib/components/tab.js
+++ b/lib/components/tab.js
@@ -46,6 +46,7 @@ export default class Tab extends React.PureComponent {
           onMouseEnter={this.handleHover}
           onMouseLeave={this.handleBlur}
           onClick={this.props.onClick}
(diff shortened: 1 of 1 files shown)


lib/components/tab.js
```diff
@@ -54,7 +55,7 @@ export default class Tab extends React.PureComponent {
           {this.props.customChildrenBefore}
           <span
             className={`tab_text ${isLast ? 'tab_textLast' : ''} ${isActive ? 'tab_textActive' : ''}`}
-            onClick={this.handleClick}
```
chabou: Why do you remove this handler (and keep it on `<li>` component)?

If you add onMouseUp, I suppose that is because onClick is not triggered with the middle button?
I would suggest splitting `handleClick` handler (because you have to test `isMiddleClick` only in a `onMouseUp` handler)

Issue #280, opened 2016-07-19
Create options around scrollback buffer


chiefy:
Easy way to crash Hyperterm:

```
find /
```

Since there's unlimited scrollback buffer, memory just keeps getting eaten especially if you're tailing logs or catting large logfiles. It would probably solve the issue if there was an option to limit scrollback (check out iTerm2's options around this)

rgl:
This is a hterm issue that is described at https://crbug.com/220406.

ppot:
I would need to play with this. How far in the scroll back Iterm2's handle ? @chiefy

chiefy:
@ppot iTerm's scrollback buffer is configurable by numbers of lines or "unlimited"

cavaliercoder:
It would also be useful to disable the buffer entirely (as I do in OS X Terminal), so I can use tmux instead for scrollback.

chabou:
@cavaliercoder maybe [hyperterm-alternatescroll](https://github.com/lkzhao/hyperterm-alternatescroll) can help ?

rickarubio:
Is there any way to set the scrollback buffer size / scrollback lines in hyper ? I am running into an issue with hyper 2.0.4 where I can only scroll a few hundred lines at most, but I would really like to be able to scroll back through my terminal output much further than this.

AuHau:
Is there any update on this? The current size of scrollback buffer is not really sufficient...

chabou:
We need to use `xtermjs` `scrollback` option: https://github.com/xtermjs/xterm.js/blob/a3ce881897818d1c61476620e7d037717ca9a68f/src/Terminal.ts#L113

Pretty easy to implement and very useful.

chhschou:
Closed #3056 as #3038 already exists and pending review by @chabou.

chhschou:
This can close, #3038 has been merged.

albinekb:
Thanks for the help @chhschou 🎉

jorisw:
So? What is the option and what kind of values does it take?

albinekb:
@jorisw  It's documented here: https://hyper.is/#cfg
![image](https://user-images.githubusercontent.com/5027156/45495131-eae79380-b772-11e8-8ffd-8cbd7f66fe08.png)

sobolevn:
Is it possible to disable it completely? `Terminal.app` has this mode called "as much as my memory allows".

And I am quite used to it. So, setting any kind of `scrollback` numbers is not the same.

[...]

PR #3056, opened 2018-05-24
add scrollback buffer config option

chhschou:
As stated. The config option added is called `scollback` and takes a number.
Without the option in config the xterm.js default is used (1000?) I believe.

Discussed in #280


jaridmargolin:
I think this is more or less a duplicate of #3038

chhschou:
Cool thanks @jaridmargolin I didn't see it referenced to the issue. I'll close this.

Issue #3060, opened 2018-05-25
"yarn run dist" produces an invalid binary on mac with latest canary


MisterTea:
Upon running the canary binary produced by yarn run dist, I get this:

![image](https://user-images.githubusercontent.com/306879/40569837-d301fcfa-6039-11e8-9224-abebbaec7994.png)

tjwelde:
This relates to this PR: https://github.com/zeit/hyper/pull/2866. I commented there with two solutions, but I need feedback.

jcrben:
We should also update the test to actually test that a valid binary is produced - right now it's just hanging with no success or failure https://github.com/zeit/hyper/blob/canary/test/index.js#L11

LabhanshAgrawal:
seems to be fixed

PR #3065, opened 2018-05-30
macos build failure should show

jcrben:
Trying to reproduce build failure noted in https://github.com/zeit/hyper/issues/3060


albinekb:
But how are you making it test the release version of the app? Is it already building the release version of Hyper.app and running it in CI? 🤔 I don't think so?

PR #3067, opened 2018-06-04
Fix module hijacking

chabou:
Regression was introduced by #2866. Release version of Hyper failed to start.
Requiring react(-dom) in main process makes no sense.

### DIFF
--- a/app/plugins.js
+++ b/app/plugins.js
@@ -49,24 +49,19 @@ config.subscribe(() => {
 // so plugins can `require` them without needing their own version
 // https://github.com/zeit/hyper/issues/619
 function patchModuleLoad() {
-  const React = require('react');
-  const PureComponent = React.PureComponent;
-  const ReactDOM = require('react-dom');
-
   const Module = require('module');
   const originalLoad = Module._load;
   Module._load = function _load(modulePath) {
     // PLEASE NOTE: Code changes here, also need to be changed in
     // lib/utils/plugins.js
     switch (modulePath) {
-      case 'react':
-        return React;
-      case 'react-dom':
-        return ReactDOM;
-      case 'hyper/component':
-        return PureComponent;
+      // case 'hyper/component':
+      //   return PureComponent;
       // These return Object, since they work differently on the backend, than on the frontend.
       // Still needs to be here, to prevent errors, while loading plugins.
+      // case 'react':
+      // case 'react-dom':
+      case 'hyper/component':
       case 'hyper/Notification':
       case 'hyper/notify':
       case 'hyper/decorate':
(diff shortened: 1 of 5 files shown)


app/plugins.js
```diff
@@ -49,24 +49,19 @@ config.subscribe(() => {
...
     // PLEASE NOTE: Code changes here, also need to be changed in
     // lib/utils/plugins.js
     switch (modulePath) {
-      case 'react':
-        return React;
-      case 'react-dom':
-        return ReactDOM;
-      case 'hyper/component':
-        return PureComponent;
+      // case 'hyper/component':
```
albinekb: Shouldn't we try to revert this file to what it was before it changed? So that we won't get regressions with plugins that relied on this weirdness? 🤔

tjwelde:
@albinekb Please see https://github.com/zeit/hyper/pull/2866 to see, why it is necessary for plugins to have this.
Since main and render process hooks are located in one file at the plugin side, plugins relying on module.load hijacking on the render side fail, when they load at the main side.

chabou:
@albinekb This is the (better) error handling in our v2 that prevent these regular plugins to be loaded.

Question is: did these plugins really need this weird module hijacking?

tjwelde:
I don't know what will be the issue further down the road, when react versions differ...

chabou:
But if we decide to break this module hijacking, we need to deprecate this usage with some warning before cutting it.

tjwelde:
Yes. Someone introduced module hijacking. Is he reachable? What was the reasoning?

chabou:
First module hijacking was introduced by https://github.com/zeit/hyper/commit/a3d5004c1342d65bcb3c6301bc932957ff51ceff

And I think that React version consistency is the main reason. And plugins size because React is mutualized.

Other related PRs: #977, #1261 and #1878

tjwelde:
With this PR, no error is thrown in the built version, but the hyperline plugin doesn't load.

chabou:
@tjwelde oh, you're right... We should keep `react(-dom)` but not really require it.

I'm trying to fix this.

chabou:
Ok same conclusion than https://github.com/zeit/hyper/pull/2866#issuecomment-390469645

> So, we either add react and react-dom also to app/package.json, or we decouple "frontend hooks" from "backend hooks" at the plugin side, e.g. we have two entry points in a plugin, one for the Electron hooks, one for the Renderer hooks.

Third solution would be to stop exposing React/ReactDom/PureComponent and let plugins that need it to embed it.

Regular plugins can decorate components without requiring React because React is passed as parameter to decorator functions.

I think that third solution is better. But we should evaluate how many plugins relay on this. (My guess is: only `hyperline` :trollface:)

For now, I will measure impact on performance/starting time of adding react to main process dependencies. It could be a good temporary solution

tjwelde:
All 3 solutions would be fine for me.

chabou:
I tested launch time. And this is 6% slower with embedding react.
I added some deprecation warning. And we need to remove them ASAP.

But this is ok for next canary

Issue #1612, opened 2017-03-08
Hanging when behind a proxy with authentication


julianonunes:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 10 Build 14393.693
- **Hyper.app version**: 1.2.1
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/julianonunes/43558260b32de2e8d19cbc7e6470e416
- **Relevant information from devtools** _(CMD+ALT+I on Mac OS, CTRL+SHIFT+I elsewhere)_: N/A
- **The issue is reproducible in vanilla Hyper.app**: Is Vanilla

## Issue
I just downloaded the app and installed, and when I run, after a few seconds it just freezes. I'm behind a corporate proxy using authentication and this often causes problems with Node.js applications that require internet access (but IMHO, I should be able to use Hyper in "disconnected mode").

Strike117:
same here

chabou:
Maybe same issue than https://github.com/zeit/hyper/issues/1829

PatMis16:
I am using Hyper 3 and I cannot install plugins because I am behind a proxy. I tried to enter proxy information into the configuration but it did not work.

bdubn:
There are several threads related to hyper.js proxy issues. What are the FQDNs and/or public IP ranges that are utilized by the hyper.js update mechanism? If we had this information, we would be able to exclude/bypass proxy for this process.

Lunchb0ne:
The update-checker process checks for a feed from https://VERSIONTYPE.hyper.is/update/PLATFORM/VERSION
Where VERSIONTYPE is the type of the app (release or releases-canary)
PLATFORM is the platform of the app
VERSION is the version of the app (eg 3.0.9)

So for example, if I want the feed for Linux stable, I would need to go to

https://releases.hyper.is/update/deb/3.1.0-canary.4

enhsaihan-d:
I have same problem behind a corporate proxy even with proxy configured within npm.

jbrown16:
Can this issue get addressed again please? I am having the same issue as my network uses a corporate proxy. Are @jynolen's changes valid?

digeomel:
Same here 😞

adityamaliyan:
is this still on?
can I work on this?

PR #3108, opened 2018-06-29
Replace node-fetch with electron-fetch

mdogadailo:
When a user starts```hyper```, it sends the request to ```https://hyper-news.now.sh``` from [notifications.js](https://github.com/zeit/hyper/blob/canary/app/notifications.js#L19) file. If the user is  behind the proxy server, this request is going to fail, because in this case, we have to find
- if the user using the proxy
- the configuration of this proxy 
- and credentials

(another scenario from Linux auto update)

This pull request replace [node-fetch](https://github.com/bitinn/node-fetch) with [electron-fetch](https://github.com/arantes555/electron-fetch). 

The ```electron-fetch``` repeats API from ```node-fetch``` (it was forked from last one), but it's using Electron [net](https://github.com/electron/electron/blob/master/docs/api/net.md) module, so ```hyper``` don't need to care about proxy configuration. 
The ```net``` module provides 
- Automatic management of system proxy configuration, support of the wpad protocol and proxy pac configuration files.
- Automatic tunneling of HTTPS requests.
- Support for authenticating proxies using basic, digest, NTLM, Kerberos or negotiate authentication schemes.


albinekb:
Will this solve #1612 ?

mdogadailo:
@albinekb I'll try to reproduce #1612 next week if it solves that, I'll reply in #1612.

mdogadailo:
Dear @albinekb I can't reproduce issue from #1612, looks like there are not a proxy issue. 
I tried to replace response  (it's possible that proxy server just replace content) from https://hyper-news.now.sh with 503 and 404.
I tried to made redirect. I tried to create 30 sec timeout on response. All good, just error to console.

PR #3117, opened 2018-07-04
Fix issue when contents are present above module.exports in .hyper.js

### LINKED ISSUE #3076: `hyper install` crashes if .hyper.js has other lines above the `module.exports`

- **OS version and name**:  macOS 10.1.4
- **Hyper.app version**:  2.0.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/legodude17/de5c254a5087f1e8f6e3cfb27c04faa2
- **Relevant information from devtools** _(CMD+ALT+I on macOS, CTRL+SHIFT+I elsewhere)_:  N/A

## Issue
`hyper install` crashes if if you include other JS than just the `module.exports = ` at the top.

astronomersiva:
Fixes  #3076.

### DIFF
--- a/cli/api.js
+++ b/cli/api.js
@@ -44,7 +44,18 @@ const getFileContents = memoize(() => {
 
 const getParsedFile = memoize(() => recast.parse(getFileContents()));
 
-const getProperties = memoize(() => getParsedFile().program.body[0].expression.right.properties);
+const getProperties = memoize(() => {
+  let parsedFile = getParsedFile();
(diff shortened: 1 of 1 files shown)


astronomersiva:
I would really appreciate some eyes on this @timothyis.

albinekb:
We're thinking of moving to json5 format and not allow config to be in JS because things like these... thoughts @chabou

Stanzilla:
I'll close this since I actually think switching formats is the best way forward

PR #3118, opened 2018-07-06
add background to system-context-menu

LevitatingBusinessMan:
I started using Hyper recently and I think it's amazing. I did notice though, that the Open Hyper Here explorer option was only enabled on directories themselves. So I decided to add a background instance myself. This allows the same option to open up when right-clicking inside directories/desktop.

### DIFF
--- a/app/system-context-menu.js
+++ b/app/system-context-menu.js
@@ -1,7 +1,11 @@
 const Registry = require('winreg');
 
 const appPath = `"${process.execPath}"`;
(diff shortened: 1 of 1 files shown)


Stanzilla:
@LevitatingBusinessMan is this still needed? I can currently right click inside directories and get the "Open Hyper here" option?

LevitatingBusinessMan:
In that case I guess not, the code is pretty badly written anyway. I'll be closing it.

PR #3123, opened 2018-07-12
Add examples of Cygwin and Git Bash

pd4d10:


BananaAcid:
It should also include wsl.exe or wsl's bash.exe

Issue #3025, opened 2018-05-16
Ctrl + Backspace do not delete the word


hknielsen:
I would like the ability to `Ctrl + Backspace` / `Ctrl + Delete` to delete text in Hyper as possible in other terminals / Text Editors.

OS: Windows
Hyper version: 2.0 stable

rfgamaral:
Is <kbd>Ctrl</kbd>+<kbd>Backspace</kbd> / <kbd>Ctrl</kbd>+<kbd>Delete</kbd> suppose to delete word-by-word the text on the right and left respectively, or something else?

Because that's the current behavior on my Hyper terminal running Windows Subsystem for Linux. Maybe it's a shell thing? Which shell are you using?

FirstWorldSlardar:
"Because that's the current behavior on my Hyper terminal running Windows Subsystem for Linux." for me too :)

hknielsen:
Im using powershell as shell. I normally used ComEmu with Powershll before, and there it works

hknielsen:
@rfgamaral "Is Ctrl+Backspace / Ctrl+Delete suppose to delete word-by-word the text on the right and left respectively, or something else?"
Yes it should. But this is fx 3 words: `Something-Else`
I normally compare what Sublime do. But mostly this is also how fx ConEmu handle words

cwadrupldijjit:
As a note, the <kbd>Ctrl</kbd>+(<kbd>Backspace</kbd>/<kbd>Delete</kbd>) does work in PowerShell's terminal itself, just when I've tried it with Hyper, it doesn't do it.  I'm not sure if that is something that bash with WSL automatically includes or if it's something that just works with the way Hyper's set up.  I'd guess it's probably the first.

Which means that it's possible that PowerShell doesn't include a case for <kbd>Ctrl</kbd>+(<kbd>Backspace</kbd>/<kbd>Delete</kbd>), but the terminal does.  I am not sure if that is actually the case or not, but it's an initial guess.

mallochio:
Same here. Key combination works on standard powershell but not from powershell within Hyper.

enbandari:
Still the same.

OS: Windows
Hyper version: 3.0.2 stable
Shell: Powershell

malik446644:
me too .. i have the same probleme .. please fix it fast

ravindersaluja:
[...]

Issue #2873, opened 2018-04-18
Ctrl+Arrow Key and Ctrl+Shift Selection no longer work on Windows


sbatten:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 10 x64 1709
- **Hyper.app version**: 2.0.0

## Issue

Using Ctrl + Arrow Keys on windows used to control movement as expected. Now it outputs characters. Combining this with shift should select the text but now does nothing.

Tyriar:
ctrl+shift+arrows should work, a Hyper keybinding might be intercepting it?

oujunhao:
I don't know if this helps, but `Ctrl+Left` inserts `b` and `Ctrl+Right` inserts `f`.

Xapphire13:
@Tyriar, those commands (at least on my machine) control tab selection. I think @sbatten is referring to navigating through input text

Tyriar:
@Xapphire13 yeah, my statement stands; Hyper is intercepting it so it's not being interpreted by the shell.

lordspamalot:
Same problem here. 

Using Hyper 2.0.0 and Windows 10 1709.

If i use the key combination "ctrl+left arrow" then a "b" is written and if i use the combination "ctrl+right arrow" a "f" is written.

This behavior also occurs when I execute the function via the menu (Edit > Move to > Prevoius/Next Word).

Adjusting the keyboard shortcuts didn't help either:

```
"editor:movePreviousWord": "ctrl+left",
"editor:moveNextWord": "ctrl+right",
```

or;

```
"editor:movePreviousWord": "alt+left",
"editor:moveNextWord": "alt+right",
```

jnpwly:
For me, using Hyper 2.0.0 (stable), and no plugins, on Windows 10 Professional (x64) Version 1803 (build 17134.228), the key combinations [ALT]+[Left Arrow], [ALT]+[Right Arrow] move backwards and forwards through the entered text. I would *expect* [CTRL]+[Left Arrow] and [CTRL]+[Right Arrow] to do this, but as others have mentioned above, [CTRL]+[Left Arrow] writes out a `b` and [CTRL]+[Right Arrow] writes out an `f`.

DSlatkin:
[...]

PR #3146, opened 2018-08-01
Windows Keybindings was not correctly setup

hknielsen:
I created https://github.com/zeit/hyper/issues/3025 because on Windows, using Powershell as shell, key's didnt work for fx. move word.

What I saw was, "alt+arrows" worked to some degree, but not the expected keymappings defined in `win32.json`.
So for some reason the beneath didnt work, and if evaluated directly in Paths.js it did. Dont know enough of nodejs to say what are happening. But this fixed it. Could be a Windows specific NodeJs bug.

If you got any better way of doing this let me know.

### DIFF
--- a/app/config/paths.js
+++ b/app/config/paths.js
@@ -45,18 +45,22 @@ const darwinKeys = join(keymapPath, 'darwin.json');
 const win32Keys = join(keymapPath, 'win32.json');
 const linuxKeys = join(keymapPath, 'linux.json');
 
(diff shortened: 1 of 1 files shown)


DSlatkin:
Commented more on the issue [here](https://github.com/zeit/hyper/issues/2873#issuecomment-420769458) regarding this fix in the context of the central issue at hand.

Stanzilla:
@hknielsen willing to pick this up again?

Issue #2697, opened 2018-02-22
Support for vertical selection


kenorb:
- [X] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [X] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: 10.13.2 (17C2120)
- **Hyper.app version**: Hyper 1.4.8 (stable)

## Issue

In _Terminal_ app I could press _Alt_ key and make a vertical selection, similar in iTerm. E.g.

<img width="194" alt="screen shot 2018-02-22 at 16 01 22" src="https://user-images.githubusercontent.com/266306/36549109-dbd38b02-17e9-11e8-8129-a22e3c69241d.png">

However vertical selection is not supported in Hyper.

chabou:
In our v2, selection is handled by xterm.js.
Thus, we need to implement this in xterm.js repository.

Stanzilla:
Hey there, to verify if this is indeed a Hyper issue or upstream at https://github.com/xtermjs/xterm.js/ could you please test if another xterm.js powered terminal shows the same problem you are experiencing? I suggest using https://github.com/Microsoft/vscode for testing. Thanks!

kenorb:
Seems it's fixed now in [upstream](xtermjs/xterm.js/issues/1300).

Tyriar:
You'll get this just by upgrading to 3.5 which was just released 👍, though you will need to expose `macOptionClickForcesSelection` if you want to enable alt+click to force selection on macOS (more info at https://github.com/xtermjs/xterm.js/releases/tag/3.5.0).

LabhanshAgrawal:
fixed in #3255

PR #3178, opened 2018-08-16
Upgrade to xterm 3.6.0 for column selection. Refs #2697

jocull:
This includes the xterm version bump required to enable the column selection feature. See #2697 for more details. I did not add anything for the `macOptionClickForcesSelection` option mentioned, but it seems to work out-of-the-box just fine on my mac.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6693,9 +6693,9 @@ xtend@~2.1.1:
   dependencies:
     object-keys "~0.4.0"
 
-xterm@3.4.1:
-  version "3.4.1"
-  resolved "https://registry.yarnpkg.com/xterm/-/xterm-3.4.1.tgz#12452979ea2db3371f8195832d7407abba988980"
+xterm@3.5.1:
+  version "3.5.1"
+  resolved "https://registry.yarnpkg.com/xterm/-/xterm-3.5.1.tgz#d2e62ab26108a771b7bd1b7be4f6578fb4aff922"
 
 y18n@^3.2.1:
   version "3.2.1"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -6693,9 +6693,9 @@ xtend@~2.1.1:
   dependencies:
     object-keys "~0.4.0"
 
-xterm@3.4.1:
-  version "3.4.1"
-  resolved "https://registry.yarnpkg.com/xterm/-/xterm-3.4.1.tgz#12452979ea2db3371f8195832d7407abba988980"
+xterm@3.5.1:
+  version "3.5.1"
+  resolved "https://purecloud.artifactoryonline.com/purecloud/api/npm/inin-internal-npm/xterm/-/xterm-3.5.1.tgz#d2e62ab26108a771b7bd1b7be4f6578fb4aff922"
```
Stanzilla: wat

jocull: Bah, internal repos. Let me fix

jocull: Updated - thanks for the catch 👍

Tyriar:
You can probably update to 3.6 just as easily, it was mostly a stability release https://github.com/xtermjs/xterm.js/releases/tag/3.6.0

jocull:
I’m not opposed. I only did 3.5.1 this time because 3.5 was minimum with the needed feature and may as well pick up the .1 for any bug fixes :)

Tests did not complain and everything seemed fine with 3.5.1 - not sure if there’s anything else that really needs to be done.

JCMais:
merging this will allow us to use font ligatures again on Hyper

Tyriar:
I would actually wait for 3.7.0 as there was a pretty bad perf regression in either 3.5 or 3.6: https://github.com/xtermjs/xterm.js/issues/1620

PR #3193, opened 2018-08-26
Conventional paths

miyaokamarina:
As described [at MSDN](https://msdn.microsoft.com/en-us/library/bb762584(VS.85).aspx#FOLDERID_RoamingAppData) (with more detailed explanation [here](https://msdn.microsoft.com/en-us/library/bb762494%28VS.85%29.aspx?f=255&MSPPError=-2147217396#CSIDL_APPDATA)), at [XDG](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) and by [Apple](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/MacOSXDirectories/MacOSXDirectories.html), it's recommended to store user-specific applications data in specific subdirectories rather than directly at `$HOME` to avoid bloating it with dozens of configs and caches.

Unfortunately, Hyper doesn't respect these guidelines at the moment and bloats ~.

To avoid this without breaking changes, I propose to move default configs location to the platform-specific path, but if no data present at new path, try to load it from legacy locations. When loaded, configs will be moved to right location. If data is present at both old and new locations, new location takes priority, and the warning will be displayed at console.

### DIFF
--- a/app/config/paths.js
+++ b/app/config/paths.js
@@ -1,10 +1,10 @@
 // This module exports paths, names, and other metadata that is referenced
 
-const { homedir, platform } = require('os');
-const { statSync } = require('fs');
-const { resolve, join } = require('path');
+const {homedir} = require('os');
+const {statSync} = require('fs');
+const {resolve, join} = require('path');
 
-const { app } = require('electron');
+const {app} = require('electron');
 
 const isDev = require('electron-is-dev');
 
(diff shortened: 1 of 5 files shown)


yochem:
Is this still something that is being considered?

Stanzilla:
@miyaokamarina is this still a problem with the current version (`canary`)?

PR #3198, opened 2018-08-30
Simplify promises in `cli/api.js`

itaisteinherz:
Note to reviewers: the PR is just a simple code refactor, nothing significant.

### DIFF
--- a/cli/api.js
+++ b/cli/api.js
@@ -90,40 +90,32 @@ function getPackageName(plugin) {
 
 function install(plugin, locally) {
   const array = locally ? getLocalPlugins() : getPlugins();
-  return new Promise((resolve, reject) => {
-    existsOnNpm(plugin)
-      .then(() => {
-        if (isInstalled(plugin, locally)) {
-          return reject(`${plugin} is already installed`);
-        }
+  return existsOnNpm(plugin)
+    .then(() => {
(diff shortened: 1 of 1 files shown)


cli/api.js
```diff
@@ -90,40 +90,32 @@ function getPackageName(plugin) {
...
-        if (statusCode && (statusCode === 404 || statusCode === 200)) {
-          return reject(`${plugin} not found on npm`);
-        }
-        return reject(`${err.message}\nPlugin check failed. Check your internet connection or retry later.`);
-      });
-  });
+  return existsOnNpm(plugin)
+    .then(() => {
+      if (isInstalled(plugin, locally)) {
+        return Promise.reject(`${plugin} is already installed`);
```
albinekb: Is't this the same as ```throw new Error(`${plugin} is already installed`)```

itaisteinherz: No, using [`Promise.reject`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject) returns a rejected promise, but we could do
```js
return Promise.reject(new Error(`${plugin} is already installed`));
```
instead.

itaisteinherz:
@albinekb I'm pretty sure the PR is ready to be merged. Please reply if you have anything you'd like me to change

Issue #3205, opened 2018-09-04
No way to get hyper version from command line


chadasapp:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: 10.14 Beta (18A377a)
- **Hyper.app version**: 2.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: N/A
- **Relevant information from devtools** _(CMD+ALT+I on macOS, CTRL+SHIFT+I elsewhere)_: N/A
- **The issue is reproducible in vanilla Hyper.app**: Is Vanilla

## Issue
`hyper --version` or `hyper version` should produce the version number alone and exit cleanly. Should be pretty easy.

albinekb:
PR welcome if it's pretty easy!

hayer:
@albinekb PR submitted. Propose to use #3207 as it prints only the version which makes it a bit easier to use in scripts checking version - so you don't need to remove the prefix string.

hayer:
PR got merged at least, so yes? But I am unsure.

LabhanshAgrawal:
tested on canary and it's working

PR #3207, opened 2018-09-06
added 'version' command.

hayer:
Adds the possibility to write `hyper version` which will print the version defined in `app/package.json` `version`.

Ready for merge - at least I think so?

Related to issue #3205 .

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -3,6 +3,7 @@
 const {spawn, exec} = require('child_process');
 const {isAbsolute, resolve} = require('path');
 const {existsSync} = require('fs');
+const {version} = require('../app/package');
 const pify = require('pify');
 const args = require('args');
 const chalk = require('chalk');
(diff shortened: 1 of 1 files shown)


hayer:
I but the `const {version} = require('../app/package');` at the top as I didn't see any "middle-file includes" anywhere else.

cli/index.js
```diff
@@ -133,6 +134,12 @@ args.command(['d', 'docs', 'h', 'home'], 'Open the npm page of a plugin', (name,
   process.exit(0);
 });
 
+args.command(['version'], 'Show the version of hyper', () => {
+  checkConfig();
```
chabou: This command doesn't need a valid config file.
I would remove this check.

hayer: @chabou done.

chabou:
@hayer Thank you so much for your help.

Issue #1229, opened 2016-12-15
need better instructions for Linux install


anthonyettinger:
is it possible to get a ppa for Ubuntu based systems?

pablopunk:
@anthonyettinger What's wrong with the `.deb` installer? 😜

anthonyettinger:
I do not see one. Where is it documented?

ppot:
@anthonyettinger https://hyper.is/#installation

anthonyettinger:
A ppa would allow me to get updates though. installing manually with .deb is not ideal.

toddhalfpenny:
If it's a PPA, then perhaps this is a dupe of https://github.com/zeit/hyper/issues/1115 ?

gluons:
Can it be easier by [Snaps](https://snapcraft.io/)? Cross-linux system installation.

albinekb:
Yes. We were working on creating a snap but we didn't get around to. There's not much demand and supporting a snap would be extra effort for little usability, but if someone knows how, feel free to PR the snap build scripts and we can work together on getting it out there 🎉

gluons:
OK. I'm not serious about that.
I hope one day I can install Hyper via command line easily. PPA or Snaps are fine. 😄 
Love Hyper anyway. 😘

philpalmieri:
If no one else is taking this on, i'm going to start playing with it to see if i can get it running as a snap, please leave this open.

albinekb:
@philpalmieri that would be amazing, tell me if there's anything we can do to help you 🙏

philpalmieri:
I stayed up a bit last night and made a simple Node Snap - https://snapcraft.io/insultme - its actually pretty straight forward, and there are build pipline tools out of the box - ill spend some more time on it tonight and see if i can get a demo running in a dev branch.

vinay72:
Yes , it is possible to get a ppa for Ubuntu Based Systems.

EfstathiadisDimitris:
If this isn't resolved. I would like to work on it. Just tell me, if someone else is working on it..

philpalmieri:
I worked on it for a bit, but got sidetracked with work - I had it compiling to a snap - but having electron issues - go for it! - I won't be able to come back to it for a bit (might try flatpak soon though)

PR #3216, opened 2018-09-10
Add snapcraft build

pablopunk:
Not sure why this was removed at some point but I got it working on my pc. The only thing left is to upload the snap to snapcraft (https://docs.snapcraft.io/build-snaps/electron) and set up the github integration for automatic builds and updates (https://snapcraft.io/build)

This is related to https://github.com/zeit/hyper/issues/1229

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ addons:
       - snapd
 
 before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo snap install snapcraft --classic; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then export DISPLAY=:99.0; sh -e /etc/init.d/xvfb start; sleep 3; fi
 
 cache: yarn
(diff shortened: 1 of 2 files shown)


albinekb:
How do I go about testing this? @pablopunk just running the build and it should build the snap too? What platforms can it be built on? Ubuntu only?

pablopunk:
@albinekb 

Yep, `npm run dist` should get you a snap inside `dist/` folder. And then install it with `sudo snap install ./dist/...`. Of course you need `snapd` to do that.

The good thing about snapcraft.io is that the packages are universal, so they work across distros. I did try it only in Ubuntu but it should work okay in others...

pablopunk:
@chabou hey! Thanks for the merge. One more thing, when are we getting the snap uploaded to [snapcraft.io](https://snapcraft.io/)? Someone should link this repo to snap (see https://snapcraft.io/build)

chabou:
I (or any helper) will upload the snap after next stable release

PR #3245, opened 2018-10-08
Minor formatting changes

szincone:
- Made slight changes in the config-store.prod.js file for better style consistency with the surrounding code. 
- Updated package.json file for better readability.


chabou:
#### `config-store.prod.js`

If we want to enforce this, we need to add this to eslint. Maybe https://eslint.org/docs/rules/arrow-body-style?

There are 2 advantages to use a rule: consistency and autocorrect (combined to prettier).

I would use a `as-needed` value for this rule.

#### `package.json`

We need to enforce this (line-width) in prettier.

PR #3277, opened 2018-10-31
Fix OS X 'reload' accelerator key mappings

### LINKED ISSUE #921: Prevent CMD+R from refreshing and losing progress (or remove refresh)
I sometimes alt+tab into Hyper and accidentally refresh, which kills all my tabs, running process, and is generally super painful. Could we have an option that asks whether you're sure you want to reload?

olsgreen:
Changed OS X 'reload' accelerator key mappings be more in line with Linux & Windows and further away from browsers.

For people constantly switching between browser and command line, having `Cmd+R` nuking output in your hyper session is a killer...

Fixes #921

### DIFF
--- a/app/keymaps/darwin.json
+++ b/app/keymaps/darwin.json
@@ -1,7 +1,7 @@
 {
   "window:devtools": "command+alt+i",
-  "window:reload": "command+r",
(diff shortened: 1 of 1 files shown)


rauchg:
This is huge. Thanks @olsgreen

PR #3284, opened 2018-11-10
Fix CLI tool with python3

MatthiasWinkelmann:
I realise having ```python``` linked to ```python3``` is currently discouraged, but being compatible does not hurt, and may avoid problems whenever it becomes more common for python 3 to be the default.


rauchg:
Thanks!

PR #3290, opened 2018-11-19
FIX markdown hyperlinks

robertgrzonka:
Some of the hyperlinks were written with wrong `Markdown syntax`; should be `[link](https://example.com)` but was (link)[https://example.com]`

Ready to merge.


chabou:
Thank you for your help 🙏

PR #3324, opened 2018-12-12
Add tray icon and set single instance mode to improve startup time on Windows

phungtuanhoang1996:
This is my attempt to address ther slow startup time on Windows, and also my first attempt to contribute to an open-source project (because I really like Hyper)

Please feel free to review and comment on what can be improve / went wrong. Thanks! 😄

Changed hyper to be running single instance mode on windows
Added a tray icon with exit option and new window option

Always Preloads a hidden window (without rpc handler and terminal session)
When a new window needs to be opened, the preloaded window will have its rpc and terminal session attached and brough to front

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -1,3 +1,18 @@
+const {app} = require('electron');
+
+// Multiple instance check on windows
+if (process.platform === 'win32') {
+  const isSecondInstance = app.makeSingleInstance((commandLine, workingDirectory) => {
+    // When tried to run a second instance, load a preloaded window instead on the main instance
+    app.loadPreloadedWindow(workingDirectory);
+    app.createPreloadedWindow();
+  });
+
+  if (isSecondInstance) {
+    app.quit();
+  }
+}
+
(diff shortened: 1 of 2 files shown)


chabou:
🙏 @phungtuanhoang1996 awesome work!

I will review this ASAP.
Thank you so much for your help.

My first thought was: is this a common behavior on Windows to not quit an app? Maybe we should add an opt-out option?
I remember that other apps (like Skype) offer a way to really quit the app (in context menu on tray icon, if I remember correctly).

phungtuanhoang1996:
I did include an option to really quit the app in the context menu.
I can make an opt-out option as well.

chabou:
Oh... I didn't see the systray context menu. Sorry.

For the option, let see if user are complaining about this.
This is the Hyper behavior on macOS and this is totally ok to me.

chabou:
We are really honored to merge your first open-source contribution 🙏 
Thank you!

PR #3325, opened 2018-12-12
Upgrade electron to v3

juancampa:
Bumping electron version to v3 in preparation for Hyper v3. Electron v4's seems to be coming along so we might end-up bumping to v4 before the release if it becomes stable.

I'm running a few sanity checks to verify that nothing broke. This shouldn't be merged until everything is checked off.

 - [x] No errors on output
 - [x] Plugins work (tested installing and using https://github.com/nyx-team/hyper-nyx)
 - [x] Multiple tabs
 - [x] Multiple windows
 - [x] Splits
 - [x] Menus
 - [x] Auto update (pop up shows)
 - [x] Clean shutdown
 - [x] Vim session with splits
 - [x] Clipboard works
 - [x] Output errors are expected
 - [x] Initial zoom is the same as canary and pinch remains disabled
 - [x] Performance is at least on par with electron v1
   - [x] Throughput (2% slowdown, possibly due to different zoom levels)
   - [x] Refresh rate
 - [x] Sanity macOS (tested by @chabou)


.travis.yml
```diff
@@ -6,7 +6,7 @@ language: node_js
 matrix:
   include:
     - os: linux
-      node_js: 8
+      node_js: 10.2.0
```
juancampa: There doesn't seem to be an equivalent for circleCI. Not sure if it's needed

juancampa:
I just realized that everything looks smaller on electron v3. It might be the reason for the perf degradation I'm seeing (more cells on screen)

juancampa:
Running a throughput test using this script:

```bash
#!/bin/bash

RUNS=5

for i in `seq 1 $RUNS`; do
    starttime=`date "+%s%3N"`

    # Generate a significant amount of output
    ls -Rl /usr/ /usr/ /usr/ /usr/

    endtime=`date "+%s%3N"`

    durations[$i]=$(($endtime-$starttime)) 
done

avg=0
for i in `seq 1 $RUNS`; do
    duration=${durations[$i]}
    avg=$(($avg + $duration))
    echo Duration: $(($duration/1000)).$(($duration%1000))s
done
avg=$(($avg/$RUNS))

echo Average: $(($avg/1000)).$(($avg%1000))s
```

Canary: Average: `9.341s`
This branch: `9.513s`

This branch seems to be ~2% slower, nothing obvious pops up on the profiler so it's really hard to tell what's going on. I don't think this is a big issue considering the perf improvements coming down the pipeline

juancampa:
This shows up when I click on "Download" on the update dialog:

```
[22543:22543:1212/161913.636090:ERROR:sandbox_linux.cc(379)] InitializeSandbox() called with multiple threads in process gpu-process.
```

Might be because I'm running the AppImage without installing it though

juancampa:
@chabou @rauchg this branch is ready to be merged. The one thing I need is for someone to test it on macOS since I don't have a mac anymore (destroyed last month by Florida's downpour :sweat:)

chabou:
I will test it on mac.

juancampa:
Closing in favor of #3329

PR #3328, opened 2018-12-14
do a yarn upgrade to add integrity file and upgrade major versions

zerdos:
Re-generate the yarn file with the newest version of yarn.
This bumps up the major versions of dependencies, also allow me to progress audit - and send later PR's with upgraded dependencies (right now yarn audit found 10 low vulnerabilities)


chabou:
Thank you for your help.
It will be merged soon. 🙏

chabou:
If I understand this correctly, you just upgrade yarn version?

How can we ensure that other contributors will use the same and don't overwrite your changes (intergrity-checks etC...)?

zerdos:
Add engines check to the package.json file can prevent the developers to use an old version, (without integrity checks)

I also ran a "yarn upgrade", which regenerates your yarn.lock based of the best semver match in that time when I was running it. 

If it goes trough on the CI, and some manual testing - then probably the app is still working.

PR #3329, opened 2018-12-15
Upgrade Electron to v4

juancampa:
Supersedes #3325 

Talked to @rauchg and we think it's better to skip v3 altogether, go straight for v4, and use the canary channel to test it.

Sanity checks:
 - [x] No errors on output
 - [x] Plugins work (tested installing and using https://github.com/nyx-team/hyper-nyx)
 - [x] Multiple tabs
 - [x] Multiple windows
 - [x] Splits
 - [x] Menus
 - [x] Auto update (pop up shows)
 - [x] Clean shutdown
 - [x] Vim session with splits
 - [x] Clipboard works
    - [x] Through menu
    - [x] Hotkeys (**broken but I added a work-around for now**)
 - [x] Initial zoom is the same as canary and pinch remains disabled
 - [x] Performance is at least on par with electron v1
   - [x] Throughput (2% slowdown, possibly due to different zoom levels)
   - [x] Refresh rate
 - [x] Sanity macOS
 - [ ] Sanity Windows
 - [x] Went through all [listed breaking changes](https://github.com/electron/electron/blob/master/docs/api/breaking-changes.md)


app/commands.js
```diff
@@ -53,6 +53,13 @@ const commands = {
       webContents.openDevTools({mode: 'detach'});
     }
   },
+  'editor:copy': focusedWindow => {
+    if (!focusedWindow) {
+      return;
+    }
+    const webContents = focusedWindow.webContents;
+    webContents.copy();
+  },
```
juancampa: Had to add this because `editor:copy` was not being fired with `ctrl+shift+c` on Linux/electron4. It's a bizarre bug since paste and cut still work, I'm opening an issue on the electron repo

app/auto-updater-linux.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-const fetch = require('electron-fetch');
+const fetch = require('electron-fetch').default;
```
chabou: This is due to `electron-fetch` version bump right?

app/commands.js
chabou: Can you add in comment what you explain here and the link to the issue?

lib/index.js
```diff
@@ -17,8 +17,9 @@ import {loadConfig, reloadConfig} from './actions/config';
 import HyperContainer from './containers/hyper';
 import configureStore from './store/configure-store';
 
-// Disable pinch zoom
-webFrame.setZoomLevelLimits(1, 1);
+// The default zoom was somehow changed with Electron 3 (or maybe 2). Setting
+// zoom factor to 1.2 brings back the normal default size
+webFrame.setZoomFactor(1.2);
```
chabou: Unfortunately, it appears zoomed on macOS 😞

app/commands.js
chabou: copy/paste still works on macOS

chabou:
We have a problem with titlebar:
image

app/auto-updater-linux.js
juancampa: Yes. I tried using `import` but it didn't work so I'm assuming this is not being transpiled

lib/index.js
juancampa: hmm weird, probably a linux-only thing. I'll investigate

juancampa:
Hmm ok, seems like given that v3 worked fine but v4-beta has issues on Mac (and maybe windows?), it might be better to go with v3 for now as opposed to writing temporary work-arounds, since they might get fixed by the electron team eventually.

@rauchg, thoughts on this?

Stanzilla:
4.0 is final as of today https://github.com/electron/electron/releases/tag/v4.0.0

lib/index.js
juancampa: Weird, I just tested on MacOS and this line fixes it

Stanzilla:
That looks like electron-rebuild needs an update for electron v4 @MarshallOfSound

Stanzilla:
ah, great, yarn was a bit annoying back in the day and would not upgrade that without electron-rebuild using it.

juancampa:
> ah, great, yarn was a bit annoying back in the day and would not upgrade that without electron-rebuild using it.

Yeah, I struggled with that too. The trick is to delete `node-abi` from `yarn.lock` and have yarn figure out what the latest version is again.

Here's the yarn issue: https://github.com/yarnpkg/yarn/issues/4986

juancampa:
> node version has to be updated for AppVeyor and CircleCI as well

@Stanzilla, not sure what you mean here. Is there anything else other than these two lines?

https://github.com/zeit/hyper/pull/3329/files#diff-354f30a63fb0907d4ad57269548329e3R9
https://github.com/zeit/hyper/pull/3329/files#diff-180360612c6b8c4ed830919bbb4dd459R13

Stanzilla:
@juancampa nope that's fine, I just did not see the appveyor change in your PR. For Circle, I don't know how they specify node versions, I did not set up that CI.

juancampa:
@Stanzilla, did you by any chance tested it on windows? I'm looking for someone to help me there, since I don't have a windows machine.

Stanzilla:
Sure thing! It looks like you forgot to address https://github.com/electron/electron/blob/master/docs/api/breaking-changes.md#appmakesingleinstance and it errors because of it.

juancampa:
@Stanzilla sorry for the delay, I just had to merge canary, we reverted the change that introduced that call. Feel free to test again when you get a few.

Stanzilla:
Does the config have a new location? it ignores what I do in ~/.hyper.js

juancampa:
> Does the config have a new location? it ignores what I do in ~/.hyper.js

It doesn't. Seems to work fine on Linux and MacOS (I just changed the font size)

Stanzilla:
font size works, shell does not,

juancampa:
Yeah, I'm seeing it too, but it doesn't seem to work on the canary branch either so it must be something else

Stanzilla:
QA's been on a long vacation :D

juancampa:
@Stanzilla, Ok, I just looked into it and it was introduced by my [optimistic session creation PR](https://github.com/zeit/hyper/pull/3352). Working on a fix

Stanzilla:
Sounds great!

Stanzilla:
well that was a quick and unexpected merge, but I can confirm that it works as expected now.

juancampa:
Optimistic merge

PR #3336, opened 2018-12-16
Performance - Data batching

juancampa:
## Problem

When the pty produces a lot of output in a very short period of time (e.g. `cat very-long-file`), many IPC `session data` messages are passed to the renderer. This causes a lot of thrashing (see profiler screenshot) due to `ipcListener` emitting a ton of chunks.

Each chunk is typically 4k but the vt can parse/render a lot more per frame (especially with the coming WebGL renderer). Each message is prefixed with the window's `uid`, which is then removed on the renderer process before parsing, adding GC pressure (probably) and CPU cost on both processes (we're CPU-bound on the renderer process due to VT parsing being very costly)

## Solution

Batch data in a buffer before sending it to the renderer.

## Benchmarks

(left: no batching, i.e. canary, right: batching, i.e. this branch)

`find ~`

| | canary | canary + webgl | this branch | this branch + webgl :star: |
|-|-----------|-----------|-----|---------------------------|
| Throughput | ~35s | ~36s :one: | ~35s | ~15s :two: |
| FPS | 3* then 17 :three: | 40+ | 12-16 :four: | 40+ |

This change makes a huge difference with the WebGL renderer enabled (compare :one: vs. :two:) because the renderer is focused on processing one big chunk instead of a ton of tiny ones + canvas rendering. Without WebGL, it makes the framerate less stuttery (compare :three: vs :four:)

[*]: while the benchmark process is still generating output

Process: Throughput is measured as the time between running the command until the next prompt is rendered. FPS are taken from DevTools renderer FPS widget.


app/session.js
```diff
@@ -21,8 +21,60 @@ try {
 
 const envFromConfig = config.getConfig().env || {};
 
+// Max duration to batch session data before sending it to the renderer process.
+const BATCH_DURATION_MS = 16;
+
+// Max size of a session data batch. Note that this value can be exceeded by ~4k
+// (chunk sizes seem to be 4k at the most)
+const BATCH_MAX_SIZE = 200 * 1024;
```
juancampa: I tried different values but this one (200k) had the best throughput. Using, e.g. 100k, can lower the throughput by ~25% :fire:

juancampa: A future improvement could determine this value dynamically

chabou:
Be careful, this PR contains #3329 commits (that should be merged before)

juancampa:
> this PR contains #3329 commits (that should be merged before)

Yeah, I didn't want to test this on canary since it might change the benchmark results. I made the warning more prominent :)

app/session.js
jerch: Few notes from my side:
Keep in mind that the 16ms stack on top of the input roundtrip latency and ppl might start complaining about lousy terminal response while typing. In the xterm.js demo app I found 5 ms helping alot while not yet being noticeable, 15 ms already introduced a tiny lag when typing really fast.

4k is the default pty buffer size on many linux distros, but you cannot rely on that (can be changed for every pty and system wide). BSDs (and macOS) use an adaptive buffer size typically ranging from 16k to 64k depending on the current data pressure.

app/session.js
```diff
@@ -21,8 +21,60 @@ try {
...
+// Data coming from the pty is sent to the renderer process for further
+// vt parsing and rendering. This class batches data to minimize the number of
+// IPC calls. It also reduces GC pressure and CPU cost: each chunk is prefixed
+// with the window ID which is then stripped on the renderer process and this
+// overhead is reduced with batching.
+class DataBatcher extends EventEmitter {
+  constructor(uid) {
+    super();
+    this.uid = uid;
+    this.decoder = new StringDecoder('utf8');
```
jerch: UTF8? hyper is already using binary transport for the pty? We might add this eventually to xterm.js in the future, had some good results with it lifting some burden from the input parsing.

app/session.js
jerch: Btw 200k seems quite big to me, you might want to test this with `yes`, it triggers a very exp. 'lineFeed' event in the terminal for every second byte. Additionally the event can be hooked by external handler, that might introduce even more workload (not sure if you allow plugins to consume it).

juancampa: > Keep in mind that the 16ms stack on top of the input roundtrip latency and ppl might start complaining about lousy terminal response while typing. In the xterm.js demo app I found 5 ms helping alot while not yet being noticeable, 15 ms already introduced a tiny lag when typing really fast.

I had the same concern (main reason I use kitty) but I haven't felt any lag so far. We'll have to keep an eye on feedback.

> 4k is the default pty buffer size on many linux distros, but you cannot rely on that (can be changed for every pty and system wide). BSDs (and macOS) use an adaptive buffer size typically ranging from 16k to 64k depending on the current data pressure.

Good point, I actually have to remove the "Note..." part, It was happening in my origin implementation, but it didn't feel right to go over a "max", so I fixed it to never exceed `BATCH_MAX_SIZE`

juancampa: > what strategies would you use for dynamic computation of this value? Maybe we can create an RFC issue?

I was thinking the parser could generate "back pressure". Basically "Hey, the chunks you are sending are too big for me to parse and render in one frame, slow down". This could be less important if we [run the parser on a WebWorker](https://github.com/xtermjs/xterm.js/issues/1515)

app/session.js
juancampa: Not sure I follow this one. IIUC electron can't send `Buffer` objects to the renderer process so we must turn the input into utf8. If we move parsing to a worker, we might be able to use `SharedArrayBuffer` to minimize conversions

jerch: But then why are you dealing with UTF8 at all? node-pty is supposed to return JS strings already, if encoding is omitted (defaults to UTF8 being decoded by the underlying `Socket`).

So how do you send data over to the window process? This is not able to handle binary data?

app/session.js
juancampa: > Btw 200k seems quite big to me, you might want to test this with yes, it triggers a very exp. 'lineFeed' event in the terminal for every second byte

I just tested `yes` and the buffering definitely doesn't help (drops FPS from ~50 to ~30). Basically 200k is a good value for `find ~` which is arguably a more common output than `yes`. This is all more evidence for implementing @jerch's proposal here: https://github.com/xtermjs/xterm.js/pull/1818#issuecomment-445457198

jerch: Yeah I think its better suited in `Terminal.write`. We already do xon/xoff flow control there, imho its better to keep it in one place. Otherwise every buffering step involved would have to implement flow control on its own to get a working back pressure chain, which gets quite cumbersome with node's async nature.

app/session.js
juancampa: > but gets BASE64 encoded and a JSON envelope

Exactly :(

> falling back to websockets

IPC doesn't seem to cost that much though, V8 strings are pretty fast. In this profile of one frame, IPC only takes a tiny sliver (<0.5ms), not a lot of gains there. It might become more prominent when we do WebWorkers though, with an extra hop, but then again, we'll probably use `SharedArrayBuffer` for that

image

juancampa: I just realized that real gains are hidden in the parsing of strings part (i.e. your fix https://github.com/xtermjs/xterm.js/pull/1796), so ignore that last part :sweat_smile:

jerch: Yeah thats the basic idea behind my optimization plan (#1811) - by getting of any string handling in the input chain and working directly on arraybuffer we can lower the runtime needed for input handling by >50%.
This gets even more if we change to direct utf8 binary input as it would save the nonsense conversions. If you are interested I have a heavily optimized playground branch, where you can test it in the demo app (https://github.com/jerch/xterm.js/tree/tttt, also needs https://github.com/jerch/node-textdecode)

jerch: If you cannot get it run, here are my results: https://github.com/xtermjs/xterm.js/issues/791#issuecomment-447993547 :smile:

juancampa:
@chabou @rauchg now that electron 4 is merged, this is ready to go, if you guys approve.

PR #3352, opened 2018-12-19
Optimistic session creation

juancampa:
## Problem
Launch takes too long

## Solution

Not a complete solution but an improvement.

Instead of waiting for the window to initialize then *then* send a `new` message so that we can create a new session. The session is optimistically created so that when the `new` messages is received, a session is already warmed up

## Benchmarks

This measures the time from when `app/index.js` first runs until the first `session data` message is sent to the renderer. Ran it 5 consecutive times.

canary: `749, 728, 765, 777, 762`
this branch: `617, 656, 615, 590, 596`

Averages:

| canary | this branch | diff |
|----|----|----|
| 756ms | 614ms | 142ms |

Not a hell of a lot but doesn't hurt either. I've only tested on Linux, but [should help even more on other platforms](https://stackoverflow.com/a/51396188/653799)

### DIFF
--- a/app/ui/window.js
+++ b/app/ui/window.js
@@ -91,16 +91,16 @@ module.exports = class Window {
         const uid = uuid.v4();
         const session = new Session(Object.assign({}, options, {uid}));
         sessions.set(uid, session);
-        return { uid, session };
+        return {uid, session};
       }
 
       // Optimistically create the initial session so that when the window sends
       // the first "new" IPC message, there's a session already warmed up.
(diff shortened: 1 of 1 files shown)


rauchg:
Succinct and effective!

juancampa:
I also did a video recording (thanks @chabou!) to verify that the actual "time to prompt" was improved and got results that are consistent with the measurements above: ~150ms speed-up on Linux

PR #3362, opened 2018-12-22
Revert "Add tray icon and set single instance mode to improve startup…

chabou:
Reverts zeit/hyper#3324

#3324 PR creates two windows on start: this PR tries to hide a preloaded window. But Hyper automatically hides every window (except on debug and dev) on creation and shows every window after its RPC initialization: https://github.com/zeit/hyper/blob/76ab6abd682327a76455d5661c57e31dc2b0d48f/app/ui/window.js#L63

In addition, this PR introduces some code duplication that could be avoided.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -1,18 +1,3 @@
-const {app} = require('electron');
-
-// Multiple instance check on windows
-if (process.platform === 'win32') {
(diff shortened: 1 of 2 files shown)


juancampa:
> @juancampa I preserve your launch optimistic session creation (#3352) but can you confirm that everythings is ok?

LGTM

Issue #3340, opened 2018-12-17
2.1.0 - Broken in OSX - Copy / Paste File or Directory to get full path in terminal


jq-87:
- **OS version and name**: OSX 10.13.6
- **Hyper.app version**: 2.1.0
- **Relevant information from devtools** : this.term.send is not a function at Term.onWindowPaste

This feature used to work, was addressed here -> [1285](https://github.com/zeit/hyper/issues/1285)

## Issue
Steps to reproduce:
1. Copy File or Directory from finder (command + c)
2. Try pasting the file/directory into hyper (command + v) - this is when the error in console appears

What should happen after step 2:
File/Directory path SHOULD display in terminal but currently it does not anymore.

kgroat:
This has only been fixed in `3.0.0-canary` versions, and is still an issue in `2.1.1`.  Would it be possible to get #3383 cherry picked into `master`?

jq-87:
I what version will this feature be restored? still not working in `2.1.2` on mac

kgroat:
@jq-87 it looks like this is already in the 3.0 canary branch, but it's not been merged into 2.x

@chabou are there plans to get this back-ported to 2.x?

PR #3383, opened 2019-01-03
Fixed file paste

ulentini:
This commit fixes zeit/hyper#3340.
Since `xterm.js` changed its interface, the `send()` method has been replaced by `handler()` method.
It also fixes macOS clipboard handling for files, discarding the `file-url` that could contain url-encoded string that wouldn’t work as path.


chabou:
Thank you for your help 🙏

PR #3388, opened 2019-01-04
Emoji Fix

juancampa:
## Problem
cmd+ctrl+space didn't show the emoji dialog

## Solution
Stop intercepting this keymapping and let the browser handle it. This makes sense now that we're always using a `textarea` as our source of input

**NOTE:** The rest of the changes to make emoji work are on this PR: https://github.com/zeit/xterm.js/pull/6

### DIFF
--- a/app/keymaps/darwin.json
+++ b/app/keymaps/darwin.json
@@ -45,7 +45,6 @@
   "editor:deleteBeginningLine": "command+backspace",
   "editor:deleteEndLine": "command+delete",
   "editor:clearBuffer": "command+k",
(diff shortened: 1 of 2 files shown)


app/keymaps/darwin.json
```diff
@@ -45,7 +45,6 @@
   "editor:deleteBeginningLine": "command+backspace",
   "editor:deleteEndLine": "command+delete",
   "editor:clearBuffer": "command+k",
-  "editor:emojis": "command+ctrl+space",
```
chabou: Very nice catch.

PR #3405, opened 2019-01-09
Change directory for local plugins in comments

wanderingstan:
https://hyper.is/#extensions-module-loading has new path for local plugins

image

> When developing, you can add your plugin to ~/Library/Application Support/Hyper/.hyper_plugins/local and then specify it under the localPlugins array in .hyper.js.
> ...
> Note: plugins at ~/.hyper_plugins still supported, but will be ignored, if plugins in application directory present. Otherwise they will be moved to the application directory at first run.

_Side note: I'm still not able to get local plugins to load, even using this updated directory. :(_


chabou:
`~/.hyper_plugins/local` location works great for me (on macOS).
If you want to develop a plugin I would suggest you to run Hyper in dev mode.

You can find some explanations here: https://github.com/zeit/hyper/blob/canary/PLUGINS.md

ppot:
@wanderingstan 2 Pointer. The docs is valid for canary. Hyper latest still use `~/.hyper_plugins/`

Stanzilla:
I think this is obsolete now.

PR #3406, opened 2019-01-10
Add macOptionSelectionMode option

chabou:
Two possible value: 
* `'vertical'` (Default): enable column selection when Option key is hold
* `'force'`: force selection regardless of whether the terminal is in mouse events mode


juancampa:
It would be cool if this was available on all platforms and not just mac, but I'm seeing that xterm.js implemented it as a mac-only thing. Oh well.

Tyriar:
@juancampa I think shift or something forces on Linux, it's a mac only option as it's a conflict only on mac.

juancampa:
@Tyriar, thanks for clarifying! :raised_hands:

PR #3407, opened 2019-01-10
Fix webgl config

onecamp:
Fixes:

 1. The `webGLRenderer` flag was in the wrong place in `config-default.js`. Not letting new users change the renderer.
 1. The check to see if the user config has a value for `webGLRenderer` was incorrect. Defaulting to the canvas renderer.

### DIFF
--- a/app/config/config-default.js
+++ b/app/config/config-default.js
@@ -129,7 +129,7 @@ module.exports = {
     // bellSoundURL: 'http://example.com/bell.mp3',
 
     // Whether to use the WebGL renderer. Set it to false to use canvas-based
-    // rendering (slower)
+    // rendering (slower, but supports transparent backgrounds)
     webGLRenderer: true,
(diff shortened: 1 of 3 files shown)


chabou:
@rauchg Do you want an automatic fallback to canvas if the user is using a non-opaque background color?

With this PR, `WebGLRenderer` will be activated by default and will break transparency.

app/config/config-default.js
```diff
@@ -128,6 +128,10 @@ module.exports = {
     // URL to custom bell
     // bellSoundURL: 'http://example.com/bell.mp3',
 
+    // Whether to use the WebGL renderer. Set it to false to use canvas-based
+    // rendering (slower)
```
j-f1:
```suggestion
    // rendering (slower, but supports transparent backgrounds)
```

PR #3409, opened 2019-01-10
 Fix xterm.js resource leaks in split pane

### LINKED ISSUE #3281: Memory leak

- **OS version and name**: Mac OS X 10.14.1
- **Hyper.app version**: 2.0.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/EddiG/f6198a586c460f73f5c78207d0624859
- **The issue is reproducible in vanilla Hyper.app**: `Is Vanilla`

#### How to reproduce
Open 10 tabs and close them.

As you can see on the [video](https://monosnap.com/file/ZSpiVbBwbTd36pQByAgzTxc3BjXnGg) the memory consumption grows really fast and even worse it stays held after closing tabs.
I'm not experienced in Electron but if someone advice how to localize the problem I will be glad to provide additional information.

### LINKED ISSUE #3351: Re-opening #3318 - Cursor Erratically Blinks After Splitting Pane and Switching Focus

- **OS version and name**: macOS Sierra Version 10.12.6
- **Hyper.app version**: Hyper 2.1.0 (stable)
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/JeffReeves/1b21874bfb9fafa331ebbe26df67b02e 
- **The issue is reproducible in vanilla Hyper.app**: `Is Vanilla`

## Issue
If you split the pane (horizontally or vertically), put any text within the first pane, then shift focus to another window (from another program) and back, the cursor blink becomes erratic. See #3318 . 

It gets worse each time you do this. It almost feels like whatever event handler there is for the cursor blink gets recreated each time, and pretty soon you have like 10+ of them trying to blink the cursor...

I'm not running ANY plugins, just vanilla Hyper.js.

I've had to switch to iTerm 2 for now until this can be resolved.

### LINKED ISSUE #3408: High CPU usage, resource leaks, and erratic cursor blinking after splitting panes

- **OS version and name**: Fedora 29
- **Hyper.app version**: 2.1.1

## Issue

Splitting panes and closing them repeatedly leads to continuous high CPU usage, a memory leak, as well as the erratic cursor blinking behaviour described in #3318 and #3351.

### Steps to reproduce

- Split the pane (horizontally or vertically).
- Close the new pane, leaving the old one in place.
- Repeat 30+ times for a noticeable effect.

A similar effect can be seen when opening and closing multiple tabs as described in #3281.

The memory leak can be confirmed via Developer Tools by going to Memory > Take Snapshot and observing multiple non-GCed instances of type Term (React component) and Terminal (xterm.js classes) even when only one pane remains open.

### Root cause

The issue is caused by unreleased event listeners in xterm.js, and in particular by calling `Terminal.open()` repeatedly on the same instance. **I will be submitting a PR shortly with a fix.**

igorsdv:
Fixes #3408, fixes #3351, fixes #3281.

Repeated calls to `Terminal.open()` on the same instance register a number of event listeners each time, which leads to resource leaks and erroneous listener duplication. It's not clear to me if multiple calls to `Terminal.open()` is a use case supported by xterm.js -- if it is, this issue should be fixed there. In the meantime, this is an ugly workaround to release all the event listeners prior to calling `Terminal.open()` again, while preserving the state we need to keep.

First, we run the equivalent of `Terminal.dispose()` to release disposables and events on the `Terminal` object. After this, some event listeners on non-disposable properties of the `Terminal` object itself remain set, and we must handle those separately. This is important because the old event listeners indirectly close over the previous `_parent` elements (now detached), which hold references to unmounted React components and old props, which in turn refer to old `Terminal` objects representing closed panes and tabs. Thus a single unreleased listener can be enough to cause a significant resource leak.

This can be seen using the Developer Tools by going to Memory > Take Snapshot. Multiple non-GCed `Term` components and `Terminal` instances are reachable even when only one pane remains open.

Additionally, the necessary cleanup on `TERM_GROUP_EXIT` in `lib/components/terms.js` is moved from `onRef()` to `componentWillReceiveProps()`. This is because `this.props.sessions` is not yet updated when `onRef()` is called, and currently the `delete` statement is never executed.

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -75,13 +75,11 @@ export default class Term extends React.PureComponent {
   constructor(props) {
     super(props);
     props.ref_(props.uid, this);
-    this.termRef = null;
     this.termWrapperRef = null;
     this.termRect = null;
     this.onOpen = this.onOpen.bind(this);
     this.onWindowResize = this.onWindowResize.bind(this);
     this.onWindowPaste = this.onWindowPaste.bind(this);
-    this.onTermRef = this.onTermRef.bind(this);
     this.onTermWrapperRef = this.onTermWrapperRef.bind(this);
     this.onMouseUp = this.onMouseUp.bind(this);
     this.termOptions = {};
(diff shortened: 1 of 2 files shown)


rauchg:
@igorsdv this is absolutely amazing. Thank you so much!

lib/components/term.js
```diff
@@ -91,6 +91,23 @@ export default class Term extends React.PureComponent {
   componentDidMount() {
     const {props} = this;
 
+    if (props.term) {
+      let _core = props.term._core;
+      // Release listeners and disposables
+      _core._disposables.forEach(d => d.dispose());
+      _core._disposables.length = 0;
+      _core._events = {};
```
Tyriar: Not all listeners are installed when open is called, doing this on multiple objects will almost certainly break other components. Can the code be changed to call `open` once? I think it works fine if you open on some parent element and transfer that between split elements.

Also here is upstream request https://github.com/xtermjs/xterm.js/issues/1323

igorsdv: @Tyriar Thanks! The only listener not registered in `open` that I noticed was InputHandler. In the end the object ends up with the same number of listeners as before the `dispose` calls. But yes I agree that calling `open` once and reattaching the container as you suggest would be better, not sure how much more complicated it is with React but it should be possible.

igorsdv:
Implemented @Tyriar's suggestion to reattach the xterm.js parent element without calling `open` again, in the end it turned out less intrusive than I anticipated. The performance gains are still there, but without messing with xterm.js internals.

chabou:
Removing `this.termRef` is a breaking change
https://github.com/zeit/hyper/pull/3409/files#diff-268710614c2274f611be667a5f72bb13L79

It has broken some plugins: https://github.com/chabou/hyper-pane/issues/42

PR #3411, opened 2019-01-12
Fix and simplify resize handling using ResizeObserver

### LINKED ISSUE #3227: Terminal prompt is beyond viewing range
I'm facing an issue where the current terminal and prompt is beyond the visual range. I am attaching a screenshot. If I type and command and press enter, the command is successfully executed, but I cannot see the cursor or the prompt in my viewing range. It is somewhere "below" on the "next page".

OS Version: Debian 9
Hyper version: Stable and Hyper 2.1.0-canary.2
Dev tools console:

```
(re)loading renderer plugins
bundle.js:1 Plugin hyper-opacity (1.0.0) loaded.
bundle.js:1 Plugin hyper-arc-dark-controls (1.2.0) loaded.
index.html:35 total init time 277.74000000000007
```
screenshot from 2018-09-29 10-28-56

igorsdv:
Fixes #3227.

Terminals do not get properly resized when switching between single-tab and multiple-tab layout, causing a rendering issue at the bottom of the viewport. This can also be triggered when splitting panes vertically in a small window.

This is fixed by using a [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) on the wrapper element to trigger the resize, which incidentally simplifies a lot of the resizing code. In particular, `onOpen` is no longer needed as the observer fires immediately when registered.

Note that ResizeObserver is an experimental API, but has been available on Chrome stable since Chrome 64.

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -1,4 +1,4 @@
-/* global Blob,URL,requestAnimationFrame */
+/* global Blob,URL,requestAnimationFrame,ResizeObserver */
 import React from 'react';
 import {Terminal} from 'xterm';
 import * as fit from 'xterm/lib/addons/fit/fit';
@@ -78,8 +78,6 @@ export default class Term extends React.PureComponent {
     this.termRef = null;
     this.termWrapperRef = null;
     this.termRect = null;
-    this.onOpen = this.onOpen.bind(this);
-    this.onWindowResize = this.onWindowResize.bind(this);
     this.onWindowPaste = this.onWindowPaste.bind(this);
     this.onTermRef = this.onTermRef.bind(this);
     this.onTermWrapperRef = this.onTermWrapperRef.bind(this);
@@ -105,8 +103,6 @@ export default class Term extends React.PureComponent {
     if (this.props.isTermActive) {
       this.term.focus();
     }
-
-    this.onOpen(this.termOptions);
 
     if (props.onTitle) {
       this.disposableListeners.push(this.term.addDisposableListener('title', props.onTitle));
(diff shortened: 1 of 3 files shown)


chabou:
I have a minor regression.

Terms are blinking when resize terms
hyper-resize

lib/components/term.js
```diff
@@ -299,6 +272,13 @@ export default class Term extends React.PureComponent {
 
   onTermWrapperRef(component) {
     this.termWrapperRef = component;
+
+    if (component) {
+      this.resizeObserver = new ResizeObserver(() => this.fitResize());
```
juancampa:
```suggestion
      this.resizeObserver = new ResizeObserver(() => {
        if (this.resizeTimeout) {
          return;
        }
        this.resizeTimeout = setTimeout(() => {
          delete this.resizeTimeout;
          this.fitResize()
        }, 0);
      });
```

@chabou found an issue with the panes blinking in and out.

I'm not sure why it happens but it seems due to `fitResize` being called at the wrong part of the event loop. I tried wrapping it with `setTimeout(() => ..., 0)` and that seems to fix it (`requestAnimationFrame` did not)

hyper-resize

onecamp:
Hey @igorsdv, fyi I'm pushing the couple of changes directly to this PR, because we're trying to get it merged as soon as possible (It helps a lot :)

lib/components/terms.js
```diff
@@ -144,11 +144,34 @@ export default class Terms extends React.Component {
             left: 0;
             bottom: 0;
             color: #fff;
-            transition: ${isMac ? 'none' : 'margin-top 0.3s ease'};
           }
 
           .terms_termsShifted {
             margin-top: 68px;
+            animation: shift-down 0.2s ease-out;
```
onecamp: I made the shift-down a little faster (and eased out) because it overlaps a bit with the tabs

PR #3416, opened 2019-01-16
Style invisible term groups with display: none

onecamp:
This solves two problems:

 1. Makes resizing the window way more responsive (~8X) when multiple tabs are open (e.g. maximize a small window with 10+ tabs)
 1. When using WebGL:
    1. maximize a small window
    1. switch to another tab
    1. result: the old contents of the canvas are flashed for a frame or two, scaled and distorted.

I'm not entirely sure why this improves (1). I think it's a combination of:
 - The browser doesn't need to run layout for the invisible term-groups
 - From the profiles, I can see that rendering is faster but I'm not sure why (see screenshots)

Before (canvas renderer):
before
After (canvas renderer):
after

## Testing

 - [x] The `IntersectionObserver` still pauses rendering correctly
 - [x] Canvas renderer
 - [x] WebGL renderer

### DIFF
--- a/lib/components/terms.js
+++ b/lib/components/terms.js
@@ -152,7 +152,7 @@ export default class Terms extends React.Component {
           }
 
           .terms_termGroup {
-            display: block;
+            display: none;
             width: 100%;
             height: 100%;
             position: absolute;
@@ -161,6 +161,7 @@ export default class Terms extends React.Component {
           }
 
           .terms_termGroupActive {
+            display: block;
             left: 0;
           }
         `}</style>


juancampa:
Maybe @Tyriar has an insight on what's happening with the renderer.

chabou:
I think that termgroup was not hidden for `fit` xterm addon. If a termgroup is not displayed, term container width/height was unknown and made fit resizing buggy.

But it could have changed since we try it, months ago.

chabou:
Better with non opaque background:
hyper-resize2

juancampa:
> I think we need to force-resize all term of a tab when it becomes active

Makes sense, I'll try this

Tyriar:
xterm.js turns off the canvas/webgl renderers when the terminal is not visible:

https://github.com/xtermjs/xterm.js/blob/509ce5fa3a698ee7847419117e9dd6b979b105bf/src/renderer/Renderer.ts#L70-L89

If you're just changing opacity of "backgrounded" terminals or something like that, the intersection observer won't fire.

juancampa:
Seems like this was related to the electron v4 issue. So I might end up closing this in favor of moving back to electron v3

PR #3420, opened 2019-01-17
Downgrade electron to v3

juancampa:
### Problem
Electron v4 has a big performance issue with 2d canvases Tracked here:

 https://github.com/electron/electron/issues/16439

### Solution

Revert back to Electron v3 for now.

### Testing

  - [x] No errors on output
  - [x] Plugins work (tested installing and using hypercwd)
  - [x] Multiple tabs
  - [x] Multiple windows
  - [x] Splits
  - [x] Menus
  - [ ] Auto update (pop up shows)
  - [x] Clean shutdown
  - [x] Vim session with splits
  - [x] Clipboard works
     - [x] Through menu
     - [x] Hotkeys
  - [x] Performance looks fine
     - [x] canvas
     - [x] webgl
----
  - [x] Linux
  - [x]  MacOS
  - [x]  Windows

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ language: node_js
 matrix:
   include:
     - os: linux
-      node_js: 10.11.0
+      node_js: 10.2.0
       env: CC=clang CXX=clang++ npm_config_clang=1
       compiler: clang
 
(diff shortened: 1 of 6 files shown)


juancampa:
> Can you commit your `yarn.lock`?

Thanks, I guess I also missed the last yarn.lock. Not sure why I keep doing this :sweat_smile:

PR #3435, opened 2019-01-23
Verify that webgl2 is supported before using it

### LINKED ISSUE #3433: WebGL2 not supported
- [x] I am on the latest release (canary)

- **OS version and name**: Ubuntu 16.04
- **Hyper.app version**: Hyper 3.0.0-canary.5
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: no need, reproducible in vanilla
- **Relevant information from devtools** _(CMD+ALT+I on macOS, CTRL+SHIFT+I elsewhere)_: screenshot below

## Issue

Hi, unfortunately the latest canary version makes hyper unusable.. I think the screenshots describe the issue...

There is no cursor nor input on the terminal, I tried Ctrl+D and it did nothing:

image

Ctrl+Shift+I worked, it shows this:

image

The last error about dimensions might have something to do with my window manager that always makes Hyper start at maximized-size but not really maximized.

This started happening when I installed canary.5. Let me know if you need more info!

onecamp:
Some hardware/drivers seem to not support WebGL2 (now our default renderer) which completely breaks hyper.

Check that webgl2 is supported before using.

This seems to add ~10ms to my startup time, which I think it's acceptable. The alternative being catching an exception in `Terminal.open` and *then* falling back to canvas.

fixes #3433

### DIFF
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -20,11 +20,27 @@ const CURSOR_STYLES = {
   BLOCK: 'block'
 };
 
+const isWebgl2Supported = (() => {
+  let isSupported;
+  return () => {
+    if (isSupported === undefined) {
+      const canvas = document.createElement('canvas');
+      const gl = canvas.getContext('webgl2', {depth: false, antialias: false});
+      isSupported = gl instanceof window.WebGL2RenderingContext;
+      if (!isSupported) {
+        // eslint-disable-next-line no-console
+        console.warn(`WebGL2 is not supported on your machine. Falling back to canvas-based rendering.`);
+      }
+    }
+    return isSupported;
+  };
+})();
+
 const getTermOptions = props => {
   // Set a background color only if it is opaque
   const needTransparency = Color(props.backgroundColor).alpha() < 1;
   const backgroundColor = needTransparency ? 'transparent' : props.backgroundColor;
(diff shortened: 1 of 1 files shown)


lib/components/term.js
```diff
@@ -20,11 +20,27 @@ const CURSOR_STYLES = {
...
+const isWebgl2Supported = (() => {
+  let isSupported = window.WebGL2RenderingContext ? undefined : false;
+  return () => {
+    if (isSupported === undefined) {
+      const canvas = document.createElement('canvas');
+      const gl = canvas.getContext('webgl2', {depth: false, antialias: false});
+      isSupported = gl instanceof window.WebGL2RenderingContext;
+      if (!isSupported) {
+        // eslint-disable-next-line no-console
+        console.warn(`WebGL2 is not supported on your machine. Falling back to canvas-based rendering.`);
```
chabou: I would move this warning in `getTermOptions` where fallback occurs. This is juste a test here.

We can warn if webGL is disabled due to a non-opaque backgroundColor too.

onecamp: Done!

onecamp: Sorry, linting issues on push. One sec

onecamp: Ok, *now* it's done. I need to set up eslint integration in vim.

PR #3437, opened 2019-01-24
Restore termRef

igorsdv:
Fixes breaking change introduced in #3409


rauchg:
Is it possible to add a test to prevent regressions like this one?

igorsdv:
@chabou Not sure why that happened, should be fixed now.

PR #3441, opened 2019-01-24
Display the renderer type in the About dialog

onecamp:
Since the actual renderer used is determined not only by the config but also by the bg opacity and general support. We want to display the renderer in an easy to access place for diagnostics.

### DIFF
--- a/app/menus/menu.js
+++ b/app/menus/menu.js
@@ -13,6 +13,7 @@ const helpMenu = require('./menus/help');
 const darwinMenu = require('./menus/darwin');
 const {getDecoratedKeymaps} = require('../plugins');
 const {execCommand} = require('../commands');
+const {getRendererTypes} = require('../utils/renderer-utils');
 
 const appName = app.getName();
 const appVersion = app.getVersion();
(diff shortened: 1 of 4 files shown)


onecamp:
cc @rauchg

chabou:
I am not fan of this solution.

About box is a global app window, but our renderer is Term specific.
Some plugin like [hyper-autoprofile](https://github.com/chabou/hyper-autoprofile) can change background color of each term (then the renderer).

onecamp:
> can change background color of each term

Hmm interesting, I didn't know this was possible. Maybe we can show this instead:

`WebGL renderer support: yes|no|disabled due to translucent background`.

OTOH, the number of users with both renderers is probably low enough that it won't matter. This should be used as a diagnostics tool more than anything else.

onecamp:
@chabou let me know what you think. IMO, this version should get the job done for diagnostics:
. But I'd gladly improve it if you think it's worth it

app/menus/menu.js
```diff
@@ -39,10 +40,18 @@ exports.createMenu = (createWindow, getLoadedPluginVersions) => {
     const pluginList =
       loadedPlugins.length === 0 ? 'none' : loadedPlugins.map(plugin => `\n  ${plugin.name} (${plugin.version})`);
 
+    const rendererCounts = Object.entries(getRendererTypes()).reduce((acc, [, type]) => {
```
chabou:
```suggestion
    const rendererCounts = Object.values(getRendererTypes()).reduce((acc, type) => {
```

PR #3451, opened 2019-01-30
Hyper v3 icons

onecamp:


juancampa:
Merging so that we can include this in the next canary with the electron bump. Though it might now show due to CI replacing it with the "canary" icon

Koolstr:
The new icon is not showing in Windows. Or is this new icon for macOS only? It is nicer, so it would be great if we can get it on Windows too.

Alternatively, it might just not be reflecting for me since I am on the Canary/beta build. If this is the cause, any chance we can get the Canary icon adapted into this new one as well?

PR #3457, opened 2019-02-04
Fix electron window.open vulnerability

chabou:
Apply suggested workaround: https://electronjs.org/blog/window-open-fix

### DIFF
--- a/app/ui/window.js
+++ b/app/ui/window.js
@@ -226,6 +226,9 @@ module.exports = class Window {
       }
     });
 
+    // mitigate a security issue: https://electronjs.org/blog/window-open-fix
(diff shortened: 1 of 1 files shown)


app/ui/window.js
```diff
@@ -226,6 +226,9 @@ module.exports = class Window {
       }
     });
 
+    // mitigate a security issue: https://electronjs.org/blog/window-open-fix
```
juancampa:
```suggestion
    // mitigate a security issue: https://electronjs.org/blog/window-open-fix
    // TODO: remove when merged into Hyper 3 (which already has the fix via a new version of electron)
```

PR #3458, opened 2019-02-04
Bumping electron to 3.1.3

juancampa:
This version includes a vulnerability fix but also broke our copy mechanism due to [this change](https://github.com/electron/electron/pull/15892). Adding `registerAccelerator: true` seems to fix it on Linux. 

 - [x] Linux
 - [ ] MacOS


juancampa:
MacOS seems to also work. Can't test on Windows right now but it *should* be fine

Issue #3446, opened 2019-01-27
RFC: Session API


chabou:
# Goal
`Session` and `pty` decoupled. 
Some plugins need more control about Session instances in main process. It would be required, for example, if a plugin want to launch the `pty` in another host (like #2978) or intercept pty stream to duplicate/record/stream/log it.
Interception can already be done with a middleware in renderer process, but having a Session without a `pty` can't.

# Proposal
I would suggest to add a new API entry: `decorateSessionClass(Session)`. A plugin can expose this function that would return a new class that inherit from the (kind of) Session class received in parameter.

If multiple plugins declare this hook, we would finally get a class that have been multiple times inherited. It can lead to some conflicts but this is already the case for other decorate functions.

ppot:
Decoupled architecture: Session into single process. You can attach them from their sessionId which are registered from hyper-core. Enabling background job profiling etc. Event-hook for session notification.  Sometimes you wan't to close the pane but not the shell process inside it.

chabou:
I think that is out of my context. I was only thinking about offer possibility to plugin to modify Session behavior. Sometimes you don't want that a session create a pty and for now, there isn't any way to do this.

With this RFC, you'll be able to achieve your single process isolation with a plugin 😆

ppot:
Yeah. That's what i'm saying as attaching them :). More versatility by having them not strictly bound to the view as a pty-session as of ATM.

j-f1:
Can this be closed?

chabou:
Yes! Thanks!

MisterTea:
@chabou  Is there a PR to track the progress of this RFC?

PR #3464, opened 2019-02-06
Add Session class and Session options decoration by plugins

chabou:
This PR implements [this RFC](https://github.com/zeit/hyper/issues/3446)

It adds 2 new plugin hook to our API `decorateSessionClass` and `decorateSessionOptions`.

It unlocks some awesome new plugin possiblities. I made a plugin as an example (and a proof of concept): https://github.com/chabou/hyper-session-log

### DIFF
--- a/app/plugins.js
+++ b/app/plugins.js
@@ -417,12 +417,12 @@ exports.getDecoratedBrowserOptions = defaults => {
   return decorateObject(defaults, 'decorateBrowserOptions');
 };
 
-exports.decorateSessionClass = Session => {
-  return decorateClass(Session, 'decorateSessionClass');
-};
-
 exports.decorateSessionOptions = defaults => {
   return decorateObject(defaults, 'decorateSessionOptions');
+};
+
+exports.decorateSessionClass = Session => {
+  return decorateClass(Session, 'decorateSessionClass');
 };
 
 exports._toDependencies = toDependencies;
--- a/app/plugins.js
+++ b/app/plugins.js
@@ -332,9 +332,9 @@ exports.onWindow = win => {
   });
 };
 
-// decorates the base object by calling plugin[key]
+// decorates the base entity by calling plugin[key]
 // for all the available plugins
-function decorateObject(base, key) {
+function decorateEntity(base, key, type) {
   let decorated = base;
   modules.forEach(plugin => {
     if (plugin[key]) {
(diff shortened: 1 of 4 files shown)


app/plugins.js
```diff
@@ -409,4 +417,12 @@ exports.getDecoratedBrowserOptions = defaults => {
   return decorateObject(defaults, 'decorateBrowserOptions');
 };
 
+exports.decorateSessionClass = Session => {
+  return decorateClass(Session, 'decorateSessionClass');
+};
+
+exports.decorateSessionOptions = defaults => {
+  return decorateObject(defaults, 'decorateSessionOptions');
```
ppot: non breaking change

`exports.decorateSessionOptions` should come before `exports.decorateSessionClass ` to follow code structure of  `decorateObject ` and `decorateClass `

app/ui/window.js
```diff
@@ -101,8 +102,9 @@ module.exports = class Window {
         extraOptions,
         {uid}
       );
-
-      const session = new Session(options);
+      const options = decorateSessionOptions(defaultOptions);
```
juancampa: This is cool, more extensibility! :tada:

app/session.js
```diff
@@ -141,12 +141,12 @@ module.exports = class Session extends EventEmitter {
   }
 
   write(data) {
-    this.pty.write(data);
+    this.pty && this.pty.write(data);
```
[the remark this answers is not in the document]
chabou: I anticipated the case: A plugin overrides the constructor and prevents the pty creation (without override this `write` method). It seems legit to fail silently.

But thanks to your comment, I think this is currently pretty hard to prevent this pty creation because it occurs in constructor. Such a plugin can't call `super()` and would break other plugins that have previously decorated this Class.

I think we should isolate pty creation in a `init(options)` function that plugin would override to prevent this pty creation without breaking constructor chain.

app/plugins.js
chabou: You're right, I'll fix this

juancampa:
Thanks for addressing my feedback @chabou! Also, thanks for being patience with my response delay, I'm in the process of moving to a new place and these days have been pretty hectic :sweat_smile: 

I think it's reasonable to expect that a plugin that overrides `init` to prevent `pty` creation should also override other methods that use the pty (such as `write`). It might be better to have the base class be consistent and allow any potential inconsistencies to be handled by plugins that introduce them.

This approach could cause problems if, for example, there's a plugin that overrides `init` to do something else but still wants the `pty` to be created. It won't see any errors when `write` is called

juancampa:
In short: if a plugin wants to remove the `pty`, it should be its responsibility to also override the other methods.

chabou:
Look at `destroy()` method.
I don't want to force plugin to override it only to prevent an error with `pty.destroy()`.

My point is that plugin should focus on what it adds and not what it removes. Having base class that have conditional behavior, anticipating that some part could me remove is a good thing.

But I totally understand your point that missing pty could be an unwanted error and silent protection could be harmful.

I will add some warning.

PR #3470, opened 2019-02-10
Expose active window positon on move to reducer

ppot:
Expose window position `{x, y}` once the active window when it's moved.

In order to add modularity for plugins: 
  - The `onWindow` only expose the window creation. But once the window is created, moved action are not accessible. 
- Allow plugins to know the current position of the active window based on the `UI_WINDOW_MOVE` from the window move action lunched by the `RPC`.
- Enable better state recording for plugins such as session restoration

### DIFF
--- a/app/ui/window.js
+++ b/app/ui/window.js
@@ -215,7 +215,7 @@ module.exports = class Window {
     for (const ev of ['maximize', 'unmaximize', 'minimize', 'restore']) {
       window.on(ev, () => rpc.emit('windowGeometry change'));
     }
-    rpc.window.on('move', () => {
+    window.on('move', () => {
       const position = window.getPosition();
       rpc.emit('move', {bounds: {x: position[0], y: position[1]}});
     });
--- a/app/ui/window.js
+++ b/app/ui/window.js
@@ -216,7 +216,8 @@ module.exports = class Window {
       window.on(ev, () => rpc.emit('windowGeometry change'));
     }
     rpc.win.on('move', () => {
-      rpc.emit('move');
+      const position = window.getPosition();
+      rpc.emit('move', {position: {x: position[0], y: position[1]}});
     });
     rpc.on('close', () => {
       window.close();
--- a/app/ui/window.js
+++ b/app/ui/window.js
@@ -215,9 +215,9 @@ module.exports = class Window {
     for (const ev of ['maximize', 'unmaximize', 'minimize', 'restore']) {
       window.on(ev, () => rpc.emit('windowGeometry change'));
     }
-    rpc.win.on('move', () => {
+    rpc.window.on('move', () => {
       const position = window.getPosition();
-      rpc.emit('move', {position: {x: position[0], y: position[1]}});
+      rpc.emit('move', {bounds: {x: position[0], y: position[1]}});
     });
     rpc.on('close', () => {
       window.close();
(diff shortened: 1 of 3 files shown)


app/ui/window.js
```diff
@@ -216,7 +216,8 @@ module.exports = class Window {
       window.on(ev, () => rpc.emit('windowGeometry change'));
     }
     rpc.win.on('move', () => {
-      rpc.emit('move');
```
juancampa: Do you mind replacing `rpc.win` with `window`? I know you didn't write that part but It's a bit confusing the way it's currently written

ppot: Yeah sure!

app/ui/window.js
```diff
@@ -216,7 +216,8 @@ module.exports = class Window {
...
+      const position = window.getPosition();
+      rpc.emit('move', {position: {x: position[0], y: position[1]}});
```
chabou: This is a common usage to not separate position and size.

I would trigger:
```
const position = window.getPosition();
rpc.emit('move', {bounds: {x: position[0], y: position[1]}});
```

Or directly:
```
const bounds = window.getBounds();
rpc.emit('move', {bounds});
```

Should we do the same when window size has changed?

ppot: The reason i wrote `{x: position[0], y: position[1]}}` is because
```
const bounds = window.getBounds();
rpc.emit('move', {bounds});
```
render as  ex: `{0 : 24, 1: 52}` note that those are just random number as example.

chabou: No, `window.getPosition()` returns an array (and you were right to decompose it), but `window.getBounds()` returns `{ x: 192, y: 147, width: 624, height: 469 }` for example.

chabou: I didn't comment on your decomposition. Only on your variable name and your variable content ;)

ppot: Yeah, indeed you are right. However, for the window move action doesn't invoke a window resize. So, no need to send unnecessary data to the reducer 😄

ppot: Should be good to go 🚀

ppot: Fixed: Updated win for window on `rpc.js`

app/rpc.js
```diff
@@ -5,7 +5,7 @@ const uuid = require('uuid');
 class Server extends EventEmitter {
   constructor(win) {
     super();
-    this.win = win;
+    this.window = win;
```
chabou: Please do not rename this. This is a potential breaking change.

app/ui/window.js
```diff
@@ -215,8 +215,9 @@ module.exports = class Window {
     for (const ev of ['maximize', 'unmaximize', 'minimize', 'restore']) {
       window.on(ev, () => rpc.emit('windowGeometry change'));
     }
-    rpc.win.on('move', () => {
-      rpc.emit('move');
+    rpc.window.on('move', () => {
```
chabou:
```suggestion
    window.on('move', () => {
```

I think that @juancampa suggested to use `window` variable that is directly accessible instead of `rpc.win.on()`

ppot: Yeah. Just updated for window.

juancampa:
@chabou does this LGTM now?

chabou:
Thank you for your help @ppot

PR #3485, opened 2019-02-17
Add Window class and options plugin decoration

ppot:
- Since the current: `onWindow` is the BrowserWindow and `decorateBrowserOptions` decorate it.
- Since the current  `onWindow` doesn't refer to the `Window` object https://github.com/zeit/hyper/blob/canary/app/ui/window.js#L19
- Since #3352 Make session creation optimistic. 

The current implementation make the session creation happen before the instance of the window. There for, if the plugin want to bind a session to a window. The new `decorateSessionClass` and `decorateSessionOptions` will result on the session not being accessible to the window's instance of the plugin. And therefor, not being able to bind it to an `UID`.

This PR include more extensibility by allowing the developer to either access `onWindowClass` the window's uid, or decorate it using `decorateWindowClass`.

This enable the possibility to give window's profile ID as unique name
```
exports.decorateWindowClass = defaults => {
  defaults = {uid: 'profileId_dev'};
  return defaults;
}
```
or acces the default `UID` created: 
```
{ uid: 'b545deb5-c1e4-4ecd-af3b-e722126f1d40' }
```

Target: https://github.com/ppot/hyper-automator/issues/4

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -19,7 +19,6 @@ const checkSquirrel = () => {
     squirrel = require('electron-squirrel-startup');
     //eslint-disable-next-line no-empty
   } catch (err) {}
-
   if (squirrel) {
     // eslint-disable-next-line unicorn/no-process-exit
     process.exit();
--- a/app/plugins.js
+++ b/app/plugins.js
@@ -318,6 +318,20 @@ exports.onApp = app_ => {
   });
 };
 
+exports.onWindowClass = win => {
+  modules.forEach(plugin => {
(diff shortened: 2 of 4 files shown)


ppot:
@juancampa Review ;)

juancampa:
Can we document this on https://github.com/zeit/hyper-site? @ppot at least open an issue with a short description :raised_hands:

ppot:
ya, will do tomorrow @juancampa ;)

ppot:
PR made on the hyper-website ;)

Issue #51, opened 2016-07-05
Non-retina text rendering is bad


zpao:
There's a rule, presumably down in Electron, that is setting `-webkit-font-smoothing: antialiased` for `x-screen`. Perhaps I'm extra sensitive to this, but this looks absolutely terrible on non-retina screens. When I toggle that rule in the inspector everything looks great. Personally I think that also looks better on a retina screen. It might be worth considering enforcing subpixel-antialiasing (IMO always, but could be on a media query for < 2 pixel density), or eventually some setting even.

rauchg:
@zpao +1000. I've been seeing some screenshots from non-retina that have wanted to make me cry.

johanbrook:
I actually think antialiasing looks bad on Retina too. Compare screens below.

![skarmavbild 2016-07-06 kl 01 57 00](https://cloud.githubusercontent.com/assets/437261/16603532/0436f802-431d-11e6-8dbf-f03002ea90b3.png)

![skarmavbild 2016-07-06 kl 01 57 07](https://cloud.githubusercontent.com/assets/437261/16603536/0af7bc08-431d-11e6-8c7b-ed293cc1fad9.png)

The non antialiased one has clearer colours and more legible typography imo (might only be me). The antialiased text looks a bit weak: the black from the background sort of bleeds into the glyphs. Compare with iTerm:

<img width="786" alt="skarmavbild 2016-07-06 kl 02 00 00" src="https://cloud.githubusercontent.com/assets/437261/16603588/78c5aaf6-431d-11e6-92c6-86b27538dd27.png">

zpao:
@johanbrook I agree. I've lost the "just use subpixel-antialiasing all the time" argument too many times though so I hedged my bets 😛

zpao:
In case anybody else comes across this and it's not "fixed" the way you want, you can use `~/.hyperterm.js` to override.

```js
    termCSS: `
      x-screen {
        -webkit-font-smoothing: subpixel-antialiased !important;
      }
    `,
```

ali:
Hey all. I noticed this same issue at work today after customizing the fonts in my hyperterm config. My fonts looked great on my laptop's retina display, but looked kinda 💩 on my external monitor. I played around in the devtools and can confirm the above fix worked out for me.

Anyway, here's a hyperterm theme I made that adds subpixel-antialiased font smoothing: https://github.com/ali/hyperterm-subpixel-antialiased

leo:
[...]

PR #3490, opened 2019-02-20
Minor fix on string #51

kant:
This pull request just goes for this `readme.md` file, so, there is no break on code.


ppot:
Thank you. Merged.

Issue #3498, opened 2019-02-26
zoom out crashes


cwlowder:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Microsoft Windows 10 Version 1803 (OS Build 17134.590)
- **Hyper.app version**: Hyper 2.1.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/cwlowder/5113db7f2314bd250cef010de2f95372
- **Relevant information from devtools** [Violation] 'requestAnimationFrame' handler took 70ms
- **The issue is reproducible in vanilla Hyper.app**: Is Vanilla

## Issue
When You zoom out by holding down `ctrl+minus` for a few seconds, hyper will freeze and then crash with the following devtool output:
```
[Violation] 'requestAnimationFrame' handler took 70ms
```
You will need to change your output scope to verbose to see it. This issue occurs with my native command line aswell if I make the shell argument empty in .hyper.js

chabou:
Reproduced. But without holding my keys.
Hyper hangs up with a font-size below 2px. (`hyperpower` is slower below 4px)

I think that xterm doesn't like to have a too small font-size.

We can certainly fix it with a 5px minimum font-size

Kacppian:
Can I pick this up?

juancampa:
@Kacppian please do! :raised_hands:

Kacppian:
Submitted the PR

PR #3509, opened 2019-03-05
Fix hanging when zoom out goes below 5px (#3498)

Kacppian:
When the font size goes below 5px, xterm starts acting up and hangs. To avoid this,
the font-size is set to 5px minimum.

It's ready to be merged.

### DIFF
--- a/lib/actions/ui.js
+++ b/lib/actions/ui.js
@@ -70,7 +70,8 @@ export function decreaseFontSize() {
       effect() {
         const state = getState();
         const old = state.ui.fontSizeOverride || state.ui.fontSize;
-        const value = old - 1;
+        // when the font-size is really small, below 5px, xterm starts showing up issues.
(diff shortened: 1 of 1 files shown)


lib/actions/ui.js
```diff
@@ -70,7 +70,8 @@ export function decreaseFontSize() {
       effect() {
         const state = getState();
         const old = state.ui.fontSizeOverride || state.ui.fontSize;
-        const value = old - 1;
+        // when the font-size is really small, below 5px, xterm starts showing up issues.
+        const value = old > 5 ? old - 1 : old;
```
chabou:
```suggestion
        const value = Math.max(old - 1, 5);
```
I prefer to use `max()` to floor a value, but this is a personal flavor.
Your way is totally acceptable.

chabou:
Thank you so much for your help @Kacppian 🙏

Issue #3378, opened 2019-01-02
Fix lastest win home & end bindings


andrzejnovak:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **WIN, 18298**: 
- **3.0.1** 
- **Is Vanilla**: 

## Issue

Key bindings for beginning and end of line no longer work as defaults.
https://github.com/zeit/hyper/blob/master/app/keymaps/win32.json#L44-45

Can be fixed by as:
`  "editor:moveBeginningLine": "Home",`
`  "editor:moveEndLine": "End",`

benc-uk:
Why was this changed?!
Glad I found this, it was driving me mad

onecamp:
Not sure I understand (I'm on Linux so I can't test it). The defaults you linked are broken? and then you fix them by adding those to your `~/.hyper.js`?

Also, what version are you on? There's no `3.0.1`, you mean `3.0.0-canary.*` or `2.0.1`?

Nachtmerrie:
Could we please fix this simple typo? It is driving me crazy :-)

I have tested and can confirm that this does *not* work:
```
"editor:moveBeginningLine": "home",
"editor:moveEndLine": "end",
```
and this *does* work:
```
"editor:moveBeginningLine": "Home",
"editor:moveEndLine": "End",
```

I am running `3.0.0-canary.8`

PR #3517, opened 2019-03-09
fix home & end keymaps on windows

Nachtmerrie:
Fixes https://github.com/zeit/hyper/issues/3378


ppot:
I would ask you to change those of linux to while you are there, probably have the same problem

Nachtmerrie:
Just installed 3.0.0-canary.8 on my ubuntu 18.10 and it works without issues (home and end). The change seems not needed for `linux.json`

PR #3521, opened 2019-03-13
Fix handling ssh link with empty user

frux:
SSH links don't work when username is not specified. For example when I open _ssh://example.com_ link in Hyper I get this:
Hyper 2019-03-13 12-20-33

According to ssh help output symbol `@` is a part of optional username param.


ppot:
Thanks

PR #3559, opened 2019-04-09
Split Vertically -> Split Right, Split Horizontally -> Split Down

cameronhimself:
Splitting vertically and splitting horizontally are, as concepts, [ambiguous and unpredictable](https://english.stackexchange.com/questions/293520/split-horizontally-or-vertically-which-one-is-which). This change replaces "Vertically"/"Horizontally" with "Right"/"Down" for clarity and usability.

Note that this only changes the menu labels and keybind names. Deeper in the code, the vertical/horizontal nomenclature is still used. Changing it was beyond me at my current level of familiarity with the code, but if it's deemed critical I can invest the time.


SidneyNemzer:
👍 "Horizontal"/"Vertical" has always made me pause to consider which one I want. I think "Right"/"Down" will be much clearer.

chabou:
I agree, I always been confused too.
@Stanzilla @juancampa what do you think?

Stanzilla:
I like it

chabou:
I think that no plugin uses `pane:splitHorizontal`/`pane:splitVertical` commands. [hyper-pane](https://github.com/chabou/hyper-pane) won't be impacted for example.

But I think this is safe to keep horizontal/vertical for underlying mechanism. So this PR is just perfect, I think. Thank you @cameronhimself 🙏

PR #3578, opened 2019-05-04
Syntax issue on paragraph #44

kant:
This PR is a mere correction in the text. So, there is no break on code, so, no collateral damage on the horizon.


pcnate:
FYI, using a pound sign and then a number references an issue/PR which may be a distraction if it references the wrong thing. You can link directly to a line in a file

https://github.com/zeit/hyper/blob/15687a24caffe335ed0d174b7ddbf68845388255/PLUGINS.md#L44

Issue #3500, opened 2019-02-26
Cannot install plugins using `hyper i {plugin}`


mikemcbride:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: macOS Mojave 10.14.3
- **Hyper.app version**: 3.0.0-canary.8
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/mikemcbride/8b540c761670b3b2f5c6020deb09c60c
- **Relevant information from devtools**: No abnormal logs in devtools
- **The issue is reproducible in vanilla Hyper.app**: Yes

## Issue

Using `hyper i [plugin_name]`, I am getting the same error every time:

```
~ » hyper i hypercwd
/Applications/Hyper.app/Contents/Resources/bin/cli.js:19153
      getProperties = memoize(() => getParsedFile().program.body[0].expression.right.properties),
                                                                                    ^

TypeError: Cannot read property 'properties' of undefined
    at memoize (/Applications/Hyper.app/Contents/Resources/bin/cli.js:19153:85)
    at d (/Applications/Hyper.app/Contents/Resources/bin/cli.js:19138:22)
    at memoize (/Applications/Hyper.app/Contents/Resources/bin/cli.js:19154:34)
    at d (/Applications/Hyper.app/Contents/Resources/bin/cli.js:19138:22)
    at Object.install (/Applications/Hyper.app/Contents/Resources/bin/cli.js:19184:37)
    at Args.args.command (/Applications/Hyper.app/Contents/Resources/bin/cli.js:9934:45)
    at Args.runCommand (/Applications/Hyper.app/Contents/Resources/bin/cli.js:11318:37)
    at Args.module.exports [as parse] (/Applications/Hyper.app/Contents/Resources/bin/cli.js:15089:10)
    at main (/Applications/Hyper.app/Contents/Resources/bin/cli.js:9991:18)
    at Object.module.exports.name (/Applications/Hyper.app/Contents/Resources/bin/cli.js:10038:1)
```

It looks like [this line](https://github.com/zeit/hyper/blob/canary/cli/api.js#L47) is the offender, but I'm not sure what's causing the error. I have not yet tried building Hyper and running locally, but I could do that and add some logging to try and determine the root cause if there's not enough information here.

[...]

PR #3610, opened 2019-05-06
Add migration path from Hyper2 config location to Hyper3

juancampa:
This PR adds a migration path from Hyper2 config file locations to Hyper3 file locations.

The idea is to use the old file by moving it to the new location. While creating back ups for any existing files.

The old file is also renamed so that this happens only on the first run.

 - [x] Test on Linux
 - [ ] Test on MacOS
 - [ ] Test on Windows

### DIFF
--- a/app/config/import.js
+++ b/app/config/import.js
@@ -1,5 +1,5 @@
 const {writeFileSync, readFileSync} = require('fs');
-const {moveSync} = require('fs-extra-p');
+const {moveSync} = require('fs-extra');
 const {sync: mkdirpSync} = require('mkdirp');
 const {defaultCfg, cfgPath, legacyCfgPath, plugs, defaultPlatformKeyPath} = require('./paths');
 const {_init, _extractDefault} = require('./init');
--- a/app/config/import.js
+++ b/app/config/import.js
@@ -1,8 +1,8 @@
-const {writeFileSync, readFileSync} = require('fs');
-const {moveSync} = require('fs-extra');
+const {moveSync, copySync, existsSync, writeFileSync, readFileSync} = require('fs-extra');
 const {sync: mkdirpSync} = require('mkdirp');
 const {defaultCfg, cfgPath, legacyCfgPath, plugs, defaultPlatformKeyPath} = require('./paths');
 const {_init, _extractDefault} = require('./init');
+const notify = require('../notify');
 
 let defaultConfig;
 
@@ -37,7 +37,31 @@ const saveAsBackup = src => {
   throw new Error('Failed to create backup for config file. Too many backups');
 };
 
+const migrate = (old, _new) => {
+  if (old === _new) {
+    return;
+  }
+  if (existsSync(old)) {
+    //eslint-disable-next-line no-console
+    console.log('Found legacy config. Migrating ', old, '->', _new);
+    if (existsSync(_new)) {
+      saveAsBackup(_new);
+    }
+    const backupName = saveAsBackup(old);
+    copySync(backupName, _new);
+    return true;
+  }
+  return false;
+};
+
 const _importConf = function() {
+  if (migrate(plugs.legacyBase, plugs.base) || migrate(legacyCfgPath, cfgPath)) {
+    notify(
+      'Hyper 3',
+      `Our settings location changed to ${cfgPath}. We've automatically migrated your existing config!`
+    );
+  }
+
   // init plugin directories if not present
   mkdirpSync(plugs.base);
   mkdirpSync(plugs.local);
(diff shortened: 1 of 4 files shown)


Stanzilla:
image

juancampa:
@Stanzilla fixed. Weird that I have that installed as a indirect dependency but you don't. Anyway that's a terrible module that I'm hoping it goes away when we update dependencies:

```
$ yarn why fs-extra-p
yarn why v1.15.2
[1/4] Why do we have the module "fs-extra-p"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "fs-extra-p@7.0.0"
info Reasons this module exists
   - "electron-builder-squirrel-windows" depends on it
   - Hoisted from "electron-builder-squirrel-windows#fs-extra-p"
   - Hoisted from "electron-builder#fs-extra-p"
   - Hoisted from "electron-builder#app-builder-lib#fs-extra-p"
   - Hoisted from "electron-builder#dmg-builder#fs-extra-p"
   - Hoisted from "electron-builder#builder-util#fs-extra-p"
   - Hoisted from "electron-builder#read-config-file#fs-extra-p"
   - Hoisted from "electron-builder#builder-util-runtime#fs-extra-p"
   - Hoisted from "electron-builder#builder-util#temp-file#fs-extra-p"
   - Hoisted from "electron-builder#app-builder-lib#electron-publish#fs-extra-p"
info Disk size without dependencies: "328KB"
info Disk size with unique dependencies: "1.46MB"
info Disk size with transitive dependencies: "2.25MB"
info Number of shared dependencies: 6
Done in 0.97s.
```

Stanzilla:
So from a technical standpoint, the migration works, but it would be nice to get some message out to the user as to what happened.

juancampa:
> Could just be a simple alert ala `Hey our settings location changed, you can now find it at <new path> instead of <old path>, we already migrated your old configuration for you!`

@Stanzilla, Added this but also rewrote it so that it handles plugin migration in the same way

Stanzilla:
@juancampa Three things here:

a) the migration did not work here, the `node_modules` folder was empty.
b) I don't think we need to migrate the plugins at all, since they will get reinstalled from `package.json` on first launch anyway
c) the only additional part we should migrate is custom made plugins that are in the `hyper_plugins/local` folder since those can't be installed via `package.json`

juancampa:
Thanks! I'll fix that. I haven't got time to thoroughly test this

ppot:
Frankly: Should not update install code. but rather do it with scripting. Such as a prompt: file found. want to move file to new destinations or not?

ppot:
We should totally rename `hyper.js` for `hyper.conf` js script will still interpret is as javascript imo.

juancampa:
> Frankly: Should not update install code. but rather do it with scripting. Such as a prompt: file found. want to move file to new destinations or not?

Thanks @ppot, I actually considered that option, but I don't see any reason for users not wantintheir files migrated. And also there are other complications that I don't think are worth handling right now: dialog would keep popping up forever until they migrate? Or just once? what if they dismiss it by accident? AFAIK we don't have a convenient way to add an arbitrary action to a popup, except for a URL link.

I definitely agree about the name being weird! I don't want to introduce more breaking changes atm though

ppot:
@juancampa Why pop-up? I was implying of a shell script. With a question in a shell pty session and response so a single flag on update or install after this. nada.

juancampa:
@ppot I see what you're saying, I like that approach because it can show any errors to the user. Main issue is that we'd need to write a portable `bash` script and a windows version. And some of my questions above are still valid (when do we stop showing it? where do we store this state?)

juancampa:
OTOH this works and we need to get this out ASAP. It's already too late unfortunately :(

Stanzilla:
@juancampa what about my point c)?

ppot:
@juancampa @Stanzilla Here why my shell script since we can export the whole `dir` on windows and on linux. We can execute one powershell script  on WSL and a bash script on OS X/ linux

Stanzilla:
Small problem with the last commit, you copy the `local` dir but also the `local.backup` dir

juancampa:
@Stanzilla, can't repro, are you testing the latest latest? I changed the order so that I first copy and then create the backup folder :thinking:

Stanzilla:
I tested https://github.com/zeit/hyper/pull/3610/commits/41e4e12524583bcad6786bc4f144d9b02cc5e81c let me try again

juancampa:
Oh I think what you're seeing is the backup of the *new* `local/` in the destination path (which is probably empty)

Stanzilla:
Yeah, it is in the destination dir but it is not empty. Fun fact, this only happens when running the installer, not when starting Hyper normally.

Otherwise seems to be good apart from one thing. If we don't copy `package.json` as well, Hyper won't install the non-local addons on first run and needs a manual `Plugins -> Update`.

juancampa:
@Stanzilla so many edge cases! Ok adding `package.json`. Thanks a lot for the support

juancampa:
Oh actually, I'm not going to migrate it. The notification says to restart hyper, that should also do it, right?

Stanzilla:
Hrm not the best of UX but should be okay.

pcnate:
Does this check whether the new file has already been modified by the user? many of us have already done this.

juancampa:
@pcnate It doesn't but it backs it up for that case

pcnate:
@juancampa good enough. looks like needs testing. is there a short howto for how to test PRs?

juancampa:
@pcnate, what I've been doing is testing different scenarios, installling from scratch with old files, without old files, etc.

@Stanzilla would you say that it works correctly on Windows?

Stanzilla:
@juancampa sadly not, a restart does not trigger the installation of plugins, a manual trigger is still required, copying `package.json` also did not help.

Stanzilla:
That might just be a hyper bug in general, maybe it should update plugins right after startup?

juancampa:
yeah, it works on Linux so I'm surprised about that one

juancampa:
Merging this and running more tests on windows before publishing a canary

voronoipotato:
not sure if this applies but I saw some .hyper.js hardcoded floating around. 

https://github.com/zeit/hyper/issues/3500

 :) hope that helps and if not, sorry for bothering!

PR #3635, opened 2019-05-08
Hyper plugin conflicts

charlesscheuer:
In my plugins on the `.hyper.js` file I have 
```
plugins: [
    'hyper-papercolor',
    'hyper-tab-icons',
    'hyperterm-mactabs',
    'hyper-statusline'
  ]
```
When I first load hyper, I can not see the red, yellow, and green macintosh window circles, but commenting out hyperterm-mactabs and reloading brings them back. If I then uncomment this plugin and reload then the window circles remain in the position. What might be causing this issue and is there a way to resolve?


juancampa:
I'm not sure why this is a PR with 150k lines changed. Is this supposed to be a bug report?

PR #3636, opened 2019-05-08
fix(cli): fix 3.0 compatibility for CLI

chabou:
Fix config file path

### DIFF
--- a/cli/api.js
+++ b/cli/api.js
@@ -6,13 +6,19 @@ const pify = require('pify');
 const recast = require('recast');
 const path = require('path');
 
+// If the user defines XDG_CONFIG_HOME they definitely want their config there,
+// otherwise use the home directory in linux/mac and userdata in windows
+const applicationDirectory =
+  process.env.XDG_CONFIG_HOME !== undefined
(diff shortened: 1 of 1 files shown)


cli/api.js
```diff
@@ -6,13 +6,19 @@ const pify = require('pify');
 const recast = require('recast');
 const path = require('path');
 
+// If the user defines XDG_CONFIG_HOME they definitely want their config there,
+// otherwise use the home directory in linux/mac and userdata in windows
+const applicationDirectory =
+  process.env.XDG_CONFIG_HOME !== undefined
+    ? path.join(process.env.XDG_CONFIG_HOME, 'hyper')
+    : process.platform == 'win32' ? path.join(process.env.APPDATA, 'Hyper') : os.homedir();
```
juancampa: Nice workaround

ZaLiTHkA:
just out of curiosity, is this meant to resolve the config path issue that's causing #3500? my install just upgraded to `3.0.2`, but I still need to keep [the symbolic link](https://github.com/zeit/hyper/issues/3500#issuecomment-490252223) for the config file's old path, otherwise the CLI commands still fail.

PR #3640, opened 2019-05-09
Skip Hyper3 migration if user has made changes 

juancampa:
Before we release Hyper2 -> Hyper3 config migration I wanted to add couple of extra conditions:

 - Don't migrate if the user already changed their Hyper3 file (they already understand what happend)
 - Don't migrate if the file is a symlink (probably not an issue but just in case)


Stanzilla:
Should probably notify the user as well about what happened, that migration was skipped because changes were detected and that the old config file is untouched.

Is it intended that the old config is not renamed to .backup?

juancampa:
Closing in favor of #3643

PR #3643, opened 2019-05-09
Simplify migration of Hyper2 -> Hyper3 config

juancampa:
This whole migration thing is tricky because of all the possible states files can be. I now think that our best approach is to handle 99% of the cases where people are back to default Hyper by automatically migrating their `.hyper.js` and not handle special cases:

 - Only migrate `.hyper.js`. If you have `local` plugins you're probably a collaborator and know how to fix it manually
 - Don't migrate if anything is a symlink. This is an edge case that can have multiple complications so let's just not try to be clever here
 - Don't migrate if the user has made changes to the new config (they already figured it out and might have made improvements over the past config)

There are a number of issues that can arise if we try to handle those two cases automatically and 99% of users are not using either `local` or symlink.

### DIFF
--- a/app/config/import.js
+++ b/app/config/import.js
@@ -1,4 +1,4 @@
-const {moveSync, copySync, existsSync, writeFileSync, readFileSync} = require('fs-extra');
+const {moveSync, copySync, existsSync, writeFileSync, readFileSync, lstatSync} = require('fs-extra');
 const {sync: mkdirpSync} = require('mkdirp');
 const {defaultCfg, cfgPath, legacyCfgPath, plugs, defaultPlatformKeyPath} = require('./paths');
 const {_init, _extractDefault} = require('./init');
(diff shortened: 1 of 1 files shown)


Stanzilla:
meh

juancampa:
> meh

@Stanzilla care to elaborate? we're trying to fix a problem here

Stanzilla:
I'd like how you treated 1) and 3) in the other PR better, but if you don't think it's worth the risk, there is little I can do here.

juancampa:
@Stanzilla the main issue I'm trying to avoid is that people might want to keep these files in their home for convenience, so they might symlink either home -> user-data or the other way around (e.g. If I'm working on a plugin I don't want to go all the way to the new `local/`).

So automatically migration means that people won't be able to do this symlinking because we would be renaming things under their feet.

In any case, this whole thing is a clusterf*** and I want to move on, users will see this once and never again. Thanks for taking a look

juancampa:
Merging this to build a canary and test it

Issue #3648, opened 2019-05-10
Unable to install themes


ThatTonybo:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- Windows 10 Pro, 1809
- 3.0.2
- [Gist can be found here](https://gist.github.com/ThatTonybo/18d24d58002063ea48f48319c77c0fba)
- N/A
- Yes, reproducible in vanilla Hyper

## Issue
I'm unable to install themes or plugins using `hyper i ..`, as it errors somewhere within the CLI.

![image](https://user-images.githubusercontent.com/28176227/57494386-538c0d80-730c-11e9-8fa9-3c3bf803dad5.png)

tats-u:
Are there any differences from #3620 except for the versions of Hyper?

ThatTonybo:
> 
> 
> Are there any differences from #3620 except for the versions of Hyper?

Doesn't seem like it, just that mine affects themes and plugins.

DaneTheory:
It was an issue with the "rigidity" in the previous version of `getProperties` method and its subsequent use in returning values from `getPlugins` and `getLocalPlugins`. The PR I have up fixes these things.

ThatTonybo:
Sounds good, so just wait for the PR to be merged and this should be fixed?

DaneTheory:
@ThatTonybo yep. Not sure why it hasn't been merged yet.

ArgentSun:
Getting a different theme installation error on macOS Mojave after applying this manual fix https://github.com/zeit/hyper/issues/2829#issuecomment-468933774 for #2829 : 

```
| => hyper i verminal
Disabling Chromium GPU blacklist
running in prod mode
electron will open file:///Applications/Hyper.app/Contents/Resources/app.asar/index.html
2019-09-17 13:54:43.761 hyper[77123:6142813] *** WARNING: Textured window <AtomNSWindow: 0x7fc5b39a7360> is getting an implicitly transparent titlebar. This will break when linking against newer SDKs. Use NSWindow's -titlebarAppearsTransparent=YES instead.
Setting Hyper as default client for ssh:// protocol
Linking HyperCLI
{ [Error: EEXIST: file already exists, symlink '/Applications/Hyper.app/Contents/Resources/bin/hyper' -> '/usr/local/bin/hyper']
  errno: -17,
[...]

Issue #3620, opened 2019-05-07
cannot install any plugin


spacekestrel:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 10
- **Hyper.app version**: 3.0.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: 
- **Relevant information from devtools** _(CMD+ALT+I on macOS, CTRL+SHIFT+I elsewhere)_: 
- **The issue is reproducible in vanilla Hyper.app**: 

## Issue
Every time i want to install plugin, i have this messiage:

C:\Users\space\AppData\Local\hyper\app-3.0.0\resources\bin\cli.js:19153
      getProperties = memoize(() => getParsedFile().program.body[0].expression.right.properties),
                                                                                    ^

TypeError: Cannot read property 'properties' of undefined
    at memoize (C:\Users\space\AppData\Local\hyper\app-3.0.0\resources\bin\cli.js:19153:85)
    at d (C:\Users\space\AppData\Local\hyper\app-3.0.0\resources\bin\cli.js:19138:22)
    at memoize (C:\Users\space\AppData\Local\hyper\app-3.0.0\resources\bin\cli.js:19154:34)
    at d (C:\Users\space\AppData\Local\hyper\app-3.0.0\resources\bin\cli.js:19138:22)
    at Object.install (C:\Users\space\AppData\Local\hyper\app-3.0.0\resources\bin\cli.js:19184:37)
    at Args.args.command (C:\Users\space\AppData\Local\hyper\app-3.0.0\resources\bin\cli.js:9934:45)
    at Args.runCommand (C:\Users\space\AppData\Local\hyper\app-3.0.0\resources\bin\cli.js:11318:37)
    at Args.module.exports [as parse] (C:\Users\space\AppData\Local\hyper\app-3.0.0\resources\bin\cli.js:15089:10)
    at main (C:\Users\space\AppData\Local\hyper\app-3.0.0\resources\bin\cli.js:9991:18)
    at Object.module.exports.name (C:\Users\space\AppData\Local\hyper\app-3.0.0\resources\bin\cli.js:10038:1)

rromanv:
I have the same issue

waigx:
same issue here

seergiioo6:
same issue here

[...]

PR #3654, opened 2019-05-10
Fix Hyper (canary) CMDs for install, list and silent failures

DaneTheory:
Issue refs: #3620 and #3648

## The Problem
Running `hyper install <plugin_name>` ultimately failed on each run. This much was obvious.
What was not immediately obvious was the silent failure when running `hyper list`.

## The Solution
This commit primarily focusses on altering the logic behind the following function expressions:
- `getProperties`
- `getPlugins`
- `getLocalPlugins`

By using native, immutable JS Array methods rather than the former practice of hard coding index values.

As a side note, the function `existsOnNpm` needed a return of the `res` added to fulfill it's promise output correctly,  rather than `undefined`. Was not necessary but was another silent failure that wasn't immediately noticeable.

`hyper list` required broadening the context of `getProperties`

Cheers!

### DIFF
--- a/cli/api.js
+++ b/cli/api.js
@@ -54,23 +54,23 @@ const getParsedFile = memoize(() => recast.parse(getFileContents()));
 const getProperties = memoize(() => getParsedFile().program.body.map(obj => obj));
 
 const getPlugins = memoize(() => {
-  let plugins
+  let plugins;
   getProperties().find(property => {
     return Object.values(property.expression.right.properties).filter(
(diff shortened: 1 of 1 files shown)


DaneTheory:
@juancampa or @chabou would either of you mind fixing the lint errors causing the build fail? My IDE is setup using voodoo configs from work or else I'd gladly do this on my own.

DaneTheory:
Nvm. OCD wins again. Seemed like this issue needed immediate attention as users posting in reference to this were beginning to craft some rather unique workarounds. Should help. Cheers

PR #3672, opened 2019-05-15
fix: load scoped plugins

erictaylor:
Scoped plugins (ie. `@scoped/some-package`) are downloaded successfully, but fail to load. This was due to the resolved file name being returned as an empty string from improper string splitting.

This PR is ready for merge.

### DIFF
--- a/app/plugins.js
+++ b/app/plugins.js
@@ -245,7 +245,7 @@ exports.subscribe = fn => {
 function getPaths() {
   return {
     plugins: plugins.plugins.map(name => {
-      return resolve(path, 'node_modules', name.split('#')[0].split('@')[0]);
+      return resolve(path, 'node_modules', name.split('#')[0]);
     }),
     localPlugins: plugins.localPlugins.map(name => {
       return resolve(localPath, name);


j-f1:
I believe this `split` call was added to allow people to specify package versions by writing `package@version`. Would it be better to use a package specifier parser to pull out the package name?

erictaylor:
> I believe this `split` call was added to allow people to specify package versions by writing `package@version`. Would it be better to use a package specifier parser to pull out the package name?

Yeah I assumed the original author had done this for that specific purpose, but it doesn't work obviously for scoped packages. From what I could tell according to the config documentation the version should be specified by `@scope/package#version` (with `@scope/` and `#version` being optional).

This PR assumes that plugins are correctly written in the documented config as `@scope/package#version`.

I would think that plugins should be validated for parsing issues _before_ downloading. This PR change just handles loading of plugins already downloaded.

j-f1:
<code>require(['npm-package-arg'](https://www.npmjs.com/package/npm-package-arg))(spec).name</code> should do the trick.

szul:
Any update on a merger for this, or should people stay far away from naming their plugins with scopes?

chabou:
Oh god.
This was introduce by someone (me) who didn't correctly understand the use of `#`.
I agree with @j-f1. We should use npm/yarn version specification and maybe introduce [npm-package-arg](https://www.npmjs.com/package/npm-package-arg).

But for now, we should fix this issue with scoped packages. User with `package@version` in their config will migrate to `package#version`

Sorry for the delay

Issue #3375, opened 2019-01-01
Upload snap to Snapcraft


dfrankland:
This is a follow up to https://github.com/zeit/hyper/pull/3216#issuecomment-431477529 and #3374 for uploading the snap image to Snapcraft.

Larry1123:
Could the assert file be provided for snapd to accept the .snap file while this is being done? I have been wanting to use the snap version for some time but it seems that it has been moving very slowly to that point. I would love to have details on how to use/test the snap version.

dfrankland:
Sorry for misunderstanding, are you asking for how to install the `.snap`  file?

tetienne:
Any update on this?

zicklag:
I would really appreciate having the snap on the store. Automatic updates on the large list of supported Linux distros would be awesome.

mwoz123:
Any updates? 
There's a howto :
https://snapcraft.io/docs/releasing-your-app

Can those steps be integrated with CI ?

PR #3687, opened 2019-05-21
Support dark mode on macOS

pablopunk:
Using this option of `electron-builder` makes the native title bar to match system theme (for example using [leo/hyper-native](https://github.com/leo/hyper-native))

dark hyper

### DIFF
--- a/package.json
+++ b/package.json
@@ -158,7 +158,8 @@
     },
     "mac": {
       "category": "public.app-category.developer-tools",
-      "extendInfo": "build/Info.plist"
+      "extendInfo": "build/Info.plist",
+      "darkModeSupport": true
     },
     "deb": {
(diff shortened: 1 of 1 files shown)


pablopunk:
@chabou could you take a look at this one? It's a small change but with great impact IMO

I see a lot of PRs are waiting, I would be glad to be a collaborator just to review some, like the snapcraft build, that one seems important for a lot of users https://github.com/zeit/hyper/issues/3375

ppot:
What if you are on a previous macOS build and didn't update. Would It break?

pablopunk:
@ppot I can't test it but I wouldn't say it's likely, this is only an option for electron-builder to make the window dark mode compatible. If you're not in dark mode or in an older build it shouldn't make a difference.

PR #3753, opened 2019-07-02
Added pin to header.

TheRealSyler:
i think that having an option to pin the window is helpful, so i added it this pr is really simple it just adds 
 a button to the header that toggles the alwayOnTop options from electron.


chabou:
Thank you for your work but it should be a plugin. 
You can create your own or make a PR to enhance [hyper-always-on-top](https://github.com/chabou/hyper-always-on-top)

Really sorry but this PR won't be merged.

PR #3757, opened 2019-07-11
Add proxy capabilities

jynolen:
Related to #1612, This small change add the capabilities to set config option for http_proxy and https_proxy.

If those options are defined it's passed to yarn as args.


trickreich:
I like this idea, no change to get this?

giriloou (approved):

giriloou (approved):

Issue #3736, opened 2019-06-16
Hyper window freezes when I try to close it


rofazayn:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Version 1903 Windows Pro Insider Preview
- **Hyper.app version**:  3.0.2

## Issue

I am using hyper with a cmd shell, I also have wsl and oh my zsh set up.
when I run wsl command in my hyper terminal, wsl and ohmyzsh works just fine, but whenever I try to exit the window, it freezes and it stays like that until I kill the process through the Task Manager

Pluriscient:
I can +1 this, also happens when trying to do a full reload. 

OS version and name: Microsoft Windows 10 Education Version	10.0.18362 Build 18362
Hyper.app version: 3.0.2

yoyoys:
me too, wsl always crash when I exit by hyper.
But it safe when I use exit command.

szmarczak:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate



- **OS version and name**: Windows 10 [1903]
- **Hyper.app version**: 3.0.2 (stable)
- **The issue is reproducible in vanilla Hyper.app**: **yes**

## Issue
This (be patient and wait a few seconds):

![3](https://user-images.githubusercontent.com/36894700/59636554-0fcec400-9154-11e9-856d-39ededae5d6f.gif)

szmarczak:
It doesn't crash when you run `exit` on the Linux instance and then close the window.

rofazayn:
> It doesn't crash when you run `exit` on the Linux instance and then close the window.

Hey there, I use wsl, so whenever I type in exit in wsl, I get the linux instance, and if I type in exit again on the linux instance, wsl shell reappears.

And it still crashes when I try to close the window

![Capture](https://user-images.githubusercontent.com/37586909/59693551-05a5d780-91df-11e9-8c42-8d95aa312a57.PNG)

rofazayn:
> It doesn't crash when you run `exit` on the Linux instance and then close the window.

It still crashed, the reason it did that is because I was using oh my zsh, and when I ran the `exit` command, I only exited oh my zsh to the linux instance.

[...]

Issue #3758, opened 2019-07-12
Black Screen on Launch of Hyper [Windows 10]


crunchprank:
## Info

- **OS version and name**: Windows 10 64-bit Version 1809 OS Build 17763.615

- **Hyper.app version**: 3.0.2 (stable)

- **Hyper Config**: https://pastebin.com/raw/F747wqxU

- **Relevant information from devtools**: 

```
Uncaught TypeError: Cannot destructure property `uid` of 'undefined' or 'null'.
    at EventEmitter.u.a.on (bundle.js:1)
    at EventEmitter.emit (events.js:182)
    at t.a.ipcListener (bundle.js:1)
    at EventEmitter.emit (events.js:182)
```

- **The issue is reproducible in vanilla Hyper.app**: Yes

## Issue

I haven't done anything other than go to the official `hyper.is` website, download the Windows executable file, perform the installation, launch Hyper, go to Plugins -> Install Hyper CLI.

Seeing as I've never used this program before, I'm not sure what to expect.

Basically I launch Hyper which by default opens the Windows command prompt. I type `hyper` (which I'm assuming I'm supposed to type to interact with the `hyper cli`), it opens up a black window with a magenta cursor, and then nothing. I cannot type in the window. I provided the output above from the Developer Tools console. [Here is a screenshot](https://i.imgur.com/Hnau4Ga.png) of the Hyper window and dev console side by side.

Searching exisiting bugs that seem to be similar, all resolutions pointed to uninstalling or disabling plugins. However the only plugin I have touched is the official Hyper CLI one.

Any help would be appreciated.

patroclos:
I have the same problem, also when starting hyper via the startmenu item the cmd works, but it doesn't load any plugins.

**OS**: Win 10 18362.239 Hyper 3.0.2

binaryfunt:
I get the same problem, only when I try opening a repository in Hyper from the GitHub desktop app (but I don't get an uncaught type error in the developer console). Hyper opens normally when I open it from the Start menu.

Windows 10 64-bit v1903
Hyper 3.0.2 (stable) with no plugins

It doesn't matter if I set `shell` to `'C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe'` or leave it empty in `.hyper.js`

joonas-fi:
My Hyper screen went black after it worked for many days straight. Just this morning when I came back, it's all black:

[...]

PR #3781, opened 2019-08-02
fix: Hyper window freezes when I try to close it

ivanwonder:
I found the windows freezes because of node-pty can not be killed, and the new version [0.8.1](https://github.com/microsoft/node-pty/releases/tag/0.8.1) of node-pty fix it. and I also add a config `useConpty`. the related issues #3736.
another related issues #3758 happened because of the warmed-up session exit(I don't know why). If set config `useConpty` to false, it will open correctly according to the test on my windows 10  version 1903.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -24,7 +24,7 @@
     "lodash": "4.17.13",
     "mkdirp": "0.5.1",
     "ms": "2.1.1",
-    "node-pty": "0.8.0",
+    "node-pty": "0.8.1",
     "os-locale": "3.1.0",
     "parse-url": "3.0.2",
     "queue": "4.4.2",
--- a/app/session.js
+++ b/app/session.js
@@ -20,6 +20,7 @@ try {
 }
 
 const envFromConfig = config.getConfig().env || {};
+const useConpty = config.getConfig().useConpty;
 
 // Max duration to batch session data before sending it to the renderer process.
 const BATCH_DURATION_MS = 16;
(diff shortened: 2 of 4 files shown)


chabou:
Thank you for your help @ivanwonder on these issues!

Multiple things are mixed in this PR. And I want to be sure that they are related.

Upgrading `node-pty` is a really good catch. Isn't it enough to fix #3736 ? Or do we need your additions about `initialSession` too?

Why do you need `useConpty` config? Can you explain the issue your encountered without it?

ivanwonder:
I cannot say it is enough to fix #3736.
According to the error info of #3736. the problem is the code fire event of `session exit` without `uid`.
https://github.com/zeit/hyper/blob/758e93c21e98556dd50acd81e8dc27615eebe787/lib/index.js#L61
https://github.com/zeit/hyper/blob/758e93c21e98556dd50acd81e8dc27615eebe787/app/ui/window.js#L130
https://github.com/zeit/hyper/blob/758e93c21e98556dd50acd81e8dc27615eebe787/app/ui/window.js#L136
so my thought of the additions about `initialSession` is clean it and creates a new one.
https://github.com/zeit/hyper/blob/758e93c21e98556dd50acd81e8dc27615eebe787/app/ui/window.js#L146
but the new one still exit. **so the main problem is why session exit when open hyper with the hyper cli.**
I discovered that if set the `useConpty` to false, the issues #3736 solved. But I still cannot explain the main problem why session exit and find info about it until now. and the vscode have the setting `Windows Enable Conpty`.

ThraaxSession:
I think many other windows user think is annoying. Do you think it's worth for a new release? I caught this error too.

Stanzilla:
It is, yeah. But we don't have anyone doing releases currently.

PR #3788, opened 2019-08-11
chore(deps): switch to open

jamesgeorge007:
[opn](npmjs.com/package/opn) was deprecated and renamed to [open](npmjs.com/package/open)


jamesgeorge007:
cc @chabou

Issue #3763, opened 2019-07-18
Terminal goes not responding when I close on windows with zsh.


RitikPatni:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **WIndows 10 1903**: 
- **Hyper 3.0.2**: 
- **https://gist.github.com/RitikPatni/02bda7bd2236bfb324f0e365a9fbb1f8**: 
- **Can't collect any info as it goes not responding** _(CMD+ALT+I on macOS, CTRL+SHIFT+I elsewhere)_: 
- **This issue appears when I use zsh as default shell with ohmyzsh**: 

## Issue
I configured my hyper terminal with zsh on windows but after that whenever I try to close it it goes not responding. I've tried to check the console window too but nothing appears there as the app goes not responding.
![Screenshot (2)](https://user-images.githubusercontent.com/16117262/61478784-ece03b80-a9af-11e9-9240-d1cce889e91e.png)

EndlessHollow:
Hello,
I am not sure what is your setup of zsh on windows, but I am using
```
shell: 'wsl.exe',
shellArgs: [],
```
so when you are actually closing it it will kill your wsl aswell.

The second option is to use: `exit`

RitikPatni:
Thanks for the tip. Works well for me

tayyabferozi:
> Hello,
> I am not sure what is your setup of zsh on windows, but I am using
> 
> Hyper.js
> 
> ```
> shell: 'wsl.exe',
> shellArgs: [],
> ```
> 
> so when you are actually closing it it will kill your wsl aswell.
> 
> The second option is to use: `exit`

where am I supposed to write these two lines??

RitikPatni:
@tayyabferozi In your .hyper.js file. You can open it in Preferences. It will be like this [Hyper Config](https://gist.github.com/RitikPatni/02bda7bd2236bfb324f0e365a9fbb1f8)

PR #3794, opened 2019-08-18
Use taskkill on Windows to kill all descendant processes on exit

Lyncredible:
This should fix issues like #3736 and #3763 where bash or zsh is
manually started inside cmd.exe. It also works for other long
running processes like SSH sessions inside cmd.exe or wsl.exe.

### DIFF
--- a/app/session.js
+++ b/app/session.js
@@ -1,3 +1,4 @@
+const {execSync} = require('child_process');
 const {EventEmitter} = require('events');
 const {StringDecoder} = require('string_decoder');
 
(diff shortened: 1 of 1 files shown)


Stanzilla:
Would this not cause a problem if you did this intentionally? Maybe some people want to start a long running process and close the terminal?

Lyncredible:
That is a good question, though I think this is the right behavior.

In the *nix world, background jobs are shutdown when a shell is exited. If a user wants to keep a job running even after the terminal is closed, they should use something like the `disown` or `nohup` command to dissociate the job with the current terminal.

The Windows equivalent seems to be `start`, though that will create a new terminal window.

ivanwonder:
I think it's the problem of node-pty, see the pr #3781, it has been merged,  you can update your code and try whether the problem has resolved or not. @Lyncredible

Lyncredible:
@ivanwonder That did the trick. I will close this PR. Thank you!

@Stanzilla If you look at the `node-pty` PR(microsoft/node-pty#262) that fixed the issue, they are also trying to kill the entire process tree 😃

Stanzilla:
Ah, great we figured it out, thanks guys!

PR #3785, opened 2019-08-06
Update Electron to v6

razfriman:
Update to Electron 6.0.0.

Update packages:
- `electron`
- `electron-builder`
- `electron-rebuild`

Update the `nodeIntegration` default which was changed in Electron 5.0.0.

Add missing `plist` package required from updating the packages

PR #3830, opened 2019-09-23
Update xterm to v4

Stanzilla:

### DIFF
--- a/lib/components/term-group.js
+++ b/lib/components/term-group.js
@@ -87,7 +87,6 @@ class TermGroup_ extends React.PureComponent {
       onResize: this.bind(this.props.onResize, null, uid),
       onTitle: this.bind(this.props.onTitle, null, uid),
       onData: this.bind(this.props.onData, null, uid),
-      onURLAbort: this.bind(this.props.onURLAbort, null, uid),
       toggleSearch: this.bind(this.props.toggleSearch, null, uid),
       onContextMenu: this.bind(this.props.onContextMenu, null, uid),
       borderColor: this.props.borderColor,
--- a/lib/components/term-group.js
+++ b/lib/components/term-group.js
@@ -57,6 +57,10 @@ class TermGroup_ extends React.PureComponent {
     const props = getTermProps(uid, this.props, {
       isTermActive: uid === this.props.activeSession,
       term: termRef ? termRef.term : null,
+      fitAddon: termRef ? termRef.fitAddon : null,
+      searchAddon: termRef ? termRef.searchAddon : null,
+      webLinksAddon: termRef ? termRef.webLinksAddon : null,
+      webGlAddon: termRef ? termRef.webGlAddon : null,
       scrollback: this.props.scrollback,
       backgroundColor: this.props.backgroundColor,
       foregroundColor: this.props.foregroundColor,
--- a/lib/components/term-group.js
+++ b/lib/components/term-group.js
@@ -59,8 +59,6 @@ class TermGroup_ extends React.PureComponent {
       term: termRef ? termRef.term : null,
       fitAddon: termRef ? termRef.fitAddon : null,
       searchAddon: termRef ? termRef.searchAddon : null,
-      webLinksAddon: termRef ? termRef.webLinksAddon : null,
-      webGlAddon: termRef ? termRef.webGlAddon : null,
       scrollback: this.props.scrollback,
       backgroundColor: this.props.backgroundColor,
       foregroundColor: this.props.foregroundColor,
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -162,9 +162,9 @@ export default class Term extends React.PureComponent {
       this.disposableListeners.push(this.term.onTitleChange(props.onTitle));
     }
 
-    // if (props.onActive) {
-    //   this.disposableListeners.push(this.term.addDisposableListener('focus', props.onActive));
-    // }
+    if (props.onActive) {
+      this.disposableListeners.push(this.term.textarea.addEventListener('focus', props.onActive));
+    }
 
     if (props.onData) {
       this.disposableListeners.push(this.term.onData(props.onData));
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -1,4 +1,3 @@
-/* global Blob,URL,requestAnimationFrame,ResizeObserver */
 import React from 'react';
 import {Terminal} from 'xterm';
 import {FitAddon} from 'xterm-addon-fit';
@@ -301,7 +300,7 @@ export default class Term extends React.PureComponent {
     return !e.catched;
   }
 
-  componentWillReceiveProps(nextProps) {
+  UNSAFE_componentWillReceiveProps(nextProps) {
     if (!this.props.cleared && nextProps.cleared) {
       this.clear();
     }
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -73,7 +73,6 @@ const getTermOptions = props => {
     // HACK: Terminal.setOption breaks if we don't apply these in this order
     // TODO: The above notice can be removed once this is addressed:
     // https://github.com/xtermjs/xterm.js/pull/1790#issuecomment-450000121
-    rendererType: useWebGL ? 'webgl' : 'canvas',
     experimentalCharAtlas: useWebGL ? 'webgl' : 'dynamic',
     theme: {
       foreground: props.foregroundColor,
@@ -152,7 +151,6 @@ export default class Term extends React.PureComponent {
       if (this.term.useWebGL) {
         this.term.loadAddon(webGlAddon);
       }
-      // this.term.webLinksInit();
       // this.term.winptyCompatInit();
     }
 
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -411,6 +411,11 @@ export default class Term extends React.PureComponent {
             width: 100%;
             height: 100%;
           }
+
+          .term_wrapper {
+            /* TODO: decide whether to keep this or not based on understanding what xterm-selection is for */
+            overflow: hidden;
+          }
         `}</style>
       </div>
     );
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -144,14 +144,14 @@ export default class Term extends React.PureComponent {
 
     if (!props.term) {
       this.term.attachCustomKeyEventHandler(this.keyboardHandler);
+      this.term.loadAddon(fitAddon);
+      this.term.loadAddon(searchAddon);
+      this.term.loadAddon(webLinksAddon);
       this.term.open(this.termRef);
       fitAddon.fit();
+      this.term.loadAddon(webGlAddon);
       // this.term.webLinksInit();
       // this.term.winptyCompatInit();
-      this.loadAddon(fitAddon);
-      this.loadAddon(webGlAddon);
-      this.loadAddon(searchAddon);
-      this.loadAddon(webLinksAddon);
     }
 
     if (this.props.isTermActive) {
@@ -159,20 +159,20 @@ export default class Term extends React.PureComponent {
     }
 
     if (props.onTitle) {
-      this.disposableListeners.push(this.term.addDisposableListener('title', props.onTitle));
+      this.disposableListeners.push(this.term.onTitleChange(props.onTitle));
     }
 
-    if (props.onActive) {
-      this.disposableListeners.push(this.term.addDisposableListener('focus', props.onActive));
-    }
+    // if (props.onActive) {
+    //   this.disposableListeners.push(this.term.addDisposableListener('focus', props.onActive));
+    // }
 
     if (props.onData) {
-      this.disposableListeners.push(this.term.addDisposableListener('data', props.onData));
+      this.disposableListeners.push(this.term.onData(props.onData));
     }
 
     if (props.onResize) {
       this.disposableListeners.push(
-        this.term.addDisposableListener('resize', ({cols, rows}) => {
+        this.term.onResize(({cols, rows}) => {
           props.onResize(cols, rows);
         })
       );
@@ -180,7 +180,7 @@ export default class Term extends React.PureComponent {
 
     if (props.onCursorMove) {
       this.disposableListeners.push(
-        this.term.addDisposableListener('cursormove', () => {
+        this.term.onCursorMove(() => {
           const cursorFrame = {
             x: this.term._core.buffer.x * this.term._core.renderer.dimensions.actualCellWidth,
             y: this.term._core.buffer.y * this.term._core.renderer.dimensions.actualCellHeight,
@@ -283,7 +283,7 @@ export default class Term extends React.PureComponent {
     if (!this.termWrapperRef) {
       return;
     }
-    this.term.fit();
+    fitAddon.fit();
   }
 
   keyboardHandler(e) {
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -42,23 +42,6 @@ const getTermOptions = props => {
   const needTransparency = Color(props.backgroundColor).alpha() < 1;
   const backgroundColor = needTransparency ? 'transparent' : props.backgroundColor;
 
-  let useWebGL = false;
-  if (props.webGLRenderer) {
-    if (needTransparency) {
-      // eslint-disable-next-line no-console
-      console.warn(
-        'WebGL Renderer has been disabled since it does not support transparent backgrounds yet. ' +
-          'Falling back to canvas-based rendering.'
-      );
-    } else if (!isWebgl2Supported()) {
-      // eslint-disable-next-line no-console
-      console.warn('WebGL2 is not supported on your machine. Falling back to canvas-based rendering.');
-    } else {
-      useWebGL = true;
-    }
-  }
-  Term.reportRenderer(props.uid, useWebGL ? 'WebGL' : 'Canvas');
-
   return {
     macOptionIsMeta: props.modifierKeys.altIsMeta,
     scrollback: props.scrollback,
@@ -73,10 +56,6 @@ const getTermOptions = props => {
     allowTransparency: needTransparency,
     macOptionClickForcesSelection: props.macOptionSelectionMode === 'force',
     windowsMode: isWindows,
-    // HACK: Terminal.setOption breaks if we don't apply these in this order
-    // TODO: The above notice can be removed once this is addressed:
-    // https://github.com/xtermjs/xterm.js/pull/1790#issuecomment-450000121
-    experimentalCharAtlas: useWebGL ? 'webgl' : 'dynamic',
     theme: {
       foreground: props.foregroundColor,
       background: backgroundColor,
@@ -145,13 +124,32 @@ export default class Term extends React.PureComponent {
     this.termWrapperRef.appendChild(this.termRef);
 
     if (!props.term) {
+      let needTransparency = Color(props.backgroundColor).alpha() < 1;
+      let useWebGL = false;
+      if (props.webGLRenderer) {
+        if (needTransparency) {
+          // eslint-disable-next-line no-console
+          console.warn(
+            'WebGL Renderer has been disabled since it does not support transparent backgrounds yet. ' +
+              'Falling back to canvas-based rendering.'
+          );
+        } else if (!isWebgl2Supported()) {
+          // eslint-disable-next-line no-console
+          console.warn('WebGL2 is not supported on your machine. Falling back to canvas-based rendering.');
+        } else {
+          // Experimental WebGL renderer needs some more glue-code to make it work on Hyper.
+          // useWebGL = true;
+        }
+      }
+      Term.reportRenderer(props.uid, useWebGL ? 'WebGL' : 'Canvas');
+
       this.term.attachCustomKeyEventHandler(this.keyboardHandler);
       this.term.loadAddon(fitAddon);
       this.term.loadAddon(searchAddon);
       this.term.loadAddon(webLinksAddon);
       this.term.open(this.termRef);
       fitAddon.fit();
-      if (this.term.useWebGL) {
+      if (useWebGL) {
         this.term.loadAddon(webGlAddon);
       }
       // this.term.winptyCompatInit();
@@ -276,6 +274,7 @@ export default class Term extends React.PureComponent {
 
   resize(cols, rows) {
     this.term.resize(cols, rows);
+    fitAddon.fit();
   }
 
   selectAll() {
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -1,10 +1,9 @@
 /* global Blob,URL,requestAnimationFrame,ResizeObserver */
 import React from 'react';
 import {Terminal} from 'xterm';
-import * as fit from 'xterm/lib/addons/fit/fit';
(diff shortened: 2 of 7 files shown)


Stanzilla:
Can't get it to work atm, setting to WIP

Stanzilla:
TypeError: Invalid value used as weak map key
    at WeakMap.set (<anonymous>)
    at TermGroup_.bind (bundle.js:1)
    at TermGroup_.renderTerm (bundle.js:1)
    at TermGroup_.render (bundle.js:1)
    at h (bundle.js:1)
    at beginWork (bundle.js:1)
    at d (bundle.js:1)
    at f (bundle.js:1)
    at g (bundle.js:1)
    at m (bundle.js:1)
h @ bundle.js:1

LabhanshAgrawal:
Hi, I tried to do this and I'm able to get the terminal window and the shell, but there are some random characters appearing at the top(when using webgl). You can take a look [here](https://github.com/LabhanshAgrawal/hyper/commit/7f46c74ae872db71a171612b0ba38c71f282c5b4)

LabhanshAgrawal:
> ```
> TypeError: Invalid value used as weak map key
>     at WeakMap.set (<anonymous>)
>     at TermGroup_.bind (bundle.js:1)
>     at TermGroup_.renderTerm (bundle.js:1)
>     at TermGroup_.render (bundle.js:1)
>     at h (bundle.js:1)
>     at beginWork (bundle.js:1)
>     at d (bundle.js:1)
>     at f (bundle.js:1)
>     at g (bundle.js:1)
>     at m (bundle.js:1)
> h @ bundle.js:1
> ```
> 
> on
> 
> ```
> class TermGroup_ extends react__WEBPACK_IMPORTED_MODULE_0___default.a.PureComponent {
>         constructor(a, b) {
>             super(a, b),
>             this.bound = new WeakMap,
>             this.termRefs = {},
>             this.onTermRef = this.onTermRef.bind(this)
>         }
>         bind(a, b, c) {
>             this.bound.has(a) || this.bound.set(a, {});
>             const d = this.bound.get(a);
>             return d[c] || (d[c] = a.bind(b, c)),
>             d[c]
>         }
> ```

this is coming because of onURLAbort in term-group.js as it is undefined in props

Stanzilla:
oh thank you, that's indeed better.

lib/components/term.js
```diff
@@ -144,43 +144,43 @@ export default class Term extends React.PureComponent {
...
     if (props.onTitle) {
-      this.disposableListeners.push(this.term.addDisposableListener('title', props.onTitle));
+      this.disposableListeners.push(this.term.onTitleChange(props.onTitle));
     }
 
-    if (props.onActive) {
-      this.disposableListeners.push(this.term.addDisposableListener('focus', props.onActive));
-    }
+    // if (props.onActive) {
+    //   this.disposableListeners.push(this.term.addDisposableListener('focus', props.onActive));
```
LabhanshAgrawal: This might need a bit of digging. I was not able to find any suitable solution when I was looking at this.

Stanzilla: yah same :/

Stanzilla:
@LabhanshAgrawal try this

LabhanshAgrawal:
Facing problems with yarn install with the changed node-pty version. Also the rendererType TermOption cannot be 'webgl' in the new xterm.

Stanzilla:
@LabhanshAgrawal why not? it can be if we load the plugin?

LabhanshAgrawal:
That was the error I'm getting. I think since the renderer has been moved to a plugin, it would set the renderer type when loaded.
@Tyriar might be able to clarify better.

Stanzilla:
Oh I see

LabhanshAgrawal:
Removing the rendererType TermOption and reverting the node-pty version is working for me.

LabhanshAgrawal:
The random characters at the top are also gone

Stanzilla:
Yup. pushed your suggestion, looks good!

LabhanshAgrawal:
Is there any specific reason for changing the node-pty version? 0.9.0-beta26 is not compiling for me (on mac)

Stanzilla:
@LabhanshAgrawal yes we need it for the next Electron update. Which node version are you on?

LabhanshAgrawal:
my node version is 12.6.0

Stanzilla:
@LabhanshAgrawal try with node10, I remember reading about a bug with 12

LabhanshAgrawal:
Getting the same error with node 10

Stanzilla:
With this error message? https://circleci.com/gh/zeit/hyper/2263?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

LabhanshAgrawal:
yes

LabhanshAgrawal:
Also the xterm releases [page](https://github.com/xtermjs/xterm.js/releases) mentions the compatible addons version to be 0.2.1 as of now

LabhanshAgrawal:
getting this warning
```xterm-addon-search@0.3.0-beta2" has incorrect peer dependency "xterm@^3.14.0".```
with 0.3.0-beta2
this has been bumped to ^4.0.0 in version 0.2.1 of the addons

**Comment by Stanzilla** [comment]
hrm odd, @Tyriar ?

**Comment by LabhanshAgrawal** [comment]
@Stanzilla Resize is not working properly. Only the latest opened terminal is getting resized.
So if you split, the new split will reflow on resize but the earlier one will not, same for new tab.

**Comment by LabhanshAgrawal** [comment]
similar behaviour is there for search also.

**Comment by Stanzilla** [comment]
Did you try throwing in a `fitAddon.fit();` in the `resize` function?

**Comment by LabhanshAgrawal** [comment]
Tried it, didn't work. Looks like the resize function for older terminals is not even getting triggered.

File: lib/components/term.js (hunk 1 above)
Comment (ivanwonder) [reply lib/components/term.js]: perhaps use dom listener instead. https://github.com/microsoft/vscode/blob/c46560d371702e8d6ab063e46d254b0b8afeed54/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts#L639

**Comment by Tyriar** [comment]
> rendererType: useWebGL ? 'webgl' : 'canvas',

Yes you definitely don't want this anymore 💥 

> getting this warning
> xterm-addon-search@0.3.0-beta2" has incorrect peer dependency "xterm@^3.14.0".
> with 0.3.0-beta2

That warning should be fixed in 0.2.1 of the addons

File: package.json [hunk 2]
```diff
@@ -213,7 +213,11 @@
     "stylis": "3.5.0",
     "uuid": "3.1.0",
     "webpack-cli": "3.3.7",
-    "xterm": "https://registry.npmjs.org/@zeit/xterm/-/xterm-3.12.0-1.tgz"
+    "xterm": "4.0.2",
+    "xterm-addon-fit": "^0.2.0",
```
Tyriar: You update all the addons to 0.2.1

lib/components/term.js
Tyriar: Yes the focus event was removed because you can just listen to focus on `Terminal.element`, see https://github.com/xtermjs/xterm.js/pull/2008#issue-269040997

> blur -> Removed as you can just listen to the DOM event on Terminal.textarea
> focus -> Removed as you can just listen to the DOM event on Terminal.textarea

lib/components/term.js
```diff
@@ -144,38 +143,44 @@ export default class Term extends React.PureComponent {
...
+      this.term.loadAddon(searchAddon);
+      this.term.loadAddon(webLinksAddon);
       this.term.open(this.termRef);
-      this.term.webLinksInit();
-      this.term.winptyCompatInit();
+      fitAddon.fit();
+      if (this.term.useWebGL) {
+        this.term.loadAddon(webGlAddon);
+      }
+      // this.term.winptyCompatInit();
```
Tyriar: You'll probably want to set `windowsMode: isWindows` in `Terminal.ctor`

Stanzilla:
@LabhanshAgrawal I probably won't have time to work on this in the near future, if you want to take a stab, go ahead :)

Stanzilla:
This error appears when opening a new split

```
TypeError: Cannot read property 'dispose' of undefined
    at disposableListeners.forEach.a (bundle.js:1)
    at Array.forEach (<anonymous>)
    at Term.componentWillUnmount (bundle.js:1)
    at d (bundle.js:1)
    at g (bundle.js:1)
    at commitDeletion (bundle.js:1)
    at b (bundle.js:1)
    at m (bundle.js:1)
    at w (bundle.js:1)
    at z (bundle.js:1)
```

GitSquared:
@Stanzilla you can trace this error to this line:

https://github.com/Stanzilla/hyper/blob/8f306e2e7f8f9f58015d4f56080608ea2a7a304b/lib/components/term.js#L378

lib/components/term.js
```diff
@@ -144,38 +146,44 @@ export default class Term extends React.PureComponent {
...
     }
 
     if (props.onTitle) {
-      this.disposableListeners.push(this.term.addDisposableListener('title', props.onTitle));
+      this.disposableListeners.push(this.term.onTitleChange(props.onTitle));
     }
 
     if (props.onActive) {
-      this.disposableListeners.push(this.term.addDisposableListener('focus', props.onActive));
+      this.disposableListeners.push(this.term.textarea.addEventListener('focus', props.onActive));
```
GitSquared: `addEventListener` doesn't returns an object with a `dispose` function, which means you shouldn't add it to `disposableListeners`.
There needs to be an additional `removeListener` statement to clear it, at around line 380.

GitSquared:
@Stanzilla I don't have push access to your fork, but here's a patch that solves the error.

<details>
<summary>Git diff: Fix undisposable listener error</summary>

```
diff --git a/lib/components/term.js b/lib/components/term.js
index 22f1e70..d34137a 100644
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -166,7 +166,7 @@ export default class Term extends React.PureComponent {
     }
 
     if (props.onActive) {
-      this.disposableListeners.push(this.term.textarea.addEventListener('focus', props.onActive));
+      this.term.textarea.addEventListener('focus', props.onActive);
     }
 
     if (props.onData) {
@@ -378,6 +378,8 @@ export default class Term extends React.PureComponent {
     this.disposableListeners.forEach(handler => handler.dispose());
     this.disposableListeners = [];
 
+    this.term.textarea.removeEventListener('focus', this.props.onActive);
+
     window.removeEventListener('paste', this.onWindowPaste, {
       capture: true
     });
```

</details>

Stanzilla:
Ah indeed! Good catch! Now we just need to fix the resizing issues and this should be good to go.

Stanzilla:
So as soon as there is a second pane, the xterm-screen class of the first one no longer gets updated on resizing

lib/components/term.js
```diff
@@ -143,39 +124,64 @@ export default class Term extends React.PureComponent {
...
     }
 
     if (props.onTitle) {
-      this.disposableListeners.push(this.term.addDisposableListener('title', props.onTitle));
+      this.disposableListeners.push(this.term.onTitleChange(props.onTitle));
     }
 
     if (props.onActive) {
-      this.disposableListeners.push(this.term.addDisposableListener('focus', props.onActive));
+      this.term.textarea.addEventListener('focus', props.onActive);
```
Tyriar: You can still use `disposableListeners` and keep it all here if you want:

```ts
this.term.textarea.addEventListener('focus', props.onActive);
this.disposableListeners.push({ dispose: () =>this.term.textarea.removeEventListener('focus', this.props.onActive) });
```

LabhanshAgrawal: yep, did something similar on my fork, but this looks much cleaner

Stanzilla:
Applied your suggestions

ivanwonder:
@Stanzilla because the constants `fitAddon` is global under the js file, so every new session shares the same constants and only the new session bind to it. just put the constants under the class, `this.fitAddon = new FitAddon()`. this can resolve the resize problem.

LabhanshAgrawal:
> @Stanzilla because the constants `fitAddon` is global under the js file, so every new session shares the same constants and only the new session bind to it. just put the constants under the class, `this.fitAddon = new FitAddon()`. this can resolve the resize problem.

tried this, but no luck

ivanwonder:
it works for me. can I view your code?

GitSquared:
@ivanwonder doesn't work for me either. [My diff](https://pastebin.com/rLjHBjr7).

Stanzilla:
Didn't work here either

LabhanshAgrawal:
> it works for me. can I view your code?

[here](https://github.com/LabhanshAgrawal/hyper/commit/71431f8375c8006cc9c08e8751d130a4130607d9)

ivanwonder:
> > it works for me. can I view your code?
> 
> [here](https://github.com/LabhanshAgrawal/hyper/commit/71431f8375c8006cc9c08e8751d130a4130607d9)

I run your code on my computer of Windows 10, it works fine. I use the hyper config without any plugin and set config `useConpty` to false(because debugging on my windows 10, the hyper app show nothing with `Conpty` true).

Stanzilla:
@ivanwonder That's odd. Do you have any other changes?

ivanwonder:
> > > it works for me. can I view your code?
> > 
> > 
> > [here](https://github.com/LabhanshAgrawal/hyper/commit/71431f8375c8006cc9c08e8751d130a4130607d9)
> 
> I run your code on my computer of Windows 10, it works fine. I use the hyper config without any plugin and set config `useConpty` to false(because debugging on my windows 10, the hyper app show nothing with `Conpty` true).

as I said, I run the code from @LabhanshAgrawal, maybe the only different is the hyper config.

Tyriar:
> @Stanzilla because the constants fitAddon is global under the js file, so every new session shares the same constants and only the new session bind to it. just put the constants under the class, this.fitAddon = new FitAddon(). this can resolve the resize problem.

All addons should definitely have separate instances for each terminal.

Stanzilla:
I pushed those changes now but the issue still persists for me

ivanwonder:
now, I know what you're meaning. I say resolved just for the tab resize problem. I don't realize the split view has the same problem.

ivanwonder:
https://github.com/ivanwonder/hyper/tree/xterm4. try this. when restoring the cache terminal instance, the plugin should be restored too.

GitSquared:
> [ivanwonder/hyper@`xterm4`](https://github.com/ivanwonder/hyper/tree/xterm4) [](/ivanwonder/hyper/tree/HEAD@{2019-10-03T13:41:19Z}). try this. when restoring the cache terminal instance, the plugin should be restored too.

Can confirm this works!

LabhanshAgrawal:
Resizing works!!! need to do the same for the rest of the plugins

LabhanshAgrawal:
One small bit of detail though.
When the current canary version opens, there is a resize action triggered initially.
So the terminal is rendered to the actual size, not the default value.
So the right end of my shell looks like this (the lines are because of the webgl renderer I think)
Screenshot 2019-10-03 at 20 13 20
Here, that is not occurring at the moment. The terminal renders at the default 100 cols, so this happens when the window size at opening is wider than that
Screenshot 2019-10-03 at 20 11 57
It goes away with resize, but this happens at every split and new tab.

Stanzilla:
@ivanwonder oh great find! but why only get the cached ref when webGL is not active?

LabhanshAgrawal:
One nice thing I noticed on merging changes from this pr (including the plugin fix from @ivanwonder and similar change for other plugins) with #3785  - The resize issue in the electron pr also gets fixed.
Notes: The initial resize issue I mentioned earlier is still there.
There is a little rendering issue also
Screenshot 2019-10-03 at 22 04 07
The text seems to have moved up a little because of which the triangle shapes look a little imperfect at the bottom.
You can try this by checking out this [branch](https://github.com/LabhanshAgrawal/hyper/tree/xterm-electron)

Stanzilla:
Hrm, just installed a version with this fix but now Hyper does launch the window never spawns

Stanzilla:
Alright, I think the small remaining resize issue is not important enough to block the merge here, I really want to get this and Electron v6 in. We (I hope you guys continue to contribute?) can work out the small kinks later?

lib/components/term.js
```diff
@@ -146,39 +123,72 @@ export default class Term extends React.PureComponent {
...
-      this.term.winptyCompatInit();
+      if (useWebGL) {
+        this.term.loadAddon(new WebglAddon());
+      }
+    } else {
+      // get the cached plugins
+      this.fitAddon = props.fitAddon;
+      this.search = props.searchAddon;
+      this.webLinksAddon = props.webLinksAddon;
+      this.webGlAddon = props.webGlAddon;
```
ivanwonder: can remove

lib/components/term-group.js
```diff
@@ -57,6 +57,10 @@ class TermGroup_ extends React.PureComponent {
     const props = getTermProps(uid, this.props, {
       isTermActive: uid === this.props.activeSession,
       term: termRef ? termRef.term : null,
+      fitAddon: termRef ? termRef.fitAddon : null,
+      searchAddon: termRef ? termRef.searchAddon : null,
+      webLinksAddon: termRef ? termRef.webLinksAddon : null,
+      webGlAddon: termRef ? termRef.webGlAddon : null,
```
ivanwonder: can remove

lib/components/term.js
```diff
@@ -146,39 +123,70 @@ export default class Term extends React.PureComponent {
...
       this.term.open(this.termRef);
-      this.term.webLinksInit();
-      this.term.winptyCompatInit();
+      if (useWebGL) {
+        this.term.loadAddon(new WebglAddon());
+      }
+    } else {
+      // get the cached plugins
+      this.fitAddon = props.fitAddon;
+      this.search = props.searchAddon;
```
ivanwonder: I think it should be `this.searchAddon`. I never test the search plugin. you should test them before merge.

Stanzilla:
Something is still not quite right https://dl.dropboxusercontent.com/s/e0ejvjwnvqtsugg/2019-10-04_04-37-51.mp4

ivanwonder:
the layout problem? try with vscode. maybe have the same problem.

GitSquared:
> Alright, I think the small remaining resize issue is not important enough to block the merge here, I really want to get this and Electron v6 in. We (I hope you guys continue to contribute?) can work out the small kinks later?

I'm not an approved Hyper contributor but I believe we should try to fix the remaining quirks before merging this - we've made great progress so far and I believe we should be able to iron out the last remaining bugs. I will be looking into it myself.

There's something I'm wondering about, though: this PR puts the latest xterm in place of a custom fork Zeit was maintaining for a while - [zeit/xterm.js](https://github.com/zeit/xterm.js), what about the differences that caused the fork in the first place? Were they all merged into upstream xterm.js, or are they perhaps no longer relevant today?

ivanwonder:
> One nice thing I noticed on merging changes from this pr (including the plugin fix from @ivanwonder and similar change for other plugins) with #3785 - The resize issue in the electron pr also gets fixed.
> Notes: The initial resize issue I mentioned earlier is still there.
> There is a little rendering issue also
> Screenshot 2019-10-03 at 22 04 07
> The text seems to have moved up a little because of which the triangle shapes look a little imperfect at the bottom.
> You can try this by checking out this [branch](https://github.com/LabhanshAgrawal/hyper/tree/xterm-electron)

@Tyriar it seems all the texts are top-aligned. I'm not sure if this's the canvas render problem. chrome version: "76.0.3809.146"

Stanzilla:
@GitSquared afaik all the fork did was bundling the webgl addon and enabling it. sadly @juancampa is no longer responding on the hyper project.

Stanzilla:
@ivanwonder yeah, it's fine in VS code, I tried.

ivanwonder:
after resize, continue to input and check if it's layout correctly. maybe it's not the problem of xterm. there are no windows on my side

Tyriar:
@ivanwonder haven't seen that, something like it happened in Firefox but we fixed that bug a long time ago. Note VS Code doesn't use the webgl renderer yet so it could be related to that?

Stanzilla:
My video about the resizing problem was with the canvas renderer btw

ivanwonder:
@Tyriar it's canvas render. I search the issues and find fixed. but still exist.

Tyriar:
I think this is the line that controls that:

https://github.com/xtermjs/xterm.js/blob/ff790236c1b205469f17a21246141f512d844295/src/browser/renderer/atlas/DynamicCharAtlas.ts#L254

Unless there's a bug in that version of Chrome? 🤷‍♂

Stanzilla:
I just noticed that resizing is fine for all visible lines but off for the ones currently out of view. See https://dl.dropboxusercontent.com/s/7fcwjxyw0yivl86/2019-10-04_16-16-11.mp4

do we have to call .fit() on scroll events?

Tyriar:
> I just noticed that resizing is fine for all visible lines but off for the ones currently out of view.

That's just how winpty/conpty work because they wrap lines before xterm.js gets them. With a native pty they let the terminal emulator wrap them, see https://github.com/microsoft/terminal/issues/405

Stanzilla:
Hrm, how do we best handle that then or just have to live with it?

Tyriar:
@Stanzilla you just have to live with it, as you have already been doing if you're on Windows. Until the upstream issue is fixed.

Stanzilla:
meh, okay. I also noticed that resizing duplicates content in this case, looks like this then 
image

ivanwonder:
this is new xterm 4 with chrome 66 in electron
my local chrome version is the latest  77.0.3865.90.  😂

Tyriar:
@ivanwonder oh it's in the demo? I don't see any issues on latest Edge Dev (Chromium 79):

Screen Shot 2019-10-04 at 8 08 18 AM

Might be related to the font?

ivanwonder:
`DejaVu Sans Mono for Powerline` the font I use

LabhanshAgrawal:
I am not facing any text rendering issues with this pr (using canvas renderer). The issue I had mentioned was on merging xterm4 with electron6 branch (which on itself has rendering issues for me)
The only issue in this pr for me is the initial load resize.
(Note: This is all on mac os, can't say about others)

Stanzilla:
Yeah I think we have to ignore the various Windows quirks and accept them. I only have Windows so here, so kinda need you guys to judge if we are otherwise ready or not.

ivanwonder:
It's ok for me. I can live with that.

GitSquared:
Everything works great for me on Linux, however the "About" pop-up shows the renderer is set to `canvas` - I'm not sure if that's expected?
Latest Hyper stable release defaults to WebGL instead.

LabhanshAgrawal:
@GitSquared WebGL is disabled in this pr because it had some issues. See line 139 in term.js

GitSquared:
@LabhanshAgrawal Oh OK, got it. [I wrote this comment](https://github.com/GitSquared/hyper/commit/1f2ff02329ed464316f8307e3a46f784db85c6fe#diff-268710614c2274f611be667a5f72bb13R133) but didn't think it was copied here.

Stanzilla:
Are we ready to merge then?

LabhanshAgrawal:
There is the initial load resize issue, which is small and the fixing of webgl renderer, both can be tracked in new issues. I think it should be ok to merge now.

Issue #3836, opened 2019-09-30
Porting app folder to TypeScript


LabhanshAgrawal:
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

## Issue

Now that #3816 is merged. Files in lib folder (renderer) can be ported to TypeScript.
I want some suggestions and help on porting the app folder. I have explored two approaches

1. Transpile files and copy assets from app to a new folder compiled_app and use that to run and build hyper - #3837 
2. Transpile TS files in the app folder itself  - #3838 

(tsc on file change in approach 2 not working at the moment, despite similar configuration to approach 1)
(need to configure eslint to ignore the generated files in both branches)

Please suggest any other approach that might be useful and any improvements on the above two.

ppot:
If option 2 doesn't introduce breaking change. I would go with migrating file as they are. no need for duplication.

Stanzilla:
Could reuse the dist folder for 1)?

ppot:
I commented on your 1 PR. instead of `compiled_app/` rename it `target/`

LabhanshAgrawal:
> Could reuse the dist folder for 1)?

There would already be some things in there on building, so decided against using it.

Stanzilla:
Which approach do similar apps like Terminus or FluentTerminal use?

LabhanshAgrawal:
Terminus is using awesome-ts-loader with webpack to convert the ts files to js in a dist folder.
Fluent Terminal is using ts-loader with webpack to bundle ts files to bundle.js in a dist folder.

Stanzilla:
In your personal opinion, are those superior solutions?

LabhanshAgrawal:
I tried out building and running Terminus, it also seems to be bundling the files.
Personally I think those are equivalent to approach 1 which I consider better than approach 2.
We can also choose to bundle the files if we want.

Stanzilla:
Sounds good to me!

LabhanshAgrawal:
Should we bundle?

Stanzilla:
I think so, I don't see any downsides to it? Might be a bit slower maybe?

LabhanshAgrawal:
I tried to setup bundling but getting errors with json modules.
```ERROR in ./app/package.json
Module parse failed: Unexpected token m in JSON at position 0 while parsing near 'module.exports = {"n...'
File was processed with these loaders:
 * ./node_modules/json-loader/index.js
[...]

PR #3837, opened 2019-09-30
Porting app folder to TypeScript by transpiling to another folder

LabhanshAgrawal:
Reference - #3836 
This PR tries to port app folder to TS by copying assets & transpiling to compiled_app folder and using that for building and running Hyper.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 # build output
 dist
 app/renderer
-compiled_app
+target
 bin/cli.*
 
 # dependencies
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # build output
 dist
 app/renderer
+compiled_app
 bin/cli.*
 
 # dependencies
--- a/.eslintignore
+++ b/.eslintignore
@@ -8,3 +8,4 @@ assets
 website
 bin
 dist
+target
--- a/app/utils/colors.ts
+++ b/app/utils/colors.ts
@@ -20,7 +20,7 @@ const colorList = [
 ];
 
 export const getColorMap: {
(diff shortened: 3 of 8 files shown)


.gitignore
```diff
@@ -1,6 +1,7 @@
 # build output
 dist
 app/renderer
+compiled_app
```
ppot: target

Stanzilla:
@LabhanshAgrawal hey there, what's the status on this? what's missing?

LabhanshAgrawal:
I'm facing problems with json modules while bundling.

LabhanshAgrawal:
Hey @Stanzilla I'm not able to get the bundling to work. I tried some approaches but they either give errors during webpack or launch.
This is working in it's current form also. If we need bundling it will take more time probably. Or if it is ok, we can merge this and add bundling later.

Stanzilla:
Sounds fine to me, I'm having the weirdest problem with the Electron branch as well atm, so that will probably take a while to fix. 

What do the others think about this? Would be nice to get some testing on Linux/Mac here since I don't have those. @ivanwonder @GitSquared

GitSquared:
@Stanzilla I haven't looked into what this PR changes but cloning it and running from source works flawlessly, on Linux x64.

Stanzilla:
@LabhanshAgrawal slight problem that I only noticed after merging, there is no `cp -r` on Windows so we will have to use something else here

LabhanshAgrawal:
@Stanzilla looking into it

Stanzilla:
@LabhanshAgrawal I tried cpy-cli here https://github.com/zeit/hyper/pull/3785/commits/fc47596242bd17aa24b371f894dbe034e8f887ee

LabhanshAgrawal:
#3852

Issue #3443, opened 2019-01-25
Add keymap to open hamburger menu without using the mouse


gwojan:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 10, 1803
- **Hyper.app version**: hyper 3.0.0-canary.6 (canary)
## Issue

I would really like to be able to open the hamburger menu from the keyboard. I have checked the default mappings for hyper and don't see anything related to the hamburger menu.

doniyor2109:
Hi. Can I work on this issue?

Stanzilla:
Sure, go ahead :)

doniyor2109:
@Stanzilla Should we add hotkey for linux?

Stanzilla:
I'm not familiar with the conventions on Linux so probably not the best person to ask :(

doniyor2109:
@chabou How about <code>ctrl+\`</code> ?

juancampa:
I think is fine to have the same convention on Linux, as long as it's customizable. I personally use i3wm withhout any decorations so I normally don't see/use menus. And I have removed from Hyper as well with custom CSS in my `~/hyper.js`

jsrj:
If it hasn't already been done, mind if I add the same functionality for Mac?

JoffreyGe:
@jsrj just a question, why do you want to add a shortcut on something ignored on Mac?
`// default: true on Linux, true on Windows, ignored on macOS`
If the burgerMenu can be displayed on macOS could you show me?

AkashDhiman:
Is the issue still active? Can I work on it?

Stanzilla:
Sure, go ahead :)

Amagon96:
I didn't see a merge requesto for this issue, is still opened? 😮

Lunchb0ne:
Maybe add support for remapping it?

LabhanshAgrawal:
> Maybe add support for remapping it?

You can just map the shortcut for `window:hamburgerMenu` in your config

PR #3841, opened 2019-10-02
Add shortcut for opening hamburger menu (Linux)

gabrieldrs:
I also had to edit the command `window:hamburgerMenu`  to account for default config (`showHamburgerMenu===''`)

It seems that there are some useful methods inside lib that would facilitate this command, but these are in Typescript so I couldn't use them (https://github.com/zeit/hyper/blob/canary/lib/components/header.js#L42, https://github.com/zeit/hyper/blob/canary/lib/containers/header.js#L41), I'm not sure if these are meant to be used here in the first place.

It's my first contribution, so I'm not very familiar with the project yet, let me know if there's anything else that I should consider.

Issue #3443

### DIFF
--- a/app/commands.js
+++ b/app/commands.js
@@ -111,7 +111,7 @@ const commands = {
     installCLI(true);
   },
   'window:hamburgerMenu': () => {
-    if (getConfig().showHamburgerMenu) {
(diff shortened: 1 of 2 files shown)


LabhanshAgrawal:
@gabrieldrs can you rebase on canary and push? Ignore the commands.js, just take the linux.json

PR #3843, opened 2019-10-02
Upgrade eslint to v6 and add TypeScript linting

LabhanshAgrawal:
Upgrade eslint
Add eslint ts config
Fix then emitted linting errors


Stanzilla:
needs some follow-up work in file.ts otherwise we're good :)

PR #3850, opened 2019-10-08
Bug fixes for bell configuration

ShaggyDev:
This PR addresses some issues introduced with the return of the bell sound.

1. Bug Fix: Additional null checks to ensure we aren't accidentally accessing members of undefined. If the `bell` option in the config was missing or commented out, it would cause the terminal to cash due to unhandled access of undefined variable.
2. Bug Fix: `term.js` is currently using the `UNSAFE_componentWillRecieveProps()` lifecycle method which is now legacy and will be deprecated in future react releases. It appears that this method is not running as expected, causing the logic within to never be executed. I found this to prevent the use of a custom bell sound since the bell sound logic is relying on the updates to props to determine if it is safe to load the bell sound or if it needs to change the way it loads due to configuration changes. I have updated the file to use the supported lifecycle method `componentDidUpdate`.
3. Lint Adjustment to `file.ts` to suppress warnings.

Please advise if I need to resubmit this PR in a different manner. I based this off of the same fork that implemented the "Return of the Bell" and it appears that it is showing those commits on top of these new ones.

I'd like to get feedback on the replacement of `UNSAFE_componentWillReceiveProps()` before this gets merged. I purposely didn't clean up the unsafe method yet until we decide what we'd like to do.

Thanks for reviewing and, as always, appreciate any and all feedback ;D.

Happy coding!


Stanzilla:
Awesome, thank you. I can just squash to get rid of the noise.

PR #3853, opened 2019-10-09
Add support for font ligatures via xterm-addon-ligatures - Close #3607

### LINKED ISSUE #3607: Font ligatures don't work v3

- **OS version and name**: 10.14.4 macOS Mojave
- **Hyper.app version**: 3.0.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: 
- **Relevant information from devtools** _(CMD+ALT+I on macOS, CTRL+SHIFT+I elsewhere)_: 
- **The issue is reproducible in vanilla Hyper.app**: `Is Vanilla` or with the `hyper-font-ligatures` add-on.

## Issue
I'm using Fira Code font and the ligatures don't work. I found this issue https://github.com/zeit/hyper/issues/914 and tried installing the `hyper-font-ligatures` add-on or adding this css
```
termCSS: `
  x-screen x-row {
    font-variant-ligatures: initial;
  }
`,
```
but none of these seem to work.

GitSquared:
This PR, which is possible thanks to the xTerm update in #3830, enables support for font ligatures thanks to [`xterm-addon-ligatures`](https://github.com/xtermjs/xterm-addon-ligatures).

Note that this addon doesn't support the WebGL renderer (from [`xterm-addon-webgl`](https://github.com/xtermjs/xterm.js/tree/master/addons/xterm-addon-webgl)) yet, but that renderer is considered unstable and was [disabled](https://github.com/zeit/hyper/pull/3830#issuecomment-538749847) during the upgrade to xTerm.js 4 anyway.

Hopefully by the time we get to re-enable the WebGL renderer ligatures will be supported on it. I added a comment to warn future maintainers ([see line 141 in `lib/components/term.js`](https://github.com/zeit/hyper/pull/3853/files#diff-268710614c2274f611be667a5f72bb13R141)).

Related issues:
Close #3607

### DIFF
--- a/app/config/config-default.js
+++ b/app/config/config-default.js
@@ -154,6 +154,9 @@ module.exports = {
     // rendering (slower, but supports transparent backgrounds)
     webGLRenderer: true,
 
+    // if `true` (without backticks and without quotes), Hyper will ignore ligatures provided by some fonts
+    disableLigatures: false,
+
     // for advanced config flags please refer to https://hyper.is/#cfg
   },
 
--- a/lib/components/term-group.js
+++ b/lib/components/term-group.js
@@ -97,6 +97,7 @@ class TermGroup_ extends React.PureComponent {
       quickEdit: this.props.quickEdit,
       webGLRenderer: this.props.webGLRenderer,
       macOptionSelectionMode: this.props.macOptionSelectionMode,
+      disableLigatures: this.props.disableLigatures,
       uid
     });
 
--- a/lib/components/term.js
+++ b/lib/components/term.js
@@ -154,7 +154,7 @@ export default class Term extends React.PureComponent {
       }
       if (props.disableLigatures !== true) {
         this.term.loadAddon(new LigaturesAddon());
-      } 
+      }
     } else {
       // get the cached plugins
       this.fitAddon = props.fitAddon;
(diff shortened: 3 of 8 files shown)


GitSquared:
Peek 2019-10-09 13-18

A little demo.

Stanzilla:
Can you make this optional? There a lot of people that don't like ligatures, including me :)

GitSquared:
@Stanzilla Wouldn't you simply use a font without programming ligatures if you don't want them?

So far the behaviour I've seen in other term emulators is that ligatures are automatically enabled if the font supports it - I'm not against the idea of implementing a boolean option but I think it will just confuse people trying to get ligatures in Hyper.

Just to clarify, only the handful of programming ligatures fonts like Fira Code and Iosevka are impacted by this PR; it doesn't change anything for "regular" fonts.

Stanzilla:
@GitSquared is there a non-ligature version of Fira Code? I think it would just be good practice to add an option for it, VS Code has one as well.

GitSquared:
@Stanzilla Yes, Fira Code is based on Mozilla's [Fira](https://github.com/mozilla/Fira). If we add an option, would you agree to make ligatures enabled by default?

Stanzilla:
yeah would be fine by me

GitSquared:
@Stanzilla I added a `disableLigatures` option, which defaults to `false`, and works as expected.

Stanzilla:
Awesome, build only fails because of a linting issue

GitSquared:
Should be fixed now :+1:

Stanzilla:
Great, thank you!

PR #3861, opened 2019-10-12
Fix undefined resizeObserver error during split

LabhanshAgrawal:
Fix 'cannot read property disconnect of undefined' error which is logged on creating a new split. This was introduced during update to electron v6.
Screenshot 2019-10-12 at 19 38 47


Stanzilla:
good catch!

PR #3862, opened 2019-10-12
Porting lib/utils to TS

LabhanshAgrawal:
This PR attempts to port lib/utils (well most of it) to TypeScript.
I've disabled some eslint typescript rules which were very strict, we can re-enable them once we have more type coverage.
I've used `any` type in a few places where the actual type was not obvious to me.
Feel free to suggest any improvements.


Stanzilla:
Looks like you deleted term-groups.js and then created the ts version instead of git mv, can you change that so we keep history?

LabhanshAgrawal:
I did git mv, in a99b07027bd23fbbfd7f6f11130a46330ecac000, since the file is small, in the combined diff of all the commits, a large part of the file has changed, github is showing it as a different file.

PR #3887, opened 2019-10-19
porting files in lib/ actions, reducers and utils/plugins to ts

LabhanshAgrawal:
This PR attempts to port js files in lib/ actions, reducers and utils/plugins to TypeScript.
I've used `any` type in a few places where the actual type was not obvious to me.
Feel free to suggest any improvements.


Stanzilla:
Would it make sense to add TODO comments to every "any" you are not sure about?

LabhanshAgrawal:
Ideally we should try to get rid of all the `any`s
I am still understanding and learning the codebase so there's a lot of these `any`s

ppot:
I dont have much time but you can tag me if you're not sure

PR #3898, opened 2019-10-22
Raz/GitHub actions

razfriman:
Enable Github Actions CI - Integrated directly into GitHub - Free for open source projects


Stanzilla:
What about Windows? And do you want to restrict it to only canary and master? Also nobody will be able to add environment keys not sure if that would even help then.

LabhanshAgrawal:
@Stanzilla @razfriman I think this should be closed now, since we have actions set up.

PR #3930, opened 2019-10-28
set terminal title to window title

elementc:
I have enabled edits from maintainers in case I'm missing some obvious bits of submitting a PR to this project.

On linux at least, window titles are driven by the contents of the html document's title tag.

Since the terminal title hooks don't update the contents of the title tag, by default, if you have multiple windows open, you get multiple windows with the exact same name, observe the window list at the bottom of this screenshot:

Screenshot from 2019-10-27 19-00-30

This is obviously very confusing for users who navigate windows by alt-tabbing through window lists or by clicking on system trays like I've set up here.

This changeset hooks the title event (I think that's what it's called, I've never worked in a js app so forgive wrong terminology plz :smile: ) and replaces the document title with the new terminal title if the terminal is the active terminal. It also hooks the focus event so that when you click into a tab or split pane, it updates the document title appropriately as well. Observe the window list ath the bottom of this screenshot:

Screenshot from 2019-10-27 19-03-04

(These are fish-style window titles in my personal configuration, but a bash or zsh user would see whatever they've configured their shell to set their window title to be, eg user@hostname:~ etc.)

Let me know if there's any tests I need to update or a better way to replace the contents of this tag, I'm mostly a python / C++ guy these days so I'm a little winging it here.


elementc:
ooof, im a little out of date. give me a sec, i'll rebase

elementc:
ok, NOW we're ready for review, :)

lib/components/term.js
```diff
@@ -169,6 +170,14 @@ export default class Term extends React.PureComponent {
 
     if (props.onTitle) {
       this.disposableListeners.push(this.term.onTitleChange(props.onTitle));
```
ppot: Is this still needed?

ivanwonder: this line changes the tab title, but the PR wants to change the page title to identify different hyper windows.
https://github.com/zeit/hyper/blob/621b509a0685323a8a55e04850dff97a79979ffa/app/index.html#L4
I think this can be achieved by a plugin, or open them on a single window, and distinguish by tab title.

elementc: Re:

> I think this can be achieved by a plugin, or open them on a single window, and distinguish by tab title.

I have a plugin already which does what I want-
https://github.com/elementc/hyper-window-titles

I'm proposing this for inclusion in mainline hyper because it seems kind of silly to be unable to distinguish between windows of your terminal in your system tray by default.

Specific to distinguishing by tab title: I don't always have a window visible when I want to surface it. This helps in that case. The below code: 

1: Stashes the new term title on the terminal object (I think? this.term) unconditionally every time onTitleChange fires.
2: Updates the window title if the terminal that fired this is the currently active term.
3: The focus() addition below that updates the window title with the stashed term title whenever a term becomes active.

This method means that it works with tab-terms AND split-pane-terms.

ivanwonder:
How about adding a new action `UI_SET_WINDOW_TITLE`, like the code below.
https://github.com/zeit/hyper/blob/621b509a0685323a8a55e04850dff97a79979ffa/lib/actions/sessions.ts#L111
they share the same title

lib/components/term.js
```diff
@@ -260,6 +269,7 @@ export default class Term extends React.PureComponent {
 
   focus() {
     this.term.focus();
+    document.getElementsByTagName('title')[0].text = this.term.title;
```
ivanwonder: Just use the assignment statement `document.title = newTitle;`.

elementc: sure, done, didn't know about that one :)

elementc:
re @ivanwonder above:
> How about adding a new action `UI_SET_WINDOW_TITLE`, like the code below.
> 
> https://github.com/zeit/hyper/blob/621b509a0685323a8a55e04850dff97a79979ffa/lib/actions/sessions.ts#L111
> 
> they share the same title

I don't super see this as a new action, this PR is pretty specifically advocating that the window/UI title should always match the current active terminal's shell-set title. If plugin authors wish, they can still misdirect this effect fairly straightforwardly by decorating the term component and intercepting the ontitle and onfocus events, as I did in the plugin i mentioned above (https://github.com/elementc/hyper-window-titles/blob/master/index.js), only now it'll have more of the effects they expect, ie actually setting the document and therefore OS window title. I think adding a new action here would just lead to a more confusing API- as an application, the window title should match the xterm title, and if a modder wants to change either, we should not surprise them and give one hook  for changing both.

r3nanp (approved):

izaccavalheiro (approved):

PR #3937, opened 2019-10-30
Removed optimistic session initialization

### LINKED ISSUE #3770: Initial hyper session is not created correctly

- **OS version and name**: macOS 10.14.5
- **Hyper.app version**: 3.0.2

## Issue
Hi :) Thanks for your work on this!

I use https://github.com/hharnisc/hypercwd and noticed the following bug https://github.com/hharnisc/hypercwd/issues/71:

**The problem is:**
The initial working directory is not set correctly for the very first hyper session, but people noticed that after a full reload the CWD was set correctly again. The issue  on `hypercwd` seems a bit stale and nobody seems to know how to fix it, so I took a little dive into the code and this is what I found:

**First:** the bug does not seem to come from the `hypercwd` codebase, but rather from this codebase.

The problem seems to be located in the optimistically/eagerly creating an initial session here: https://github.com/zeit/hyper/blob/c13c8a5c52cbc8e02098244193a4e715fffcd902/app/ui/window.js#L140
To be more exact in `createInitialSession()`:
https://github.com/zeit/hyper/blob/c13c8a5c52cbc8e02098244193a4e715fffcd902/app/ui/window.js#L124

Here `createSession()` is optimistically called without passing any parameter, although the function is defined as `createSession(extraOptions = {})`. Those `extraOptions` are used to set the `CWD` (among other things correctly), see: https://github.com/zeit/hyper/blob/c13c8a5c52cbc8e02098244193a4e715fffcd902/app/ui/window.js#L111

```js
    function createSession(extraOptions = {}) {
      const uid = uuid.v4();

      const defaultOptions = Object.assign(
        {
          rows: 40,
          cols: 100,
          cwd: process.argv[1] && isAbsolute(process.argv[1]) ? process.argv[1] : homeDirectory,
          splitDirection: undefined,
          shell: cfg.shell,
          shellArgs: cfg.shellArgs && Array.from(cfg.shellArgs)
        },
        extraOptions, // <-- this is the important line
        {uid}
      );
      const options = decorateSessionOptions(defaultOptions);
      const DecoratedSession = decorateSessionClass(Session);
      const session = new DecoratedSession(options);
      sessions.set(uid, session);
      return {session, options};
    }
```

This optimistically creating an initial session without passing the correct `extraOptions` is the underlying problem for the mentioned issue. The CWD is set incorrectly, because it is not set at all. After a full reload however it is then overwritten correctly.

I am unsure about the reasoning for this optimistic session creating and whether it is a technical necessity or a performance optimization. 

I tried removing https://github.com/zeit/hyper/blob/canary/app/ui/window.js#L140 this line and everything seemed to still work and the bug was fixed.

I hope this analysis provides enough insight for you to come up with a decision on how  to proceed quickly as the issue really is quite annoying.

Thanks for your help, have a nice day :)

ahmedx2:
Closes #3770 

This small PR removes the optimistic session creation (the `createInitialSession()` call in line 160 of `app/ui/window.js`), which causes a bug initialization parameters do not take effect unless the user manually executes a full reload (#3770) e.g. [the hypercwd plugin's initialWorkingDirectory config option has no effect until a full reload is run](https://github.com/hharnisc/hypercwd/issues/71).  

After some digging, I've found that this optimistic session creation added for a modest startup time improvement in [#3352](https://github.com/zeit/hyper/pull/3352), so I reckon there will not be an unacceptable increase in startup time for any supported platform.

I figure this PR is ready to merge. Manual testing suggest that these slight changes fix the problem without a noticeable difference in startup time or performance in general.

### DIFF
--- a/app/ui/window.js
+++ b/app/ui/window.js
@@ -129,35 +129,8 @@ module.exports = class Window {
       return {session, options};
     }
 
-    // Optimistically create the initial session so that when the window sends
-    // the first "new" IPC message, there's a session already warmed up.
-    let initialSession = null;
-    function createInitialSession() {
-      let {session, options} = createSession();
-      const initialEvents = [];
-      const handleData = data => initialEvents.push(['session data', data]);
-      const handleExit = () => {
-        // the warmed up session should be cleaned if exit before add a new session.
-        session.removeListener('data', handleData);
-        session.removeListener('exit', handleExit);
-        initialSession = null;
-      };
-      session.on('data', handleData);
-      session.on('exit', handleExit);
-
-      function flushEvents() {
-        for (let args of initialEvents) {
-          rpc.emit(...args);
-        }
-        session.removeListener('data', handleData);
-        session.removeListener('exit', handleExit);
-      }
-      return {session, options, flushEvents};
-    }
-    initialSession = createInitialSession();
-
     rpc.on('new', extraOptions => {
-      const {session, options} = initialSession || createSession(extraOptions);
+      const {session, options} = createSession(extraOptions);
 
       sessions.set(options.uid, session);
       rpc.emit('session add', {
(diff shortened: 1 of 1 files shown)


Stanzilla:
Nice, thank you!

ppot:
i like code removal 🎉

Issue #3955, opened 2019-11-07
Blurry font rendering across middle of screen when maximised


xandora:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Ubuntu 19.10
- **Hyper.app version**: 3.0.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/xandora/ccefa3da543e6a47e3efd7e1205f8fa3
- **Relevant information from devtools** _(CMD+ALT+I on macOS, CTRL+SHIFT+I elsewhere)_: N/A (or maybe I just don't know if any of it's relevant)
- **The issue is reproducible in vanilla Hyper.app**: `Is Vanilla`

## Issue
There is an obvious blurriness to text across the middle third of the screen when Hyper is running maximised. This does not occur when the app window has been resized to less than maximised, or when the app is running in fullscreen mode (F11).

Resized:
![Screenshot from 2019-11-07 14-16-00](https://user-images.githubusercontent.com/2021268/68351723-2e684280-0169-11ea-9fef-6d0326d7340d.png)
Maximised:
![Screenshot from 2019-11-07 14-16-06](https://user-images.githubusercontent.com/2021268/68351728-31fbc980-0169-11ea-94b6-a9926882490d.png)
Fullscreen:
![Screenshot from 2019-11-07 14-16-09](https://user-images.githubusercontent.com/2021268/68351730-33c58d00-0169-11ea-955b-00fe8ffc2446.png)

xandora:
This also seems to happen at certain vertical resolution ranges:

* 700px x 245px -> 700px x 263px
* 700px x 585px -> 700px x 605px

I think these are probably related to the breakpoints for number of visible lines, as it only changes from blurry to sharp after an additional line is snapped in.

cyh0968:
Oh if you don't mind, I will investigate this issue!

AnatoleLucet:
Looks like subpixel rendering to me.
One dirty way to fix it is to remove the fixed height on the canvas:
```js
css: `
  .xterm canvas {
    height: auto !important;
  }
`,
```

linaori:
Just encountered this issue in a maximized state but only appeared on 1 tab. Applying the above CSS "fixed" it.

PR #3981, opened 2019-11-16
Fix Issue-3955 Blurry Text

cyh0968:
Issue #3955, there is a blurry text issue when resizing the window size or zooming in/out. 
I found out that the blurry text is related to the font-size and line-height. 

Interestingly, the font-size is an even number, it becomes blurry when the line-height is 1 but not line-height is over 1.1. In contrast, a font size is an odd number, text look blurry when the line-height is over 1. I fix the issue by adding line-height in ui.js.

Perhaps, it may need to be addressed in a different way. So I hope this code will be reviewed. 

Thanks.

### DIFF
--- a/lib/reducers/ui.js
+++ b/lib/reducers/ui.js
@@ -132,8 +132,8 @@ const reducer = (state = initial, action) => {
             }
 
             if (config.fontSize) {
-                ret.lineHeight = config.lineHeight;
-              if(config.fontSize % 2 === 0){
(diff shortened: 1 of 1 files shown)


Stanzilla:
That is more of a workaround than a fix though, Is there any documentation (upstream for Chrome or so) about similar issues? Maybe we should add comments to that part of the code that documents why we had to do this.

PR #3982, opened 2019-11-17
Fix os-locale format

bet4it:
`os-locale` has changed the format of result in `4.0.0` (`en_US` → `en-US`) : sindresorhus/os-locale@2a3fccb
And we updated `os-locale` to `4.0.0` in #3913
So we need to fix the using of `os-locale`

### DIFF
--- a/app/session.js
+++ b/app/session.js
@@ -90,7 +90,7 @@ module.exports = class Session extends EventEmitter {
       {},
       process.env,
       {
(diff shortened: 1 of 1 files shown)


bet4it:
You can try `LANG=en-US.UTF-8 LC_ALL=en-US.UTF-8 htop` in hyper.
It can't be displayed correctly.

Stanzilla:
Alright, thank you then!

PR #4003, opened 2019-11-24
remove getLastTermIndex

LabhanshAgrawal:
sessions is an object {[uid]: session} not an array
getLastTermIndex doesn't seem to be referenced anywhere though, we can also remove it.


ppot:
You can remove it instead

LabhanshAgrawal:
removed

PR #4059, opened 2019-12-10
Fixes Issue-3798 add funtionality Copy with ctrl+c only when highlighted

cyh0968:
This is a temporary solution to add functionality since there is no direct way to access the highlighted text in the Hyper. It should be reviewed first then it will be enhanced furthermore later.

This code detects clipboard and when the string from the clipboard is not zero, ctrl + c key combination works as a copy key. In other cases, it works as a break session key.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   install:
     macos:
-      xcode: "9.2.0"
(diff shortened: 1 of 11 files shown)


Stanzilla:
Looks like your branch was not updated to date

Stanzilla:
@cyh0968 can you rebase this, please?

PR #4021, opened 2019-11-28
port es5 code to es6

LabhanshAgrawal:
used [lebab](https://github.com/lebab/lebab) to port es5 code to es6 with some manual fixes

PR #4060, opened 2019-12-11
change export syntax in app/plugins/install.js

LabhanshAgrawal:
It is being imported as named import in [app/plugins.js](https://github.com/zeit/hyper/blob/5b481bbc621666728a7e9ddf7b6dd911b8edb9f8/app/plugins.js#L11) (#4021 ), so changing the export.


LabhanshAgrawal:
This fixes the js error currently being triggered on changing preferences.

PR #4065, opened 2019-12-14
setup ava for ts, port test/ to ts

LabhanshAgrawal:
Added ava typescript configuration, and port tests to ts.
In index.ts changed isDevToolsOpened() with webContents.isDevToolsOpened() [ref](https://electronjs.org/docs/all#contentsisdevtoolsopened)
Found it through a ts error :)


LabhanshAgrawal:
the test was not included in yarn run test, need to check the devtools change.
Will confirm when ready to merge.

LabhanshAgrawal:
ready to merge
[ref](https://github.com/electron-userland/spectron/blob/master/lib/spectron.d.ts#L202)

PR #4081, opened 2019-12-17
Fix issue where hyper {command} fails

mikemcbride:
This is an attempt at fixing #3500. I am not super familiar with this codebase but I logged out the values here and found that when this was failing, it needed to be grabbing values from the other index on `body`. This is a relatively naive approach since I don't fully understand what's going on, but it does fix the issue as far as I was able to test. I don't really know how it even gets in a state where this fails, but as I said this seems to do the trick. Definitely open to feedback or other approaches...

cc: @chabou since he is assigned to the open issue

### DIFF
--- a/cli/api.js
+++ b/cli/api.js
@@ -50,7 +50,8 @@ const getFileContents = memoize(() => {
 
 const getParsedFile = memoize(() => recast.parse(getFileContents()));
 
-const getProperties = memoize(() => getParsedFile().program.body[0].expression.right.properties);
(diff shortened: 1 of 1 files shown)


Stanzilla:
Hey thanks for that! Hopefully someone will have time to review this, I am pretty busy with work at the moment.

mikemcbride:
Happy to do it. No worries man, totally understand. Honestly that’s the same reason it took me this long to get a PR up for this.

mikemcbride:
I think this was implemented in a much more robust manner in #4086, so we can close this out

PR #4100, opened 2019-12-23
add app/node_modules to circleci persist list

LabhanshAgrawal:
To fix 'module not found' error on CircleCi build


LabhanshAgrawal:
@Stanzilla can you please merge this, I'm quite positive that this should fix the ci issue.

Stanzilla:
I don't think it's a fix, more of a workaround because it does not build locally here as well

PR #4102, opened 2019-12-25
fix babel command syntax and update

LabhanshAgrawal:
Used the fix from https://github.com/babel/babel/issues/10922#issuecomment-568774757
as https://github.com/babel/babel/pull/10923 might take time to merge and get released.


Stanzilla:
Awesome!

PR #4116, opened 2019-12-27
Add action types

LabhanshAgrawal:
Adding types for actions, to be used in reducer.


lib/constants/ui.ts
```diff
@@ -22,3 +22,104 @@ export const UI_ENTER_FULLSCREEN = 'UI_ENTER_FULLSCREEN';
...
+}
+interface UIContextmenuOpenAction {
+  type: typeof UI_CONTEXTMENU_OPEN;
+}
+interface UICommandExecAction {
+  type: typeof UI_COMMAND_EXEC;
+}
+
+export type UIActions =
+  | UIFontSizeSetAction
```
ppot: Shouldn't there be no `|` before `UIFontSizeSetAction ` ?

LabhanshAgrawal: It's actually inserted by Prettier, allows for easier addition/removal/rearranging of values. Similar to extra comma after the last value in a list.

ppot: Oh okay

LabhanshAgrawal: It's just syntax sugar, doesn't impact code logic/execution

LabhanshAgrawal:
@ppot Is there anything else blocking this?

PR #4133, opened 2020-01-02
Add Hyper Dispatch type and add/update action creators typings

LabhanshAgrawal:
Add the type for the store dispatch method which will allow type checking actions during dispatch.


LabhanshAgrawal:
looks like I missed some stuff, will update, and notify when ready to merge.

LabhanshAgrawal:
Ready to merge
Updated types for action creators (which were already in ts) and ported the rest.
Some files had quite a bit of changes, so put the renames in a separate commit, don't squash while merging

PR #4136, opened 2020-01-05
enable webgl render

ivanwonder:
There are still some issues with [webgl render](https://github.com/xtermjs/xterm.js/issues?q=is%3Aopen+is%3Aissue+label%3Aarea%2Faddon%2Fwebgl)
Go to `about hyper` to check the render engine.


Stanzilla:
Better wait until there are a few less issues

PR #4150, opened 2020-01-11
Add github actions

LabhanshAgrawal:
Added test & build on windows, mac and linux with build artifact upload

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,5 @@
 * text=auto
 *.js text eol=lf
+*.ts text eol=lf
+*.tsx text eol=lf
 bin/* linguist-vendored
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,0 +1,118 @@
+name: Node CI
+on: [push, pull_request]
+jobs:
+  ci_macos:
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        node-version: [12.x]
+    steps:
+      - uses: actions/checkout@v1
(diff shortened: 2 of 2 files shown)


LabhanshAgrawal:
You can see a sample run [here](https://github.com/LabhanshAgrawal/hyper/runs/384730980)

Stanzilla:
I like it! Would probably make every `name` entry in title case.

LabhanshAgrawal:
> I like it! Would probably make every `name` entry in title case.

done

Stanzilla:
@LabhanshAgrawal Just noticed, you didn't do the icon replacement when on canary stuff, https://github.com/zeit/hyper/blob/canary/appveyor.yml#L27 if we add that, I think we can remove appveyor and travis at least.

LabhanshAgrawal:
Oh, I hadn't looked at appveyor or travis configs.
I'll add that.
I also noticed that since we have trigger on push and pull request, dependabot pr's are triggering double the jobs.
For that, will add a condition for push jobs to run only on master and canary

Stanzilla:
Yup, that makes sense!

LabhanshAgrawal:
> @LabhanshAgrawal Just noticed, you didn't do the icon replacement when on canary stuff, https://github.com/zeit/hyper/blob/canary/appveyor.yml#L27 if we add that, I think we can remove appveyor and travis at least.

Doesn't appveyor and travis also include the setup for releasing? So, shouldn't we keep them, to do any releases in future?

Stanzilla:
Oh hrm, well I actually don't know how they did that. Just that they used `now release`

PR #4176, opened 2020-01-15
add back the useCapture option in mousetrap

### LINKED ISSUE #4173: Keymaps regression
This bug was introduced in #3957.

We need to init `Mousetrap` with `true` as the second argument: 
https://github.com/zeit/hyper/blob/d5d5bcc74ca8de4c568c410fe042b5ef353baf67/lib/containers/hyper.js#L58
But we remove this argument after #3957:
https://github.com/zeit/hyper/blob/a1eb84d8a7c277ddf29d739e46cde14c4c14b4f7/lib/containers/hyper.tsx#L62

Refer: ccampbell/mousetrap#400

LabhanshAgrawal:
Fixes #4173

### DIFF
--- a/lib/containers/hyper.tsx
+++ b/lib/containers/hyper.tsx
@@ -55,7 +55,7 @@ class Hyper extends React.PureComponent<any, any> {
 
   attachKeyListeners() {
     if (!this.mousetrap) {
-      this.mousetrap = new Mousetrap();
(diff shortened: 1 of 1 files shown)


LabhanshAgrawal:
I was having trouble with overriding the type definition, so just used `any` as a workaround.

Issue #4115, opened 2019-12-26
Clean up babel/transpiler stuff


Stanzilla:
Since we're only targeting electron, we could just move everything over to es2018 by default and stop transpiling a few things, no?

ivanwonder:
@LabhanshAgrawal 
One problem. If hyper use the babel to compile the ts file. Why still use `tsc` here. It can be removed? 
https://github.com/zeit/hyper/blob/5bb03972d78ff885df46c9dad692ef22f9a9c047/package.json#L8
If I am wrong, correct me.

LabhanshAgrawal:
Babel bundles the files from lib, but it doesn't check for type errors

Stanzilla:
@LabhanshAgrawal what do you think about the es2018 thing though?

LabhanshAgrawal:
I think it makes sense.
The codebase is divided in 4 parts
```
app/  - already getting transpiled to es2018 by tsc.
lib/  - bundled by webpack (babel-loader) to es5 (no target specified in babel config)
        and then minified by babel (cli)
cli/  - bundled by webpack (babel-loader) to es5 (no target specified in babel config)
test/ - no transpilation, ava running with ts-node
```
From what I understand
We can configure Babel to target specific versions of electron and node (which support es2019 currently).
Also possible to take the versions from package.json, which will require to use babel.config **.js** or provide the config through webpack
For lib/ I think we can do the minification during the webpack run itself, when env is production and remove the babel cli stuff

Please correct if I missed something

LabhanshAgrawal:
Also, transpiling everything atleast once is a necessity as we need to get js from ts

Stanzilla:
yup that does look correct. should probably try moving the babel stuff to es2018 first and see how it works

PR #4191, opened 2020-01-20
Terser

Stanzilla:


Stanzilla:
@LabhanshAgrawal hey, what do you think about this? File size is a bit bigger because I only went with the default options for terser but otherwise looks cool?

webpack.config.js
```diff
@@ -53,6 +54,10 @@ module.exports = [
         }
       ])
     ],
+    optimization: {
+      minimize: true,
+      minimizer: [new TerserPlugin()]
+    },
```
LabhanshAgrawal: No need to add it here, as it(webpack job) is not really bundling anything (null-loader), just used to copy assets.

LabhanshAgrawal:
Looks good, Is this for #4115 ?
What's the advantage of using terser(after required config) over babel minify?

Stanzilla:
Yes, part of it. Mostly minify is pretty dead at the moment https://github.com/babel/minify/issues/952

LabhanshAgrawal:
Can you add a condition for checking the isProd variable, would be easier to debug during development

LabhanshAgrawal:
Need to change the webpack config a little to use the type information. I can't add changes to the pr, so I've pushed it [here](https://github.com/LabhanshAgrawal/hyper/tree/terser), you can cherrypick the latest commit, if the change seems ok to you.
Other than the above, the rest looks cool, we can configure terser after this.

PR #3769, opened 2019-07-26
Adding logic to support working directory config

kaushikchaubal:

PR #4200, opened 2020-01-25
Fix default startup directory

### LINKED ISSUE #4197: Default startup directory changed in latest canary build
- [ ] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version

- **OS version and name**: Windows 10 Version 1909 (OS Build 18363.535)
- **Hyper.app version**: 3.1.0-canary.1

## Issue
All shells now default to `~/AppData/Roaming/Hyper` instead of `~` or the otherwise default for the shell. This breaks all compat with shells, and forces you to configure each shell and .bashrc scripts outside of Hyper with the knowledge that you will then wrap them with Hyper. *Please revert to using ~ as the default.*

LabhanshAgrawal:
Fixes #4197 

The bug was introduced in #3769 , which probably went unnoticed because on mac and linux default config dir is home.

### DIFF
--- a/app/ui/window.ts
+++ b/app/ui/window.ts
@@ -6,7 +6,7 @@ import fileUriToPath from 'file-uri-to-path';
 import isDev from 'electron-is-dev';
 import updater from '../updater';
 import toElectronBackgroundColor from '../utils/to-electron-background-color';
-import {icon, cfgDir} from '../config/paths';
+import {icon, homeDirectory} from '../config/paths';
 import createRPC from '../rpc';
 import notify from '../notify';
 import fetchNotifications from '../notifications';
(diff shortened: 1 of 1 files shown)


ppot:
Honestly, can you rework this workingDirectory thing ? I know you where not the one who made the original, but `homeDirectory ` was change for `cfgDirg` and also, this should be a plugin thing. 
Also, this should be a profileDirectory but since mostly not everyone will use it, it's just noice. 

The comment  `// set custom startup directory (must be an absolute path)` should say instead
`// set profile directory for default startup (must be an absolute path)`

Stanzilla:
Could be done in a follow up PR, imo. Merging this now to get the fix in

LabhanshAgrawal:
@ppot can you elaborate a bit, not sure if I understand your comment fully.

PR #4261, opened 2020-02-15
use native-reg to add cli to path

### LINKED ISSUE #4190: Hyper breaks $PATH if path contains unicode

- **OS version and name**: Windows
- **Hyper.app version**: 3.0.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: Irrevalant

## Issue
When Hyper append its own path on the PATH environment variable, it replaces the unicode characters in the path with question mark or �.

The path of the Hyper is unaffected, but this affects all other paths in $PATH.

LabhanshAgrawal:
Fixes #4190 

Using `native-reg` to read and write to registry on windows, for adding cli to path.
native-reg is available only for windows, so added it as an optional dependency.
Had to add the typings from native-reg to the repo, to use with `require`.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,6 +4,7 @@ app/static
 app/bin
 app/dist
 app/node_modules
+app/typings
 assets
 website
 bin
--- a/app/package.json
+++ b/app/package.json
@@ -35,5 +35,8 @@
     "shell-env": "3.0.0",
     "uuid": "3.4.0",
     "winreg": "1.2.4"
+  },
+  "optionalDependencies": {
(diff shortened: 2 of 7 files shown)


LabhanshAgrawal:
The typings would need to be copied from `native-reg` if it's updated. It can be added in the `postinstall` script if desired.
I will create an issue to track replacing `winreg` with `native-reg` in rest of the places where it's currently used. We can add that to 3.2.0 milestone.

BasixKOR:
Tried on my machine and it works as expected. Nice work!

PR #4357, opened 2020-03-07
update to electron 8 and fix initial cwd

### LINKED ISSUE #4295: Updating to electron 8
#4283 is failing because we need newer node-abi also for electron-rebuild to work (see [this](https://github.com/LabhanshAgrawal/hyper/tree/electron8)).

Updating to electron 8 introduces some issues though.
One I've been able to identify is with `cwd`, it's getting set as undefined and thus hyper starts at `/` when opened or the hyper git folder during dev with `yarn run app`.
Explanation:
While creating a new session
https://github.com/zeit/hyper/blob/d0ce94ce8ea690012778069db5deb80d624f73c8/app/ui/window.ts#L135-L136
the `extraOptions` can have these keys `splitDirection, cwd, activeUid, isNewGroup` normally
Out of these, value of `cwd` is always `undefined` as it's being taken from ui state which doesn't have cwd initially and the `SESSION_SET_CWD` action is not triggered anywhere in the codebase.

This didn't cause us any problems earlier as until now the ipc had been removing the keys with `undefined` values and we're using the extraOptions here
https://github.com/zeit/hyper/blob/d0ce94ce8ea690012778069db5deb80d624f73c8/app/ui/window.ts#L118-L127
So, cwd was always getting set to `workingDirectory`

But in electron 8, this behaviour of ipc has changed https://github.com/electron/electron/pull/20214
it now sends the object properly, so now `extraOptions` has the `cwd` key with `undefined` value, which overwrites `workingDirectory` in `Object.assign` thus causing the issue.

We can fix this in a number of ways:
Setting the `workingDirectory` value in the initial ui state
Filtering `undefined` values before `Object.assign`
Emitting a `SESSION_SET_CWD` action before the session creation etc.

This was one issue which was apparent and I was able to dig into, if I find any more, I'll add it here or create separate issues.

LabhanshAgrawal:
Closes #4295

### DIFF
--- a/app/ui/window.ts
+++ b/app/ui/window.ts
@@ -112,6 +112,10 @@ export function newWindow(
 
   function createSession(extraOptions: any = {}) {
     const uid = uuidv4();
+    const extraOptionsFiltered: any = {};
+    Object.keys(extraOptions).forEach(key => {
+      if (extraOptions[key] !== undefined) extraOptionsFiltered[key] = extraOptions[key];
+    });
 
     // remove the rows and cols, the wrong value of them will break layout when init create
     const defaultOptions = Object.assign(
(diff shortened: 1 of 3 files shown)


Stanzilla:
lgtm!

PR #4359, opened 2020-03-08
use electron notification api

LabhanshAgrawal:
Using the Electron [Notification](https://www.electronjs.org/docs/api/notification) API

removes
https://github.com/zeit/hyper/blob/a6ba442c1d3f84096df0f978ebabb61f9096a7ba/app/notify.ts#L7
😜

### DIFF
--- a/app/notify.html
+++ b/app/notify.html
@@ -1,5 +1,0 @@
-<script>
-  require('electron').ipcRenderer.on('notification', (ev, { title, body }) => {
-    new Notification(title, { body });
-  });
-</script>
--- a/app/notify.ts
+++ b/app/notify.ts
@@ -1,44 +1,10 @@
-import {resolve} from 'path';
-import {app, BrowserWindow} from 'electron';
-import isDev from 'electron-is-dev';
+import {Notification} from 'electron';
+import {icon} from './config/paths';
 
-let win: BrowserWindow;
-
-// the hack of all hacks
-// electron doesn't have a built in notification thing,
-// so we launch a window on which we can use the
-// HTML5 `Notification` API :'(
-
-let buffer: string[][] = [];
-
-function notify(title: string, body = '', details: any = {}) {
+export default function notify(title: string, body = '', details: any = {}) {
   console.log(`[Notification] ${title}: ${body}`);
   if (details.error) {
     console.error(details.error);
   }
-  if (win) {
-    win.webContents.send('notification', {title, body});
-  } else {
-    buffer.push([title, body]);
-  }
+  new Notification({title, body, ...(process.platform === 'linux' && {icon})}).show();
 }
-
-app.on('ready', () => {
-  const win_ = new BrowserWindow({
-    show: false,
-    webPreferences: {
-      nodeIntegration: true
-    }
-  });
-  const url = `file://${resolve(isDev ? __dirname : app.getAppPath(), 'notify.html')}`;
-  win_.loadURL(url);
-  win_.webContents.on('dom-ready', () => {
-    win = win_;
-    buffer.forEach(([title, body]) => {
-      notify(title, body);
-    });
-    buffer = [];
-  });
-});
-
-export default notify;


LabhanshAgrawal:
I tested it with mac and windows
Can someone please check it once with linux also.

bet4it:
When will the notification happen?

LabhanshAgrawal:
changing the config or even just saving it again should trigger a notification.

bet4it:
Before:
before
After:
after

LabhanshAgrawal:
I see, it was same for mac and windows, but that might be because these two os add the icon to the notification themselves.
Can you try this branch https://github.com/LabhanshAgrawal/hyper/tree/notif2
If that looks ok, then we can add a process.platform check to add icon for linux (on mac it gives two icons if enabled)

bet4it:
Yes, that works.

LabhanshAgrawal:
It looks the same? because on mac the icon from this comes on the right.

bet4it:
Yes, the same as before.

LabhanshAgrawal:
Nice, thanks for testing it, will update it.

bet4it:
Not totally the same, in fact😅
notif

LabhanshAgrawal:
Did you have some custom icon perhaps? The hyper icon should be the round(er) one.

bet4it:
It seems that the icon before was added by Linux, which was specified in [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=hyper-git).
After your commit the icon was obtained from itself.

LabhanshAgrawal:
Seems like it's older icon taken from [here](https://github.com/zeit/art/blob/master/hyper/mark/Hyper-Mark-120%403x.png)

LabhanshAgrawal:
Shouldn't be an issue. I've pushed again with the changes.
How can we get the icon in AUR updated?

bet4it:
Why @zeit not update their https://github.com/zeit/art repository😅

Stanzilla:
@bet4it they don't care about Hyper :D

ppot:
Lord. Nice cleanup.

bet4it:
Just find that 60883049c536348061d6dd51c59cb39fd8dffbaa could work on Linux. Linux will add the (old) icon to the notification automatically. The icon disappeared before because I run it by `npm run app`. 

You could remove the check for Linux for a cleaner code.

LabhanshAgrawal:
@bet4it  After `yarn run dist` do you get new or old or both icons?

bet4it:
I just realize it may be the effect of the global icon installed by package manager, so I uninstall hyper in package manager and use the AppImage built by Github actions.
60883049c536348061d6dd51c59cb39fd8dffbaa~ : no icon
60883049c536348061d6dd51c59cb39fd8dffbaa : no icon
dbed5bac9b4beb8ee208b23a5e5f2ab862c9a95e : new icon

When the hyper (packaged with old icon) is installed by package manager and use the AppImage, I get:
60883049c536348061d6dd51c59cb39fd8dffbaa~ : old icon
60883049c536348061d6dd51c59cb39fd8dffbaa : old icon
dbed5bac9b4beb8ee208b23a5e5f2ab862c9a95e : new icon

Now it's very clear. 60883049c536348061d6dd51c59cb39fd8dffbaa behave the same way as before, and dbed5bac9b4beb8ee208b23a5e5f2ab862c9a95e will always add the new icon in notification.
Both icons will never happen.

LabhanshAgrawal:
Like you said https://github.com/zeit/hyper/commit/60883049c536348061d6dd51c59cb39fd8dffbaa doesn't have icon in the first scenario and https://github.com/zeit/hyper/commit/dbed5bac9b4beb8ee208b23a5e5f2ab862c9a95e always has the new icon, I think it's ok to leave it like that.

P.S. Both icons happen on macOS if icon in electron notification is enabled as the os also adds an icon.

Issue #4379, opened 2020-03-14
Close icon doesn't appear when you close a tab


simonxabris:
- [x ] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x ] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: macOS 10.14.6
- **Hyper.app version**: 3.0.2 stable
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/simonxabris/07326acefabdd5718900eded8b84e770
- **Relevant information from devtools** _(CMD+ALT+I on macOS, CTRL+SHIFT+I elsewhere)_: N/A
- **The issue is reproducible in vanilla Hyper.app**:  Is Vanilla

## Issue

The issue is that when you have multiple tabs open and you close one, the tabs shift into place and on the tab that "shifts under your mouse" the close icon does not appear. It only appears when the you move the mouse and leave the tab and then hover over it again.

I've tracked down where the bug is in the code. https://github.com/zeit/hyper/blob/5bb0b37682ef866ef54e46834f4993408df17eec/lib/components/tab.tsx#L48

This mouse enter event does not get triggered when you close a tab and the one on its right shifts under the mouse, therefore the tab does not 'know' that it is hovered over and the close icon does not appear.

I've fixed the bug by replacing the js hover tracking with a simple css `:hover` selector and it works well, will be submitting a PR shortly, but I thought I'd create an issue first just so it's clear what the PR does.

I've also made a GIF of the bug to make it clear what I'm trying to say.

![hyper_close_bug](https://user-images.githubusercontent.com/27497229/76677757-45fd7f00-65d2-11ea-8f92-066e853cf0a7.gif)

PR #4380, opened 2020-03-14
remove js tab hover tracking and do it from css

simonxabris:
#4379 this is the issue this PR fixes.

I think this is ready to be merged, does not depend on any other work.

If there's any more work to be done, let me know and I'll fix it!

### DIFF
--- a/lib/components/tab.tsx
+++ b/lib/components/tab.tsx
@@ -1,26 +1,10 @@
 import React from 'react';
 import {TabProps} from '../hyper';
 
-export default class Tab extends React.PureComponent<TabProps, {hovered: boolean}> {
+export default class Tab extends React.PureComponent<TabProps> {
   constructor(props: TabProps) {
     super(props);
-
-    this.state = {
-      hovered: false
-    };
   }
-
-  handleHover = () => {
-    this.setState({
-      hovered: true
-    });
-  };
-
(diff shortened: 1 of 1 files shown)


simonxabris:
I think appveyor is only failing because it was disabled by this commit: https://github.com/zeit/hyper/commit/cc93bdae501263176bc96e0ed065fa5aaee13631

lib/components/tab.tsx
```diff
@@ -88,6 +69,12 @@ export default class Tab extends React.PureComponent<TabProps, {hovered: boolean
             color: #ccc;
           }
 
+          .tab_tab:hover .tab_icon {
+            opacity: 1;
+            transform: none;
+            pointer-events: all;
+          }
+
```
LabhanshAgrawal: I think it'd be better to move this block below (just above `.tab_shape`) to it's previous location, as all the tab_icon css is there, and would better reflect the fact that the main change is from `.tab_iconHovered` to `.tab_tab:hover .tab_icon` and not the properties

simonxabris: I did it this way so all the `.tab` hover states are grouped together, but what you're suggesting also seems reasonable to me. I'll push the changes in a couple of minutes.

Stanzilla:
Thank you!

PR #4399, opened 2020-03-18
port remaining js components to ts

LabhanshAgrawal:
With this the renderer is now fully written in ts. 🎉


Stanzilla:
Time to celebrate!

PR #4407, opened 2020-03-22
add command handler to uncatch keyboard event if search is closed

### LINKED ISSUE #3929: 5bc8e0b (on canary) breaks <esc> key
- [ ] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version

- **OS version and name**: Linux
- **Hyper.app version**: 6f96d9f (latest commit on canary branch)
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: N/A

## Issue
Commit 5bc8e0b introduces in each of the `app/keymaps` files the properties `editor:search` and `editor:search-close`. The value for the latter, `"esc"`, is unfortunate because it's global, is an extremely relevant key for `vim` users, and also factors into terminal shortcuts (at least when using `bash`).

Might it be possible to enable `editor:search-close` only when actually searching? This would keep it from interfering when not using the search feature. Alternatively, might a different key sequence be used (such as `shift+esc`)?

LabhanshAgrawal:
Fixes #3929 
Added a check to see if searchbox is open before dispatching the `SESSION_SEARCH_CLOSE` action and set `catched` to false if it isn't so that xterm is able to receive it.
Doing it by using a command handler, had to provide `dispatch` to command handlers for this to work.

Do test it out to see if it's working properly, and not impacting any plugins.

### DIFF
--- a/lib/actions/sessions.ts
+++ b/lib/actions/sessions.ts
@@ -145,13 +145,19 @@ export function onSearch(uid?: string) {
   };
 }
 
-export function closeSearch(uid?: string) {
+export function closeSearch(uid?: string, keyEvent?: any) {
   return (dispatch: HyperDispatch, getState: () => HyperState) => {
     const targetUid = uid || getState().sessions.activeUid!;
-    dispatch({
-      type: SESSION_SEARCH_CLOSE,
-      uid: targetUid
-    });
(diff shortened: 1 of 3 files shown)


Stanzilla:
Worked well here, thank you!

PR #4509, opened 2020-04-24
Use pseudo element to highlight active tab

eertmanhidde:
Just installed this great terminal few minutes ago but I was having a hard time seeing which term was actually being focused when having multiple terms open. So I added some overlay to non-active terms to highlight the active term.

Example of the result: https://gyazo.com/8823a13e8707bbcebc330b90953b46a0

### Improvements

* Don't use hardcoded white overlay color but a config color
* Add option to enable or disable this overlay in the config

### DIFF
--- a/lib/components/term.tsx
+++ b/lib/components/term.tsx
@@ -423,9 +423,26 @@ export default class Term extends React.PureComponent<TermProps> {
 
         <style jsx global>{`
           .term_fit {
+            position: relative;
             display: block;
             width: 100%;
             height: 100%;
+          }
+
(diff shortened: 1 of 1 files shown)


LabhanshAgrawal:
Shouldn't this be possible to do from the 'css' field in .hyper.js
Or maybe through a plugin

bet4it:
Many themes have implemented this. You could try `hyper-snazzy`.

eertmanhidde:
Closed since made this PR too quickly before checking for other plugins that have my fix already in them.

PR #4527, opened 2020-04-29
Show more plugins when use command list-remote

bet4it:
There are so many plugins on npm and many of them don't use `hyper-` as prefix.
Formatted by eslint.

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -13,8 +13,6 @@ import got from 'got';
 import ora from 'ora';
 import * as api from './api';
 
-const PLUGIN_PREFIX = 'hyper-';
-
 let commandPromise: Promise<void>;
 
 const assertPluginName = (pluginName: string) => {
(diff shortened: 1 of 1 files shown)


LabhanshAgrawal:
the prefix things seems good, but isn't 250 a bit too many, I think 50 should be good.

bet4it:
So which of the plugins are displayed? It's unfair for most other plugins!
And 250 is not such big. Many shell commands have a output of hundreds of lines, such as `ps aux`.

LabhanshAgrawal:
I guess npm would be sending the top results ordered by popularity or optimal.

Would it make sense to show a link at the end? Which will open npm search with the keywords.

Stanzilla:
Personally not a fan of listing them in CLI at all, so would prefer a link as well.

bet4it:
But that's not the origin intention of this command. If we show a link or just open it, why should we request the API?

What I do is just some improvements of this command. I don't want to change the ways.

LabhanshAgrawal:
I don't think we should remove the list totally, I just wanted to have a list of let's say 50 plugins and then say 'to see more visit [link]'

Stanzilla:
Yup

bet4it:
So we show 50 plugins in the terminal and when people visit link they must meet these 50 plugins again?

I really can't understand why you think a terminal output with 250 lines ( or 500 if you count the empty lines ) is unacceptable. When I use many commands like `pacman -Ss plugin`, I will get longer result, and we have many tools to help us deal with huge outputs in terminal, like `less` or `grep`.
People use command `hyper list-remote` may just want to use `grep` to find the exact name of a plugin, such as `hyper list-remote | grep summon`.

LabhanshAgrawal:
Well, I guess your point about the number of lines is ok. I wanted to include the link as when installing a plugin I usually look at the npm page and/or the github repo so usually end up searching for plugins in npm or awesome-hyper repo.
So I would still suggest to include the link after the 250 results.

bet4it:
There is another command `hyper docs hyperpower` which will open the npm page of a plugin.

LabhanshAgrawal:
Oh, didn't know about that, lgtm then

Issue #4193, opened 2020-01-21
Flag to disable hyperlinks


Garethgr:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: MacOS High Sierra
- **Hyper.app version**: 3.0.2

## Issue
I am quite new to Hyper (installed it today)
I would like to know if there is a flag to disable hyperlinks found in the terminal.
When I try select them I click them instead and it ends up opening many tabs in my browser.
I see that `hyperlinks` was available for a previous version (this no longer works)

LabhanshAgrawal:
AFAIK there is no config option to disable it.

Issue #3426, opened 2019-01-21
How to disable link click action


jstangroome:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: Windows 10 Pro x64 1809
- **Hyper.app version**: 2.1.1

## Issue

When clicking in the terminal window to give focus, or to select text for copying, the click often lands on text that Hyper considers to be a URL and then Hyper opens the URL in my default browser which is unwanted in this scenario.

It appears that pre Hyper 2.x the URL click handler was a separate extension and supported the click being disabled or requiring a modified key to be held while clicking to trigger the navigation:
https://github.com/zeit/hyperlinks#change-click-action

Now that this URL click handling is built into Hyper 2.x, how can I disable the behaviour, or require a modifier key?

eromoe:
Today is my first try to hyper , this really annoy me !  I want a simple way to disable this .

gujunhk:
> Today is my first try to hyper , this really annoy me ! I need a simple way to disable this . Any workaround for now ?

The issue is watched by @ppot .
However, it hasn't be planned.

anthonylauzon:
in ui/window.js:

`
    // xterm makes link clickable
    window.webContents.on('new-window', (event, url) => {
      const protocol = typeof url === 'string' && parseUrl(url).protocol;
      if (protocol === 'http:' || protocol === 'https:') {
        event.preventDefault();
        //shell.openExternal(url); <-----the devil
      }
    });
`

anthonylauzon:
in ui/window.js:
```
    // xterm makes link clickable
    window.webContents.on('new-window', (event, url) => {
      const protocol = typeof url === 'string' && parseUrl(url).protocol;
      if (protocol === 'http:' || protocol === 'https:') {
        event.preventDefault();
        //shell.openExternal(url); <-- comment the devil here
      }
    });
```

droope:
I too would like this behaviour to be disabled or at least configurable. I don't know how to apply @anthonylauzon's patch without having to rebuild the terminal, which I can't do right now :( may try later if I have some time.

abroglesc:
Configurable would be ideal. Coming from iTerm2 they had it to where if I held command, then clicked a link it would open in a browser. Allowing to disable this functionality or add in a modifier like holding command + click link would be amazing!

[...]

PR #4543, opened 2020-05-06
Add web-link activation modifier

rocket-squirrel:
Resolves zeit/hyper#3426, resolves zeit/hyper#4193

What is this PR going to fix:
- Hyper eagerly opens any link we click on, often unintentionally, which can be very distracting.
- It is currently harder than it should be to select a web link (because Hyper opens it in the browser instead of triggering selection).

How is it going to do it:
- add new entry to hyper config - webLinksActivationKey - to configure which key modifier needs to be pressed in order for a web-link to be activated on click. 4 values (alt, ctrl, meta, shift) are prefixes of corresponding properties of [MouseEvent](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent) passed to web-links add-on click handler.
- override WebLinksAddon click handler to check if required modifier is pressed.
- additionally, provide ILinkMatcherOptions.willLinkActivate callback to fix a problem with link selection.

There's a couple things though:
- Adding new config entry in 9 different places seems awkward, maybe there's a better way?
- Hyper does not re-initialize xTerm instance on config reload. Because of that, if config entry is updated it is not picked up on the fly (requires restart).
- Some of the MouseEvent modifier keys may be used in different shortcut combinations e.g. Ctrl+Click'ing the link won't work, but I guess this may be left at user discretion. Shift and Cmd (meta) works on Mac :)

### DIFF
--- a/app/config/config-default.js
+++ b/app/config/config-default.js
@@ -154,6 +154,10 @@ module.exports = {
     // rendering (slower, but supports transparent backgrounds)
     webGLRenderer: true,
 
+    // keypress required for weblink activation: [ctrl|alt|meta|shift]
+    // todo: does not pick up config changes automatically, need to restart terminal :/
+    webLinksActivationKey: '',
+
     // if `true` (without backticks and without quotes), Hyper will ignore ligatures provided by some fonts
     disableLigatures: false,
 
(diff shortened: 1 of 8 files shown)


Stanzilla:
I like it, what do you think @LabhanshAgrawal?

ETLaurent:
`webLinksActivationKey` 👌 🥲 
How am I seeing this option only now?...
Thank you 🙏 😌

Issue #4360, opened 2020-03-09
Text disappearing when using SF Mono as the font family


brunocloutier:
- **OS version and name**: Mac OS Mojave (10.14.6)
- **Hyper.app version**: 3.1.0-canary.4
- **The issue is reproducible in vanilla Hyper.app**: No

I have been using SF Mono as my terminal font for years, and last week, whenever I started typing in Hyper, all the text would disappear. The terminal would load with the first line shown (computer name / user name), but after the first letter was typed, every thing would disappear.

After messing around in the config, disabling plugins and options, I have identified the issue being with the use of SF Mono. Reverting it back to Menlo fixed the issue. I have tested with other fonts, and the issue seams to be with SF Mono only. 

I am using the Canary version of Hyper (3.1.0), and the issue is not reproductible in vanilla (3.0.4) where SF Mono works fine.

LabhanshAgrawal:
Tried it out, but couldn't reproduce the issue. Maybe try to reinstall the font and Hyper.

LabhanshAgrawal:
@brunocloutier did you try to reinstall?

JessicaSachs:
Just reproduced this from a fresh install. It's happening 100% of the time for me -- same build -- using the following font family string: `fontFamily: '"MesloLGS NF", Menlo, "DejaVu Sans Mono", Consolas, "Lucida Console", monospace'`

There's an exception being thrown where you're looking for /Network/Library/Fonts and that directory doesn't exist.

The fonts are loaded properly, then the exception is thrown and no fonts are rendered.

![Screen Shot 2020-03-27 at 8 44 36 PM](https://user-images.githubusercontent.com/2801156/77810633-d1afea80-706b-11ea-8fd6-910405e38ec3.png)

![Screen Shot 2020-03-27 at 8 44 06 PM](https://user-images.githubusercontent.com/2801156/77810630-d07ebd80-706b-11ea-9015-8454964c71dc.png)
![Screen Shot 2020-03-27 at 8 44 30 PM](https://user-images.githubusercontent.com/2801156/77810632-d1afea80-706b-11ea-9ef4-de1317a01a55.png)

You can't easily create `/Network/Fonts` because Catalina is doing crazy "read-only" filesystem stuff. ([link](https://stackoverflow.com/a/59421615))

LabhanshAgrawal:
@brunocloutier @JessicaSachs 
Can you try once with ligatures disabled, set `disableLigatures: true` in config.
[...]

Issue #3391, opened 2019-01-05
Text rendering overlap on scroll in a terminal editor


pariola:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version

- **OS version and name**: Ubuntu 18.04.1 LTS 
- **Hyper.app version**: 2.1.0

## Issue

I noticed if i use a terminal editor like nano, when i scroll down and happen to scroll back up the texts overlap

**Before Scroll**
![before](https://user-images.githubusercontent.com/11618490/50730077-0f61e400-1145-11e9-8d89-0340ebc928c3.png)

**After Scroll**
![after](https://user-images.githubusercontent.com/11618490/50730088-4932ea80-1145-11e9-837e-0b741b65b4f5.png)

j-f1:
As a workaround, you can hit `^L` to tell Nano to redraw the entire display.

stu177:
Getting the same on Ubuntu 18.04.1 and Hyper 3.0.0-canary.4

The same behaviour occurs in htop when data is refreshing.

connorbrez:
Also can reproduce this

meandmymind:
Same issue.
Hyper 2.1.2 (stable) with no extensions
Ubuntu 18.04 + KDE

ano95:
Can confirm. Linux Mint 19, hyper 2.1.2.

voquangphu:
Same here, happens every time I use nano.
MacOS Mojave, version 2.1.2 (stable).

b3njaminall3n:
Still an issue, happens when paging up/down in nano. Repainting with Ctrl+L fixes the screen until you scroll again, as @j-f1 mentions.
Hyper.app 3.0.0, no plugins, macOS 10.14.4

PandawanFr:
Happens for me too on macOS 10.14.5 with Hyper v3.0.2 (stable).

Many resources online said that you should set the `TERM` env variable. The default for Terminal.app is `xterm-256color` which works fine when using it. But when I set it in Hyper.app (both through the `.hyper.js` config and as an export in the `.bashrc`), it still breaks. 

I've seen multiple issues on GitHub that mention this including #3629, #3491, #3142, #3664, #3596. If there are that many people mentioning this issue, I think that it's a pretty important bug. 

Note: This is not a bug specific to nano since other issues (#3629) seem to be mentioning vim.

stu177:
[...]

PR #4684, opened 2020-06-26
disable ligatures in default config

LabhanshAgrawal:
There have been few issues with font rendering due to a small bug in https://github.com/princjef/font-ligatures which is a dependency of xterm-addon-ligatures.
Ref - #4360 #3391 ..

I've opened a pr https://github.com/princjef/font-ligatures/pull/18 for it and an issue in xterm https://github.com/xtermjs/xterm.js/issues/2980

Since not everyone will face this issue, it's better to have ligatures disabled in the default config, than removing it from the code. So that if someone wants to try it out, they can enable it.
We can have ligatures enabled by default again when the issue above is closed.

### DIFF
--- a/app/config/config-default.js
+++ b/app/config/config-default.js
@@ -158,8 +158,8 @@ module.exports = {
     // todo: does not pick up config changes automatically, need to restart terminal :/
     webLinksActivationKey: '',
 
-    // if `true` (without backticks and without quotes), Hyper will ignore ligatures provided by some fonts
-    disableLigatures: false,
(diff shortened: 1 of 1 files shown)


LabhanshAgrawal:
@Stanzilla should we do something similar for webgl? I recently tested it and it seemed to work ok. We can set the default config for `webGLRenderer` to `false` so that if someone wants to enable it they would be able to do it, instead of having to build again after enabling it in the code.

Stanzilla:
Yes I think so

LabhanshAgrawal:
#4686

PR #4723, opened 2020-07-08
Change project goal, prioritize safety

fulldecent:
I am proposing a change to the project goal.

# Summary

Add "safe" as the most important product feature.

# Discussion

## Background: why do all other projects support convenience over safety/security?

Here is the most beautiful and most extensible experience for installing anything:

```sh
curl fab.io | sudo bash
```

In fact nearly all software is installed this way. How can I be so sure? Because whereas only a mere billions of operating systems are installed by humans, trillions of installations are performed by computers via continuous integration. Safety is not a concern for CI systems.

## Main discussion point

It should be clear to everybody reading this by now that safety (i.e. security) and convenience (i.e. extensibility) can never both be maximized, one must take precedence over the other.

Hyper is designed to be run by humans (mostly programmers) and to have root access to development and production systems. [These](https://theintercept.com/2014/03/20/inside-nsa-secret-efforts-hunt-hack-system-administrators/) humans [are](https://www.dailymail.co.uk/news/article-2621253/Foreign-intelligence-agencies-targeting-IT-workers-big-businesses-gain-access-sensitive-information-MI5-warns-British-corporations.html) specifically [targeted](https://www.theregister.com/2014/05/07/mi5_warning_foreign_spies_targeting_it_workers/) by [hackers](https://cryptoplanet.cash/north-korean-hackers-are-targeting-the-sysadmins-of-crypto-exchanges/) with lots of resources. So we should explicitly target safety as the top goal and build safety into all decisions, including how modularity and third-party code works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ For more details, head to: https://hyper.is
 
 ## Project goals
 
-The goal of the project is to create a beautiful and extensible experience for command-line interface users, built on open web standards. In the beginning, our focus will be primarily around speed, stability and the development of the correct API for extension authors.
+The goal of the project is to create a safe, beautiful and extensible experience for command-line interface users, built on open web standards. In the beginning, our focus will be primarily around speed, stability and the development of the correct API for extension authors.
 
 In the future, we anticipate the community will come up with innovative additions to enhance what could be the simplest, most powerful and well-tested interface for productivity.
 


LabhanshAgrawal:
Hyper's main feature is to be extensible and customisable which in my opinion is why most people use it.
By default we don't interact with user input or output much.
Plugins are basically node modules & some of which interact with the terminal to provide things like suggestions etc. and thus are to be used on user discretion.
I'd be happy to see pr's which patch any existing issue without affecting the extensibility and customisability.

fulldecent:
Understood. It seems like Hyper is yet another project that encourages people to run untested third-party code on your computer with shell access. In that case this is not a project I'm interested to be a part of.

PR #4889, opened 2020-09-15
Move electron type augmentation to separate dts file

LabhanshAgrawal:
fix electron typings hack
P.S it was still not working in `ext-modules.dts` but this seems good too.


LabhanshAgrawal:
There was some problem with the dts output during build, which seems to be fixed by updating typescript.
The mac ci seems to be failing because of some network problems, which appears to be quite frequent (also saw in other pr's)

PR #4891, opened 2020-09-16
Fix eslint warnings

LabhanshAgrawal:
updated @typescript-eslint packages to fix the warnings which were reported incorrectly.


Stanzilla:
ah perfect, thank you

PR #4913, opened 2020-09-24
fix: add a 256x256 for AppImage builder to add as .DirIcon

srevinsaju:
.Diricon generally provides the icon to the AppImage in the
common Linux file manager. As specified in the AppDir spec
https://docs.appimage.org/reference/appdir.html, a minimal AppDir
should contain a (a) AppRun (b) .DirIcon. The .DirIcon should be of
the standard size 128x128 or 256x256

Hyper did not include a .DirIcon which fails to pass AppImage tests
and Desktop integration. Adding a 256x256 icon.png to the build directory
enables appimage-builder to automatically set the .DirIcon of the AppImage
without extra configuration

> INFO:
* The PR is ready to merge

Fixes https://github.com/AppImage/appimage.github.io/pull/2349


srevinsaju:
Looks redundant

PR #5183, opened 2020-12-24
Fix installation instructions in README.md

kawarimidoll:
Fix error when installing using latest Homebrew in macOS. Just fix README.md.

Calling brew cask install is disabled


kawarimidoll:
Thank you!

PR #5244, opened 2021-01-20
Change windows build target to nsis

LabhanshAgrawal:
Build has been failing for few days because of `squirrel.windows`
I've opened an issue electron-userland/electron-builder#5551 but thought that it's about time we start using the NSIS installer.

I've added nsis script to add hyper to system context menu on install and remove on uninstall.
Also tested that it doesn't cause any unicode issues while accessing the registry.

This has made the `app/system-context-menu.ts` file unused for now.
We can delete it or use it to add an item in the plugins menu as `add hyper to system context menu` like the `install hyper command to path` menu item.

Since the nsis install path is `%LOCALAPPDATA\Programs\Hyper` instead of squirrel's `%LOCALAPPDATA\hyper\app-<version>` I've changed the cli install code to remove old paths.

Added Bonus - This has resulted in ~30% reduction in Setup file size.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -152,17 +152,10 @@ jobs:
       - name: Get Windows Artifact Names
         id: getwindowsfilename
         run: |
-          Write-Host "::set-output name=exeName::$(Get-ChildItem -Name .\dist\squirrel-windows\ | Select-String exe)"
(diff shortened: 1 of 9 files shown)


LabhanshAgrawal:
@Stanzilla what do you think about the `system-contex-menu.ts` thing?

Stanzilla:
I like the item in the plugins menu idea, some people might find that useful

PR #5349, opened 2021-02-23
change aurman to paru

lollilol:
its rust

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -157,7 +157,7 @@ args.command(
   (name, args_) => {
     const pluginName = args_[0];
     assertPluginName(pluginName);
-    open(`http://ghub.io/${pluginName}`, {wait: false, url: true});
+    open(`http://ghub.io/${pluginName}`, {wait: false});
     process.exit(0);
   },
   ['d', 'h', 'home']


LabhanshAgrawal:
Does it make any difference?
People can use whichever package manager they want.

Stanzilla:
Yeah that's not a reason to switch it out but you can add it as an additional suggestion

Stanzilla (changes requested):
see comment

lollilol:
> Does it make any difference?
> People can use whichever package manager they want.

yes people can use whatever they want. No one said they cant lol
Also yay is basically the de-facto standard when it comes to aur helper and paru is from the same devs but in rust. Its basically the best one and people can still choose what to use lmao its just the example

It doesnt make a difference per-se but paru has cool stuff like memory safety because of rust. For example 70% of all high severity security bugs in google chrome are related to memory safety. It can solve so many issues.

And yea you can reject this of course, but its objectively seen the best aur helper that exists.

LabhanshAgrawal:
I mean, people will use whatever aur helper they are already using, regardless of what we put in the example. But I understand the example can also get people to check it out.
If there's any list of aur helpers, you can edit the pr to say
```
Use an AUR [package manager](link to some list of aur helpers) eg. paru
```

LabhanshAgrawal:
was adding the requested changes and messed it up somehow
have opened #5395 for this

PR #5412, opened 2021-03-13
Update macOS icons to match the new Big Sur icon shape

leodr:
Hey, I updated the Hyper macOS icons to match the [new shape of Big Sur icons](https://developer.apple.com/design/human-interface-guidelines/macos/icons-and-images/app-icon).

Here is a comparison of the new icons that this PR would introduce on the left, vs. the current icon on the right.

| Regular | Canary |
| --- | --- |
| Bildschirmfoto 2021-03-13 um 11 08 00 | Bildschirmfoto 2021-03-13 um 11 14 42 |

As you can see, the current icon is just slightly too big and does not have the new slight drop shadow.

The Figma file for these new icons can be found [here](https://www.figma.com/file/iQXDMgBn0kMORpg81VS857/Hyper-Big-Sur-Icon?node-id=0%3A1). To match the shape and the drop-shadow I used [this](https://www.figma.com/community/file/857303226040719059) Figma template.


leodr:
CC @evilrabbit (I think you made the logo), if you can spare some time 😃

LabhanshAgrawal:
Is the only change in the size? or the shape also.
Also, there are some gridlines in the regular icon currently.
And I think we should also include the figma file here (if possible).

leodr:
> Is the only change in the size? or the shape also.
> Also, there are some gridlines in the regular icon currently.
> And I think we should also include the figma file here (if possible).

Oh yeah sorry, you are right. I'll remove the grid lines.

Do you mean committing the Figma file to the repo? Where should I put it, just next to the icons?

LabhanshAgrawal:
Is the standard shape not a square? If yes then maybe that's ok.
I'm guessing that you've verified the colors and the gradient.
From the overlap, it seems that in your design the border is thicker in general and especially on the top and bottom, haven't measured it though. Can you make it same as before and also uniform all around?
The shape of the lightning is different (look at the middle \\ portion) and the ends are rounded instead of sharp. Also, the underscore is shorter.

You can refer to https://github.com/vercel/art/tree/master/hyper . I don't know for sure but maybe you can get some details and the shapes from the sketch file there (It's outdated I guess but might be useful)

I guess putting the figma file next to the icons would be ok.

Do you think we should generate windows icons also from this design? If the main issue is size then maybe not, as I don't think the icon is too big on windows. @Stanzilla what do you say?

Stanzilla:
Think we can use the same design on Windows, yeah.

leodr:
I fine-tuned the icons with the help of the current ones and they now match almost exactly:

Bildschirmfoto 2021-03-17 um 21 12 50

The ⚡  and the underline are now the same sizes as the current icons. I also updated the border width and removed the grid lines from the stable version.

I also added the icons as `.ico`-files for the Windows icons. The `icon.fig`-file is the exported Figma file.

LabhanshAgrawal:
@leodr can you please put a comparison of the canary icons also.
Not really familiar with editing etc. to compare it myself.

leodr:
Sure thing! Here's each icon overlayed with the current one in white:

Stable Big Sur-1024
Canary Big Sur-1024

leodr:
Here's a comparison of the icons on Windows, with the left being the original one:

image

Here's the taskbar:

image

I think it looks pretty small on Windows, do we rather want to stick with the old icon there or should I make it a little bit bigger for Windows?

ralphjsmit:
Any idea when the next official release comes out with the icon?

LabhanshAgrawal:
Yeah, it does seem a bit small on windows. Actually, even the current one is a bit small compared to other icons. Can a bigger icon be exported from the same figma file?

I was comparing the icons and the colors are a bit different and also the shadow(I guess).
Screenshot 2021-03-18 at 22 07 50

For example, I checked the color for the dark area in canary icon.

Current
Current

New
New


Also, in the windows icons, right now there are 5 sizes but in yours there are 4 sizes. I'm hoping there's some standard template for windows also from which we can get the recommended sizes. Can you please check.

Thanks

LabhanshAgrawal:
> Any idea when the next official release comes out with the icon?

Can't promise anything on the next release at the moment.

leodr:
I think the differences in color might have to do with the exporting process. I did sample the colors from the current icons, so they should be the same. I'll try exporting them as PNG and creating a `.icns`-file and `.ico`-file manually.

And for the differences in shadow, the shadow is the same now for every Big Sur app icon. Before the Hyper icon had no shadow as far as I know, so I gave it the default shadow that every other Big Sur app icon now has.

LabhanshAgrawal:
Do take a look at [this](https://drive.google.com/file/d/1gwUSXt7sN1VdScI7lriZws9Fj6rHT-Xf/view?usp=sharing)
it's the svg file for the icons (didn't get the sketch or figma) maybe you can get something from it.
From what I see, the background in the canary icon is solid color with a separate translucent gradient layer on top. Instead of gradient of light brown(ish) to dark.

LabhanshAgrawal:
One more small detail, the border looks thinner at the corners. Can you please check it once?

LabhanshAgrawal:
> Do take a look at [this](https://drive.google.com/file/d/1gwUSXt7sN1VdScI7lriZws9Fj6rHT-Xf/view?usp=sharing)
> it's the svg file for the icons (didn't get the sketch or figma) maybe you can get something from it.

got this file today itself, so couldn't provide it before.

leodr:
> Do take a look at [this](https://drive.google.com/file/d/1gwUSXt7sN1VdScI7lriZws9Fj6rHT-Xf/view?usp=sharing)
> it's the svg file for the icons (didn't get the sketch or figma) maybe you can get something from it.
> From what I see, the background in the canary icon is solid color with a separate translucent gradient layer on top. Instead of gradient of light brown(ish) to dark.

That's great, I think I can get some useful information out of these SVGs! I'll have a look at it again but can't make any promises on when I'll have time for it.

leodr:
Ok, I got it now. I took your SVG images and matched everything. I took the gradients, and you were right about the border being a little bit too thin around the corners. I matched that up with the SVG icons
Bildschirmfoto 2021-04-01 um 22 26 11

Haven't done anything for the Windows icons yet. I really have no clue what makes a good Windows icon, so maybe I can revert the commit for now and if somebody else wants to redo the Windows icons they can, the Figma file is there.

LabhanshAgrawal:
Looks awesome
Thanks for the great work
Just few questions
What's the reason for using 58% for corner smoothing. If you measured it from current then that's ok. Was just wondering as Figma showed 60% as default for iOS and I'm guessing that'd be the same for mac?
And how does one export the icons from the figma file?
Other than this I think it's ready to merge.

leodr:
I'm not quite sure what you mean with the 58% corner smoothing. Could you elaborate on that? I took the border radius of the icon from this [macOS Big Sur Icon Figma template](https://www.figma.com/community/file/857303226040719059).

Here are the steps to obtain the `.icns` files from the Figma file:

1. Select the layer you want to export
2. In the bottom right tap on the button that says "Export [layer name]"
3. Take the downloaded 1024x1024 PNG and run it through [this](https://github.com/jamf/icns-Creator) script. E.g. `bash icns_creator.sh Canary.png canary`

LabhanshAgrawal:
On Figma there's a slider for corner smoothing (https://help.figma.com/hc/en-us/articles/360050986854-Adjust-corner-radius-and-smoothing#Corner_smoothing_for_squircles)
It mentions 60% to be the default for iOS
I checked the template you've mentioned and the 58% value seems to come from there, but I couldn't find any resources on whether that's the value for macOS (also seemed strange to be slightly off from iOS)

leodr:
You're right, I didn't know about corner smoothing. I could not really see a difference between 58% and 60%, so maybe it doesn't really matter.

LabhanshAgrawal:
I guess we can go with the template for now, I've already nitpicked a lot 😇

Thanks a lot for your great work.

PR #5522, opened 2021-04-24
Fix logic in saveAsBackup

bet4it:
`e.code` is `null` after `fs-extra 8.0.0`


LabhanshAgrawal (approved):
lgtm

LabhanshAgrawal:
`e.code` might be `null`, it does throw though

https://github.com/jprichardson/node-fs-extra/blob/fb6c0ca16baec6a3f5138685e72cf3659c493017/lib/move-sync/move-sync.js#L32

but checking beforehand is better

PR #5632, opened 2021-06-05
Add Windows Vibrancy

TylerSeanRau:
Vibrancy isn't supported by electrons on any platform other than MacOS. I found some years old discussions about the feature being possible and also some comments asserting that changing the opacity is good enough (it's not :P).

In my pursuit for vibrancy, I discovered https://github.com/Seo-Rii/electron-acrylic-window which is a wrapper for electron's `BrowserWindow` that changes the vibrancy property to be of a more complex type than just strings and works only on Windows.

I'm certain there is a better way to integrate this but I've only hacked this together and this definitely needs some polish before being considered for merging.

Importantly for these changes to work, after `yarn` is run you'll need to modify `node_modules/electron-acrylic-window/build_ts/browserWindow.d.ts:67:5` to change:

```
    setVibrancy(options?: Vibrancy): void;
```

to

```
    setVibrancy(options?: Vibrancy | null): void;
```

This works for me locally as a hack to be able to use these changes, but I'm not sure the correct path forward (or if there is one that can be done without submitting a PR to 'electron-acrylic-window' also.)

Other changes that probably need to go into this before it can be merged:

* Vibrancy shouldn't be always enabled like it is in the original changes
* a better conditional for `needTransparency`

### DIFF
--- a/lib/components/term.tsx
+++ b/lib/components/term.tsx
@@ -37,7 +37,7 @@ const isWebgl2Supported = (() => {
 
 const getTermOptions = (props: TermProps): ITerminalOptions => {
   // Set a background color only if it is opaque
-  const needTransparency = Color(props.backgroundColor).alpha() < 1;
+  const needTransparency = true;
   const backgroundColor = needTransparency ? 'transparent' : props.backgroundColor;
 
   return {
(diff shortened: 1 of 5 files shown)


lib/components/term.tsx
```diff
@@ -37,7 +37,7 @@ const isWebgl2Supported = (() => {
 
 const getTermOptions = (props: TermProps): ITerminalOptions => {
   // Set a background color only if it is opaque
-  const needTransparency = Color(props.backgroundColor).alpha() < 1;
+  const needTransparency = true;
```
axieum, line 40: This will always evaluate to `'transparent'`, was this left-over from testing?

LabhanshAgrawal:
Thanks for the pr
will take some time to review as focusing on doing a new stable release first, also don't have windows at hand
will push some minor fixes though

PR #5805, opened 2021-07-31
Add disableAutoUpdates config option

letsandeepio:
This PR allows the user to disable auto-updates by setting the `disableAutoUpdates` flag to true.

### DIFF
--- a/app/auto-updater-linux.ts
+++ b/app/auto-updater-linux.ts
@@ -1,6 +1,5 @@
 import fetch from 'electron-fetch';
 import {EventEmitter} from 'events';
-import * as config from './config';
 
 class AutoUpdater extends EventEmitter implements Electron.AutoUpdater {
   updateURL!: string;
@@ -16,10 +15,6 @@ class AutoUpdater extends EventEmitter implements Electron.AutoUpdater {
   }
 
   checkForUpdates() {
-    const baseConfig = config.getConfig();
-
-    if (baseConfig['disableAutoUpdates']) return false;
-
     if (!this.updateURL) {
       return this.emitError('Update URL is not set');
     }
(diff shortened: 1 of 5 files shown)


app/config/config-default.js
```diff
@@ -173,6 +173,8 @@ module.exports = {
   // to load it and avoid it being `npm install`ed
   localPlugins: [],
 
+  disableAutoUpdates: false,
```
LabhanshAgrawal, line 176: put it inside config object

app/auto-updater-linux.ts
```diff
@@ -15,6 +16,10 @@ class AutoUpdater extends EventEmitter implements Electron.AutoUpdater {
   }
 
   checkForUpdates() {
+    const baseConfig = config.getConfig();
+
+    if (baseConfig['disableAutoUpdates']) return false;
+
```
LabhanshAgrawal, line 22: Make the changes in `updater.ts` instead

LabhanshAgrawal (changes requested):

LabhanshAgrawal:
Can you do a rebase and push?

app/config/config-default.js
LabhanshAgrawal: Add a one liner too

app/config/config-default.js
```diff
@@ -54,6 +54,9 @@ module.exports = {
     // custom CSS to embed in the main window
     css: '',
 
+    // set to true to disable auto updates
+    disableAutoUpdates: false,
+
```
LabhanshAgrawal, line 59: Put it at the end, after `disableLigatures` (Line 159 i guess)

LabhanshAgrawal (changes requested):

LabhanshAgrawal (approved):

PR #5818, opened 2021-08-02
Get dependabot to update github-actions

mmorel-35:
Dependabot is also able to handle github-actions.
If you are interested, I have added here, the same way you handle your npm dependencies.


LabhanshAgrawal:
👀 Oh, didn't know about this. Thanks.

PR #6007, opened 2021-10-10
Add instructions on how to install `hyper-deb` for Ubuntu and Debian using Pacstall

Henryws:
[Pacstall](https://github.com/pacstall/pacstall) is a community-driven AUR-like package manager for Ubuntu. I am the maintainer (and the main developer of Pacstall) of the [hyper-deb](https://github.com/pacstall/pacstall-programs/blob/master/packages/hyper-deb/hyper-deb.pacscript) pacscript (similar to a PKGBUILD) in this repository. With this script, people can install and update Hyper without having to go to the download's section on GitHub every update

It can be installed on Ubuntu and Debian by running this command (after Pacstall is installed)
```bash
pacstall -I hyper-deb
```

It can also be upgraded
```bash
pacstall -Up
```
and removed when requested.
```bash
pacstall -R hyper-deb
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,13 @@ Hyper is available as [Nix package](https://github.com/NixOS/nixpkgs/blob/master
 nix-env -i hyper
 ```
 
+#### Debian/Ubuntu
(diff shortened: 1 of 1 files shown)


LabhanshAgrawal:
Pacstall looks neat, but this would just cause confusion
The case is different with nixos, macos etc. as nix-env and brew respectively are the default (or almost default) way and commonly known.

PR #6042, opened 2021-10-22
Prompt before closing window if terminals have running children

theIDinside:
On Linux the user is prompted with a message box that
child processes are running and if they want to quit
and thus kill them.

Implemented for:
- [x] Linux
- [ ] Windows
- [ ] MacOS

I might be able to add Windows on this list, but MacOS users; well they're on their own. There is absolutely no way to do this in a cross platform manner. But it works on Linux. As such the pull requests should be accepted and issue updated with information that it is not implemented for the other platforms.

Additional features that could be implemented;
- When user clicks a tab to close; ask the same question as when the window is closed. This requires **a lot** more insight into the Hyper code base however, since when the user clicks the x (close button) on a tab, the tab immediately disappears from the UI *before* any "on-close" or "exit" handlers are ran, thus it's on the Electron side, so I suppose some middleware has to be set up, to first direct control back to the "app" code of Hyper, and once handled *then* send back to Electron to remove it from the UI. Or something like that.

The changes are only made to one file;  [index.ts](https://github.com/theIDinside/hyper/blob/85b394383e687eae854c47a3bb31df2c90de122e/app/index.ts#L141). A utility closure called ```sessionHasRunningChildren``` checks the Session object's pid, to look in the in-memory drive /proc/.../pid/... to see if Linux has any child info of that pid there and handles it accordingly. It's nowhere near as well designed as this on Windows; instead there we have to jump through a lot of hoops, one of which involves spawning a child process that executes PowerShell and runs some commands - or calling WinAPIs through FFI directly, which in and of itself is very unstable. Poor design, but what else is new on Windows?

I commented in the issue thread [here](https://github.com/vercel/hyper/issues/684) some questions, but I get that this feature isn't exactly at the top of the list.

Also, if and when the PR gets accepted could you be *so* kind to tag the PR as "hacktoberfest-accepted"? It would be very nice of you.

If you, @LabhanshAgrawal for instance, have some ideas on how to solve the tab thing, or where I could do what is needed, I'd gladly add that additional (though unrequested by the issue) feature, but filed under a separate issue if you would be so kind.

### DIFF
--- a/app/index.ts
+++ b/app/index.ts
@@ -13,7 +13,7 @@ if (['--help', '-v', '--version'].includes(process.argv[1])) {
 import {resolve} from 'path';
 
 // Packages
-import {app, BrowserWindow, Menu} from 'electron';
+import {app, BrowserWindow, dialog, Menu} from 'electron';
 import {gitDescribe} from 'git-describe';
 import isDev from 'electron-is-dev';
 import * as config from './config';
@@ -26,6 +26,8 @@ import {installCLI} from './utils/cli-install';
 import * as AppMenu from './menus/menu';
 import {newWindow} from './ui/window';
 import * as windowUtils from './utils/window-utils';
+import Session from './session';
+import {readFileSync} from 'fs';
 
 const windowSet = new Set<BrowserWindow>([]);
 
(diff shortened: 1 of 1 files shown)


Giffen-good:
Hey I could provide the code to run a check for Mac OS. May I checkout your PR and add the condition for MacOS within your ```sessionHasRunningChildren``` method?

app/index.ts
```diff
@@ -136,8 +138,39 @@ app.on('ready', () =>
...
-        hwin.on('close', () => {
+        hwin.on('close', (evt) => {
+          for (const s of hwin.sessions.values()) {
+            const session: Session = s;
+            if (sessionHasRunningChildren(session)) {
+              // todo: focus the window/pane that has a child running in the foreground?
+              const opt = dialog.showMessageBoxSync(hwin, {
+                title: 'Close window & all terminals?',
+                buttons: ['Close terminal', 'Cancel'],
+                message: `There is still a process running in a hyper terminal. Closing the terminal will kill it`
```
PythonCoderAS, line 163:
```suggestion
                message: `There is still a process running in a hyper terminal. Closing the terminal will kill it.`
```

PR #6538, opened 2022-05-16
Passing `cursorAccentColor` down properly

latin-1:
Currently, Hyper does not respect `cursorAccentColor` config.


LabhanshAgrawal:
makes sense
Can you also add screenshots that show difference between this and canary when you set some cursor accent color.

latin-1:
```js
// .hyper.js
module.exports = {
  config: {
    cursorAccentColor: '#fff',
  },
};
```

| Before | After |
| - | - |
| image | image |

LabhanshAgrawal:
Nice, thanks for the contribution.

Issue #5941, opened 2021-09-16
Consider using v8 snapshots to improve performance


LabhanshAgrawal:
https://blog.inkdrop.app/how-to-make-your-electron-app-launch-1000ms-faster-32ce1e0bb52c
https://github.com/inkdropapp/electron-v8snapshots-example

Issue #4560, opened 2020-05-17
expose xterm.js' `loadAddon` to plugins


rodolf0:
Feature request:
xterm.js is extensible via addons. Hyper currently loads a static set of addons (https://github.com/zeit/hyper/blob/bad51f9c15a9f094588a3c3e36bea2b5151d4606/lib/components/term.tsx#L154-L158). It would be useful to allow plugins to enable additional addons.

One of the main motivators behind this would be to allow for custom `ILinkProvider` to enhance user interaction with text objects. Some examples of link providers beyond the static `WebLinksAddon` could be matching emails or user provided regexes as requested in https://github.com/xtermjs/xterm.js/issues/2893.

PR #6987, opened 2023-01-16
Add support for displaying Sixels

### LINKED ISSUE #6748: Any way to display images inline in terminal?
Hi, is there any way to display images in terminal, i using zsh in my wsl with neofetch and w3m and i can't display images in terminal with hyper.

marvinruder:
Closes #6748 

For a quick demo, run `curl https://raw.githubusercontent.com/saitoha/libsixel/master/images/snake.six`


LabhanshAgrawal:
Thanks for your contribution, I'll test it out soon.
From what I read, the addon is in beta and does not work with the latest xterm version (we're not on latest right now, but I'm working on that currently). So, it might be better to make it as a plugin instead, will have to take a deeper look to decide.

webpack.config.ts
```diff
@@ -118,6 +118,7 @@ const config: webpack.Configuration[] = [
       mousetrap: 'require("./node_modules/mousetrap/mousetrap.js")',
       open: 'require("./node_modules/open/index.js")',
       'xterm-addon-fit': 'require("./node_modules/xterm-addon-fit/lib/xterm-addon-fit.js")',
+      'xterm-addon-image': 'require("./node_modules/xterm-addon-image/lib/xterm-addon-image.js")',
```
LabhanshAgrawal, line 121: You'll have to require it in https://github.com/vercel/hyper/blob/canary/bin/snapshot-libs.js as well to put it in v8 snapshot.

LabhanshAgrawal:
Oh, I saw https://github.com/jerch/xterm-addon-image/issues/38 , didn't know it's fixed.

marvinruder:
If a plug-in solution is favored, xterm.js’ `loadAddon` method would likely need to be exposed to plug-ins, as requested in #4560.

jerch:
> Oh, I saw [jerch/xterm-addon-image#38](https://github.com/jerch/xterm-addon-image/issues/38) , didn't know it's fixed.

Ah sorry, forgot to close the issue along with the fix.

lib/components/term.tsx
```diff
@@ -211,6 +212,7 @@ export default class Term extends React.PureComponent<
         )
       );
       this.term.open(this.termRef);
+      this.term.loadAddon(new ImageAddon('./node_modules/xterm-addon-image/lib/xterm-addon-image-worker.js'));
```
jerch, line 215: The worker abstraction is gone for now, as it created to many frictions on integrator side. So if you are fine with the addon's default options, thats only `new ImageAddon()` currently.

marvinruder: Do those newer versions of xterm-addon-image work with xterm v4 as well? I tried, but didn't get it to work. That is what Hyper is based on at the moment.

jerch: Ah nope, the older 4.x versions need older addon versions as well, still including the worker.

LabhanshAgrawal: @jerch thanks for reviewing this. I'm trying to upgrade to xterm v5, though there are some issues with ligatures on v5, for which I've opened few prs on xterm repo. I think we can hold off on image support till we can get to v5.

marvinruder: Understandable. Will follow the topic and keep this PR up-to-date with the canary branch and its xterm version. 

Do you plan on including xterm v5 directly in the v4.0.0 release or at a later time? Any estimated release date?

LabhanshAgrawal: In v4 itself, no fixed release date at this point, but working on few significant things and plan to release soon after.

LabhanshAgrawal: I'm waiting for https://github.com/xtermjs/xterm.js/pull/4386 and https://github.com/xtermjs/xterm.js/pull/4387 to be closed, then we can upgrade to xterm v5

jerch:
Minor headsup - just added iTerm IIP support in the addon, though I consider it alpha (cannot really test&compare against iTerm2).

LabhanshAgrawal:
Sorry, I was caught up with things, have to rethink the pr on xterm's repo for ligature fix as well, I'm thinking to merge this (with xterm v5) in the coming days, even if the ligature fix is not there yet, as anyways ligatures are disabled on electron in xterm v5 currently, so we can have few canary releases without ligatures, and then work on that instead of blocking this.
Thanks

webpack.config.ts
marvinruder: Putting `xterm-addon-image` in the v8 snapshot currently breaks the build, as it uses a reference to `self`, which does not appear to be available in that environment. Unfortunately I have no idea how to fix that. Is it required to put `xterm-addon-image` into the snapshot? I am not too familiar with electron and v8 snapshots, but from what I read in #5941, #6224, it is some form of caching and only reduces startup time, but is not necessary to run the application and use the dependency therein?

LabhanshAgrawal: yeah it's not necessary to put it in snapshot, I will test it out when I'll pick up the xterm v5 pr (hoping for this weekend).

jerch: Idk anything about that snapshot functionality, would that also deal/preserve wasm state? If it cannot deal with wasm the same way, then it will prolly not work after resume from snapshot?

The initial addon loading on calling the ctor should be okish - I tried to shape most of the heavier stuff to lazy initialization upon first usage (e.g. the base64 wasm decoder will spawn on first chunk to be decoded). There still might be room for improvement, just file me a note if you find unpleasant blocking stuff in startup profiling.

LabhanshAgrawal: i'm also not sure about the wasm stuff with regards to snapshots, it basically parses js and stores the memory state. I will test and see if it works.

LabhanshAgrawal:
I've upgraded xterm to v5, please rebase.

webpack.config.ts
marvinruder: Still getting the `self is not defined` error when putting `xterm-addon-image@0.4.1` into the snapshot with xterm 5.2. Everything works fine when using the snapshot from the `canary` branch though, this appears to be the only remaining issue.

jerch: @marvinruder Does it go away, if you replace `self` with `this` in the bundled JS file? And more important, does the addon still work with this change? (In theory there should not be any explicit references to the global context from within the code, but I have not checked it, if webpack would introduce them for some other reason...)

marvinruder: `this` does not throw an error, but the next issue is `atob is not defined`. Tried to replace it with `Buffer.from`, but then `Buffer is not defined`.

jerch: Oh thats weird, all `atob` calls should already have a guard for browser vs nodejs. Whats the actual global context in a snapshot - is this a nodejs context during snapshotting, that gets simply mapped into the browser later on? Because that the only way I can think of failing the guard.

The guard looks like this:
```typescript
function _dec(s: string): Uint8Array {
  if (typeof Buffer !== 'undefined') return Buffer.from(s, 'base64');
  const bs = atob(s);
  const r = new Uint8Array(bs.length);
  for (let i = 0; i < r.length; ++i) r[i] = bs.charCodeAt(i);
  return r;
}
```

jerch: The issue is created by `mk-snapshot.js` in line 28:
https://github.com/vercel/hyper/blob/11c923842a997ef14ca5ad3eac0ac4cd29cf189f/bin/mk-snapshot.js#L28

Idk what `result.snapshotScript` contains at that time (some electronLink output), but the issue actually might be, that `vm.runInNewContext` creates just a naked V8 context, and not a browser typical context? Thus missing `self` and `self.atob`?

Maybe its already enough to stub `self` and `atob` in that context to pass the test?

LabhanshAgrawal: The v8 snapshot is just parsing the js and storing it's in memory representation, to speed up initial start. Since it doesn't need to execute it fully while snapshot creation, it just should be in the bundle, it works if we just put it in the `if(false)` block in the snapshot script. Most of the xterm packages have similar issues, you can see almost all of them are in that block.

jerch: The following fixes `self` and `atob` not being defined:
```javascript
const vm = require('vm');
vm.runInNewContext('self.atob("AAAA")', {
    atob: s => Buffer.from(s, 'base64').toString('binary'),
    self: global
});
// --> '\x00\x00\x00'
```

LabhanshAgrawal:
great work @marvinruder, also thanks @jerch for the awesome addon and your support on this pr.
Everything's looking good on local testing, just waiting for the CI.
Screenshot 2023-06-16 at 17 35 04

webpack.config.ts
jerch: > Most of the xterm packages have similar issues, you can see almost all of them are in that block.

The errors for all those package are most likely also because of the empty global context here. Maybe its a good idea, to at least shim most often used browser API stuff in that test context?

LabhanshAgrawal: Will have to check how electron-link and mksnapshot are doing things to figure out where/how to modify the context. I'll take a look.

LabhanshAgrawal: Ah, somehow didn't register your earlier comment, will try that out.

LabhanshAgrawal: I tried adding the missing definitions to the context. The `vm.runInNewContext` runs but `mksnapshot` fails later on. Will need further research, I guess.

Issue #2836, opened 2018-04-16
Bold text doesn't respect colors in config


LinusU:
- [x] I am on the [latest](https://github.com/zeit/hyper/releases/latest) Hyper.app version
- [x] I have searched the [issues](https://github.com/zeit/hyper/issues) of this repo and believe that this is not a duplicate

- **OS version and name**: macOS 10.13.4
- **Hyper.app version**: 2.0.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/LinusU/a8d67a9459c989028682803a1a4307e7
- **Relevant information from devtools**: N/A
- **The issue is reproducible in vanilla Hyper.app**: Is Vanilla

## Issue

When outputting bold text, the colors specified in my `.hyper.js` file is no longer used, and instead, the default colors are used. This makes e.g. "bold cyan" look bad on a white background:

<img width="384" alt="screen shot 2018-04-16 at 18 44 01" src="https://user-images.githubusercontent.com/189580/38825929-279345e2-41a6-11e8-9c37-2ebac052aeee.png">

## Expected output:

This is how Terminal.app renders the same thing, with the same color settings:

<img width="411" alt="screen shot 2018-04-16 at 18 44 48" src="https://user-images.githubusercontent.com/189580/38825962-435d70fe-41a6-11e8-83d0-fae81bd76f5e.png">

(note how the prompt doesn't look the same in Hyper since that's using bold colors as well)

LinusU:
This is possibly related: https://github.com/xtermjs/xterm.js/issues/1239

chabou:
Totally related to xterm.

One workaround could be to disable brightColors. You can copy regular color as bright colors in your config:

```
  colors: {
      black: '#000000',
      red: '#C51E14',
      green: '#1DC121',
      yellow: '#C7C329',
      blue: '#0A2FC4',
      magenta: '#C839C5',
      cyan: '#20C5C6',
      white: '#C7C7C7',
      lightBlack: '#000000',
      lightRed: '#C51E14',
      lightGreen: '#1DC121',
      lightYellow: '#C7C329',
      lightBlue: '#0A2FC4',
      lightMagenta: '#C839C5',
      lightCyan: '#20C5C6',
      lightWhite: '#C7C7C7',
    },
```

Or use 1.x color scheme:
```
   colors: {
      black: '#000000',
      red: '#ff0000',
      green: '#33ff00',
      yellow: '#ffff00',
[...]

PR #7019, opened 2023-02-08
add option drawBoldTextInBrightColors

### LINKED ISSUE #5361: Option to disable bold text changing to bright colors
**Is your feature request related to a problem? Please describe.**
I don't want the color of bold text to change to the brighter version.

**Describe the solution you'd like**
I want an option to turn off the bold text color change.

In the underlying `xterm.js` library, there is a flag in `ITerminalOptions` called `drawBoldTextInBrightColors`, which by default is set true. I assume that setting this flag false will give my desired behavior. Can this be exposed as an option in the `hyper.js` config file? It looks like the `getTermOptions` function in `hyper/lib/components/term.tsx` can be changed to accept this additional option. Probably needs to be changed in other places as well.

I'm not the first to bring up this issue, but I haven't seen anyone mention the flag in the `xterm.js` library, so I assume it's relatively new. If that flag works as described, this should be a straightforward feature to implement.

**Describe alternatives you've considered**
I looked into making a plugin, but it doesn't look like the right way to solve this problem.

**Additional context**
Screenshot shows current unwanted behavior:
Screen Shot 2021-02-26 at 8 14 31 PM

comicchang:
Signed-off-by: Comic Chang <comicchang@gmail.com>

This PR adds a option to set [drawBoldTextInBrightColors](https://github.com/xtermjs/xterm.js/commit/ce6139e604c8d23dd6e3be957f5bed6d7e30bc5a) to xterm.js.

this should resolve #5361 #2836

Example `.hyper.js`
```
module.exports = {
  config: {
...
    drawBoldTextInBrightColors: false,
    // copied from https://github.com/ghosh/hyper-solarized-dark/blob/master/index.js
    colors: {
        lightBlack:     '#586e75',
        black:          '#073642',
        lightGreen:     '#586e75',
        lightYellow:    '#657b83',
        lightBlue:      '#839496',
        lightCyan:      '#93a1a1',
        white:          '#eee8d5',
        lightWhite:     '#fdf6e3',
        yellow:         '#b58900',
        lightRed:       '#cb4b16',
        red:            '#dc322f',
        magenta:        '#d33682',
        lightMagenta:   '#6c71c4',
        blue:           '#268bd2',
        cyan:           '#2aa198',
        green:          '#859900'
    },
...
}
```

before this commit
image

after
image


comicchang:
sorry

LabhanshAgrawal:
Thanks for the contribution, the new config structure is not fully finalized, I'm thinking of grouping some xterm related options in a separate key. I'll keep this pr open till the structure is finalized and then we can work on getting this merged.

weiqisun:
Hi, I just want to follow up on this thread. Is the new config structure fully finalized now so that this configuration option can be added? I tested the changes in this PR (with some minor modifications) on my [local branch](https://github.com/vercel/hyper/compare/canary...weiqisun:hyper:bold-bright-config) and it worked as expected. It will be great to have this feature in the released version so that I don't have to run my local build. Thanks!

typings/config.d.ts
```diff
@@ -90,6 +90,8 @@ type profileConfigOptions = {
   fontWeight: FontWeight;
   /** font weight for bold characters eg:'normal', '600', 'bold' */
   fontWeightBold: FontWeight;
+  /** if `false` hyper will draw bold text in normal colors instead of bright colors */
+  drawBoldTextInBrightColors: true;
```
weiqisun: I think the value of this field should be `boolean`, instead of `true`.

typings/hyper.d.ts
```diff
@@ -69,6 +69,7 @@ export type uiState = Immutable<{
   fontSmoothingOverride: string;
   fontWeight: FontWeight;
   fontWeightBold: FontWeight;
+  drawBoldTextInBrightColors: true;
```
weiqisun: same here

weiqisun (commented):
Thanks @comicchang for making this work! Just a minor comment regarding the code.

comicchang:
@weiqisun Thanks for your comment, fixed.

PR #7170, opened 2023-05-06
fix ssh url handling

### LINKED ISSUE #6767: ssh://host does not get handled correctly

- **OS version and name**: MacOS 12.6
- **Hyper.app version**: 3.3.0 (arm64)
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/mazedlx/fc6f844c798f622429a64ef2665711aa

## Issue
First of all, thanks for the great app! Here's the issue I'm dealing with:

Type `ssh://host` into the address bar of your browser, and press enter. Hyper should open `ssh host` but only runs `ssh`. This worked before but now seems to be broken. Works fine in iTerm (if it is set as the designated protocol handler for ssh).

LabhanshAgrawal:
fixes #6767

`new URL()` behavior seems to be different across chrome and node, which results in incorrect parsing in the renderer. Sending parsed data over ipc seems to fix this.


tendallas:
I've got problems with installing nightly build on mac. If someone will got system error use this command before: 
`xattr -cr /Applications/Hyper.app`

PR #7405, opened 2023-08-08
perf: avoid gpu memory increase

robinfai:
This modification can effectively control the memory usage growth of GPU processes as the number of tabs increases. In our internal terminal client, this modification achieved a 75% reduction in overall memory usage at the P95 percentile.

### DIFF
--- a/lib/components/term.tsx
+++ b/lib/components/term.tsx
@@ -475,6 +475,18 @@ export default class Term extends React.PureComponent<
     if (prevProps.rows !== this.props.rows || prevProps.cols !== this.props.cols) {
       this.resize(this.props.cols!, this.props.rows!);
     }
+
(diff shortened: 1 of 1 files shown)


LabhanshAgrawal:
Thanks for the pr. I think we should be able to use the fit addon when term gets active again instead of calling internal methods. Other than that, it looks good.

robinfai:
> Thanks for the pr. I think we should be able to use the fit addon when term gets active again instead of calling internal methods. Other than that, it looks good.

Because this size adjustment only involves the visible state of the terminal on the client-side, it is desired to keep the related operations within the client scope. Using the Fit addon for size synchronization involves exchanging size information between the client and the server, which may introduce some side effects.

PR #7544, opened 2023-10-18
Support alternate certificate authorities

cjntaylor:
Respect the `NODE_EXTRA_CA_CERTS` environment variable and set alternate certificate authorities from the path it specifies if defined and the file exists. This makes network access compatible with npm and other tools that also respect this variable.

Also update the environment passed to yarn to include the existing environment with overwritten values. This should also let yarn see this variable and use it for its network access.

This is necessary in environments that use SSL interception, where all certs are replaced by a single cert chain that isn't and can't be valid (it's created on the fly). This is becoming increasingly common within corporate environments, and is critical to developers who work in these spaces (hyper is unusable without this change).


kyet:
related issue: https://github.com/vercel/hyper/issues/2863

PR #7615, opened 2023-11-23
Security fixes

LabhanshAgrawal:


lib/actions/ui.ts
```diff
@@ -292,12 +292,17 @@ export function leaveFullScreen(): HyperActions {
...
 }
 
-export function openSSH(url: string) {
+export function openSSH(parsedUrl: ReturnType<typeof parseUrl>) {
   return (dispatch: HyperDispatch) => {
     dispatch({
       type: UI_OPEN_SSH_URL,
       effect() {
-        const parsedUrl = parseUrl(url, true);
+        const resourseIsValid = /^[a-zA-Z0-9.-]+$/.test(parsedUrl.resource);
```
vin01, line 300: Nice, it would be even nicer to disregard hostnames that begin with `-` as they can be interpreted as command line flags.

e.g. (https://github.com/git/git/blob/master/path.c#L1544, https://github.com/git/git/blob/master/connect.c#L1031)

PR #7736, opened 2024-01-21
Only handle "Select All" when Term has focus

### LINKED ISSUE #6733: Pressing

- **OS version and name**: macOS 12
- **Hyper.app version**: 3.3.0 (but also built canary from source)
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: N/A

## Issue
Impossible to use the "Select All" command in the search bar. Either pressing "Cmd + A" on the keyboard, or right clicking the search bar and selecting "Select All" in the menu, performs "Select All" on the terminal window rather than the search bar. 

Selecting all text in the search bar is really useful when navigating through the terminal content, and also this selecting the terminal content while the search box is in focus is really unexpected from the usability perspective.

JJJ:
Fixes #6733.

### DIFF
--- a/lib/containers/hyper.tsx
+++ b/lib/containers/hyper.tsx
@@ -36,7 +36,8 @@ const Hyper = forwardRef<HTMLDivElement, HyperProps>((props, ref) => {
 
   const handleSelectAll = () => {
     const term = terms.current?.getActiveTerm();
-    if (term) {
+    const focus = document.activeElement?.classList.contains('xterm-helper-textarea');
+    if (term && focus) {
       term.selectAll();
     }
   };


JJJ:
I purposely am opting to use vanilla JS here (instead of internal Hyper/React properties) with future-proofing in mind.

I originally started trying to stop this from inside of the `SearchBox` component, but the only approach I could think of was more invasive (importing Hyper/Term) while also only targeting keydown on the `SearchBox` input.

Doing this from inside the `Hyper` container ensures that if some other input (say, that a plugin might introduce) has focus, the "Select All" attempt will not be handled then either.

----
Another search-specific code change could be:
```javascript
    if (term && !term.props.search) {
      term.selectAll();
    }
```

PR #8187, opened 2026-06-03
feat: support platform-specific shell and env config

### LINKED ISSUE #2584: To support a single cross-platform settings file some of the settings should be platform specific (like in VS Code)

arnavnagzirkar:
Closes #2584.

Adds platform-specific overrides for shell, shellArgs, and env so a shared config can vary between linux, macOS, and Windows.

### DIFF
--- a/app/config.ts
+++ b/app/config.ts
@@ -9,6 +9,7 @@ import _openConfig from './config/open';
 import {cfgPath, cfgDir} from './config/paths';
 import notify from './notify';
 import {getColorMap} from './utils/colors';
+import {resolvePlatformConfig} from './utils/resolve-platform-config';
 
 const watchers: Function[] = [];
 let cfg: parsedConfig = {} as any;
@@ -76,8 +77,9 @@ export const subscribe = (fn: Function) => {
   };
 };
 
+export {resolvePlatformConfig} from './utils/resolve-platform-config';
+
 export const getConfigDir = () => {
-  // expose config directory to load plugin from the right place
   return cfgDir;
 };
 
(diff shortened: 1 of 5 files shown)


hiSandog:
This looks useful for anyone sharing one Hyper config across OSes. One compatibility detail worth testing is merge order for `env`: platform-specific overrides should probably replace only matching keys while preserving common env vars, otherwise a small Windows override could accidentally drop shared values used on macOS/Linux. A schema example showing common + per-platform `shell`, `shellArgs`, and `env` together would also make the new behavior easier to discover.

PR #8206, opened 2026-08-18
Paint body background on mount so new windows keep transparency

### LINKED ISSUE #6531: No transparency on startup

- **OS version and name**: 12.3.1 MacOS Monterey
- **Hyper.app version**: 3.2.3
- **Link of a [Gist](https://gist.github.com/) with the contents of your .hyper.js**: https://gist.github.com/Kinark/3a9f2a09b47a6f4ab123c60200679092
- **Relevant information from devtools** _(CMD+ALT+I on macOS, CTRL+SHIFT+I elsewhere)_: 
- **The issue is reproducible in vanilla Hyper.app**: Not sure

## Issue
Transparency is not working on first startup, but it comes back when I resize the window:
https://user-images.githubusercontent.com/17537081/166740694-41b77656-32ad-4c18-a7c4-a8ca9a1fd7f0.mp4

trevorhreed:
Fixes #6531.

New windows open without their configured background transparency (macOS, `backgroundColor` with alpha < 1). The tint only appears after the config file is touched or the window is resized.

**Root cause.** Electron ignores the alpha channel of `BrowserWindow` `backgroundColor`/`setBackgroundColor` on `transparent: true` windows, so the translucent background has to be painted by the renderer. The old fallback in `Hyper`'s `componentDidUpdate` (removed in #7299) only painted `document.body` when `backgroundColor` *changed* between renders — but `CONFIG_LOAD` is dispatched before the first render (`lib/index.tsx`), so a fresh window's first render already holds the final value and the paint never happens.

**Fix.** Re-add the body paint as an effect keyed on `backgroundColor`; `useEffect` runs on mount as well as on change, covering both the fresh-window case and live config edits.

**Verification.** Reproduced on 3.4.1 by driving a fresh instance over CDP with `backgroundColor: 'rgba(0,0,0,0.47)'` and no CSS workarounds:

- before: `document.body.style.backgroundColor === ''`, alpha-preserving `Page.captureScreenshot` pixels `[0,0,0,0]` — no tint at all;
- after executing exactly this paint statement in the live renderer: computed body background `rgba(0, 0, 0, 0.47)`, captured pixels `[0,0,0,120]` (120/255 ≈ 0.47).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/containers/hyper.tsx
+++ b/lib/containers/hyper.tsx
@@ -27,14 +27,10 @@ const Hyper = forwardRef<HTMLDivElement, HyperProps>((props, ref) => {
     handleFocusActive(props.activeSession);
   }, [props.activeSession]);
   useEffect(() => {
-    // Electron ignores the BrowserWindow backgroundColor (and setBackgroundColor)
-    // alpha channel on transparent windows, so the translucent background must be
-    // painted by the renderer. Running on mount as well as on change fixes new
-    // windows opening without their configured transparency (#6531).
+    // Electron ignores the alpha channel of BrowserWindow backgroundColor on
+    // transparent windows, so the renderer must paint the translucent
+    // background — on mount too, or new windows lose transparency (#6531)
     document.body.style.backgroundColor = props.backgroundColor;
-    return () => {
-      document.body.style.backgroundColor = 'inherit';
-    };
   }, [props.backgroundColor]);
 
   const handleFocusActive = (uid?: string | null) => {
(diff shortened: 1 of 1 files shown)


trevorhreed:
Ready to merge — nothing pending. `lib/containers/hyper.tsx` passes eslint and tsc; the repo-wide lint/type errors on canary predate this change. Maintainer edits are enabled.

